PR #4, opened 2022-05-28
AppBanner: remove futurebuilder

Feichtmeier:


lib/pages/explore/app_banner.dart
```diff
@@ -17,42 +19,55 @@ class AppBanner extends StatelessWidget {
...
 
-    Widget image = const _FallbackIcon();
-    for (var i = 0; i < snap.media.length; i++) {
-      if (snap.media[i].type == 'icon') {
+  Widget image = const _FallbackIcon();
+
+  @override
+  void initState() {
+    super.initState();
+    for (var i = 0; i < widget.snap.media.length; i++) {
```
jpnurmi, line 33: figuring out the URL looks like "logic" that could better fit in a view model?

lib/pages/explore/app_banner.dart
```diff
@@ -17,42 +19,55 @@ class AppBanner extends StatelessWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    final borderRadius = BorderRadius.circular(10);
 
+    bool light = Theme.of(context).brightness == Brightness.light;
+    final model = context.watch<SnapModel>();
+    if (iconIndex != null && widget.surfaceTint) {
+      context
+          .read<SnapModel>()
+          .getSurfaceTintColor(widget.snap.media[iconIndex!].url);
```
jpnurmi, line 56: does this belong in the build method or could it be more like "init" time operation?

lib/pages/explore/explore_page.dart
```diff
@@ -209,12 +209,12 @@ class _FilterBarState extends State<_FilterBar> {
 
 class _AppBannerCarousel extends StatelessWidget {
   const _AppBannerCarousel({Key? key}) : super(key: key);
-
   @override
   Widget build(BuildContext context) {
     final model = context.watch<AppsModel>();
     final size = MediaQuery.of(context).size;
-    return model.featuredSnaps.isNotEmpty
+    return model.sectionNameToSnapsMap['featured'] != null &&
```
jpnurmi, line 216: the old API looked more appropriate level of abstraction. maybe it could be left for convenience? :) for example something like:

    List<Snap> get featuredSnaps => _sectionNameToSnapsMap['featured'] ?? [];

lib/pages/common/snap_model.dart
```diff
@@ -217,4 +220,15 @@ class SnapModel extends SafeChangeNotifier {
   String get versionString => channels[channelToBeInstalled] != null
       ? channels[channelToBeInstalled]!.version
       : version;
+
+  Color get surfaceTintColor => _surfaceTintColor;
+  Color _surfaceTintColor;
+  Future<void> getSurfaceTintColor(String url) async {
+    if (_surfaceTintColor != Colors.transparent) return;
+    final paletteGenerator = await PaletteGenerator.fromImageProvider(
```
jpnurmi, line 228: this tint color calculation for urls sounds like a heavy operation that involves both downloading images and probably reading through the image data pixel by pixel for the final calculation of unknown complexity? one could even break this down to a service-like unit of its own that internally maintains a [cache](https://pub.dev/documentation/quiver/latest/quiver.cache/MapCache-class.html) to avoid unnecessary recalculation. :)

lib/pages/explore/app_banner.dart
Feichtmeier: it needs to be in the build method because otherwise the color is gone when you resize the window until next card of the carousel :(

iniside initstate

https://user-images.githubusercontent.com/15329494/170831006-fbb165a4-3dcd-4bca-8010-0b96fe85c872.mp4

inside build

https://user-images.githubusercontent.com/15329494/170831014-71839a50-dcb1-4786-b4f8-3c4aeae0f501.mp4

(sorry for the flickering, but I was too lazy to record the binary for this test :D)

lib/pages/explore/app_banner.dart
Feichtmeier: I would also prefer this but i need both the index and the icon, any idea?

lib/pages/common/snap_model.dart
Feichtmeier: I tested it with the debugger: after all tints are loaded for the first time, it won't be loaded again until resizing. But you are right this is still way too heavy. Considering this was as bad as this before in the futurebuilders let's move it to an issue and I try to find a solution after this PR?

lib/pages/explore/app_banner.dart
jpnurmi: sounds strange, i can take a look later after you've merged this :)

lib/pages/explore/explore_page.dart
```diff
@@ -226,23 +225,25 @@ class _AppBannerCarousel extends StatelessWidget {
               width: size.width,
               height: 178,
               autoScroll: true,
-              children: model.featuredSnaps
-                  .map(
-                    (snap) => AppBanner(
+              children: model.sectionNameToSnapsMap['featured']!.map(
```
jpnurmi, line 228: model.featuredSnaps?

lib/pages/explore/explore_page.dart
```diff
@@ -297,14 +298,15 @@ class _ExploreGrid extends StatefulWidget {
...
-    context.read<AppsModel>().loadFeaturedSnaps();
+    context.read<AppsModel>().loadSection('featured');
     super.initState();
   }
 
   @override
   Widget build(BuildContext context) {
     final model = context.watch<AppsModel>();
-    return model.featuredSnaps.isNotEmpty
+    return model.sectionNameToSnapsMap['featured'] != null &&
```
jpnurmi, line 308: model.featuresSnaps.isNotEmpty?

lib/pages/explore/explore_page.dart
```diff
@@ -313,22 +315,23 @@ class _ExploreGridState extends State<_ExploreGrid> {
...
-                        builder: (context) => ChangeNotifierProvider<SnapModel>(
-                          create: (context) => SnapModel(
-                              client: context.read<SnapdClient>(),
-                              huskSnapName: snap.name),
-                          child: const AppDialog(),
-                        ),
-                      ),
-                    ))
-                .toList(),
+            children: model.sectionNameToSnapsMap['featured']!.map((snap) {
```
jpnurmi, line 318: model.featuresSnaps

Feichtmeier: You are right, sorry. Will use the new getter

PR #14, opened 2022-05-29
WIP: Deb installer

Feichtmeier:
ultra WIP

grafik

- [X] get infos
- [ ] install
- [ ] remove

Edit: needs also package kit I think, sorry


robert-ancell:
The change is mixing up a couple of concepts - is this supposed to install a .deb file you have, or search the archive and install from there?

For the former, you use dpkg.dart as you show, but you need to use `PackageKitClient.installFiles` which isn't implemented.

For the latter, you either use `PackageKitClient.resolve` if you have the package name you want, or `PackageKitClient.searchNames` to find a matching package. Then you use `PackageKitClient.installPackages` to install from the archive.

Feichtmeier:
I personally do not really care which package to use, just want to install debs.
Thanks, then I try to remove the dkpg from that model and try the packagekit only approach

robert-ancell:
If you do want to install .deb files directly - I've added that to packagekit.dart - https://github.com/canonical/packagekit.dart/pull/15

robert-ancell:
Note you can also use `PackageKitClient.getPackages` to get all the installed and available .debs in case you want a big list of them.

Feichtmeier:
Thanks @robert-ancell 

A first iteration of the deb installer is now something I can merge
grafik

ToDo

- check if package is installed
- add a nice "Beer filling up" animation for the icon
- add debs to the updates page

PR #80, opened 2022-08-06
Change the "All apps" icon

### LINKED ISSUE #74: Change the "All apps" icon
The compas icon 🧭 used next to **All apps** should not be a compas.

It looks a bit odd that it's placed next to the same icon in the sidebar. More importantly, I associate the compas with exploring something, but here it should illustrate that all categories are searched. I don't really have a good idea for an "all" icon.
Don't know if the magnifier glass 🔍 (like we had earlier) would be better 🤷‍♂️ 

image

Feichtmeier:
Fixes #74


madsrh:
Works much better 👍

PR #107, opened 2022-08-26
PackageService: prevent app from closing if refreshing or updating

### LINKED ISSUE #106: PackageService: prevent app from closing if refreshing or updating

Feichtmeier:
https://user-images.githubusercontent.com/15329494/186981331-c879077c-463d-41bf-9375-0defb2ce2f9e.mp4

Thanks @jpnurmi for the idea

Fixes #106


jpnurmi:
Is there a danger that something could go wrong and an exception would be thrown between the `setClosable(true)` and `setClosable(false)` calls? Should `setClosable(false)` be in a [finally](https://dart.dev/guides/language/language-tour#finally) clause to avoid that one ends up stuck with a non-closable window?

PR #111, opened 2022-08-27
Use window managet set closable in a finally block

### LINKED ISSUE #110: Use window managet set closable in a finally block
Is there a danger that something could go wrong and an exception would be thrown between the `setClosable(true)` and `setClosable(false)` calls? Should `setClosable(false)` be in a [finally](https://dart.dev/guides/language/language-tour#finally) clause to avoid that one ends up stuck with a non-closable window?

_Originally posted by @jpnurmi in https://github.com/ubuntu-flutter-community/software/issues/107#issuecomment-1229221827_

Feichtmeier:
Fixes #110

### DIFF
--- a/lib/services/package_service.dart
+++ b/lib/services/package_service.dart
@@ -207,15 +207,17 @@ class PackageService {
 
   Future<void> refresh() async {
     windowManager.setClosable(false);
-
-    setErrorMessage('');
-    setUpdatesState(UpdatesState.checkingForUpdates);
-    await _refreshCache();
-    await _getUpdates();
-    setUpdatesState(
-      _updates.isEmpty ? UpdatesState.noUpdates : UpdatesState.readyToUpdate,
-    );
-    windowManager.setClosable(true);
+    try {
+      setErrorMessage('');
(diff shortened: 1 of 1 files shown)


lib/services/package_service.dart
```diff
@@ -207,15 +207,17 @@ class PackageService {
...
-    setUpdatesState(UpdatesState.checkingForUpdates);
-    await _refreshCache();
-    await _getUpdates();
-    setUpdatesState(
-      _updates.isEmpty ? UpdatesState.noUpdates : UpdatesState.readyToUpdate,
-    );
-    windowManager.setClosable(true);
+    try {
+      setErrorMessage('');
+      setUpdatesState(UpdatesState.checkingForUpdates);
```
jpnurmi, line 212: does the window need to be non-closable when just checking for updates?

Feichtmeier: Yeah I think yes. Maybe it is a bit too paranoid but packagekit is not as robust as snapd from my impressions

jpnurmi (approved):
LGTM :+1:

PR #118, opened 2022-08-29
Dialog with app details cannot be closed with a mouse when scrolled down

### LINKED ISSUE #99: Dialog with app details cannot be closed with a mouse when scrolled down
When inspecting the details of an app/package, if I expand the description section and scroll down, the cross icon at the top-right of the dialog scrolls out of view. I can't close the dialog anymore, unless I scroll back up (I know I can with the ESC key, but not with the mouse.

How about ensuring the close icon is always visible, and maybe also the header containing the icon, app name and short description?
The header could possibly shrink to a smaller version containing only the icon and app name when scrolling down, responsive UI style.

Feichtmeier:
https://user-images.githubusercontent.com/15329494/187248046-40eaaf90-86c6-46e8-8f29-23b6b10eff2a.mp4

Fixes #99

### DIFF
--- a/lib/store_app/common/snap_dialog.dart
+++ b/lib/store_app/common/snap_dialog.dart
@@ -49,7 +49,7 @@ class _SnapDialogState extends State<SnapDialog> {
     final model = context.watch<SnapModel>();
     if (model.name != null) {
       return AlertDialog(
-        scrollable: true,
+        scrollable: false,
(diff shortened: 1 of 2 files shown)


oSoMoN:
This looks better, but the mouseover highlight for the screenshot isn't clipped properly, see this:

image

Feichtmeier:
This is a different issue then 🙂 also fixable

PR #122, opened 2022-08-30
Clip SnapDialog Carousel mouse over correctly when scrolled off screen

### LINKED ISSUE #121: Clip SnapDialog Carousel mouse over correctly when scrolled off screen
The mouseover highlight for the screenshot isn't clipped properly, see this:

image

_Originally posted by @oSoMoN in https://github.com/ubuntu-flutter-community/software/pull/118#issuecomment-1230908026_

Feichtmeier:
grafik

The hover should only be displayed by the mousecursor as with the buttons on the left and right, thus removing the hover color fixes the issue

Fixes #121

### DIFF
--- a/lib/store_app/common/snap_content.dart
+++ b/lib/store_app/common/snap_content.dart
@@ -38,6 +38,7 @@ class SnapContent extends StatelessWidget {
               children: [
                 for (final url in media)
                   InkWell(
+                    hoverColor: Colors.transparent,
                     borderRadius: BorderRadius.circular(10),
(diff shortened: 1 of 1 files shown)


oSoMoN:
In fact it's not just a mouseover thing. The highlight can also be displayed when cycling through UI elements with the TAB key, and I can trigger the same issue with this branch.

Feichtmeier:
grafik

Solution for both hover and tap: wrap inkwell inside material

PR #125, opened 2022-08-30
SnapDialog: add tooltips for the timestamps fields.

oSoMoN:


oSoMoN:
image

Feichtmeier:
Thanks! Good idea

Just one thing - could you move the abstraction of null & DateTime --> string back to the model? 😺

Feichtmeier:
In the end this might all be opinion, but the more logic you add in the view the less readable the whole tree becomes, from my experience. :)

Jupi007:
You could add a private method to free up the widget tree.

Feichtmeier:
You are right, since locals are connected to the build context it shouldn't be localised in the model

Feichtmeier:
grafik

I think it's not working as expected :) I'll dig in

PR #134, opened 2022-08-31
SnapModel: localize install date tooltip

### LINKED ISSUE #129: SnapModel: localize install date tooltip
grafik

I think it's not working as expected :) I'll dig in

_Originally posted by @Feichtmeier in https://github.com/ubuntu-flutter-community/software/issues/125#issuecomment-1232619586_

Feichtmeier:
grafik

@oSoMoN @Jupi007 I didn't find how to get a iso norm but localized so I switched back to your iso norm preference, however, this time without making the view any more "intelligent". I would really like to keep the view as stupid as possible (like in the rest of the views).

Fixes #129

### DIFF
--- a/lib/store_app/common/snap_dialog.dart
+++ b/lib/store_app/common/snap_dialog.dart
@@ -67,6 +67,7 @@ class _SnapDialogState extends State<SnapDialog> {
               connectionsExpanded: model.connectionsExpanded,
               iconUrl: model.iconUrl ?? '',
               installDate: model.installDate,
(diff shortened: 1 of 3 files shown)


oSoMoN:
> I didn't find how to get a iso norm but localized

The following should do the trick:

```dart
return DateFormat.yMd(Platform.localeName)
        .add_jms()
        .format(_localSnap!.installDate!.toLocal());
```

Note that I had also added a timestamp tooltip for the "Last Update" field in `SnapChannelExpandable`. I'll submit a PR to add it back, together with the format change I suggested above.

PR #135, opened 2022-08-31
SnapDialog: use localized timestamps for the tooltips.

oSoMoN:
Also add back a timestamp tooltip for the "Last Updated" field.

Ref: #134.


Feichtmeier (approved):
LGTM :) ignore the failed CI 
We've opened this issue often enough

Issue #128, opened 2022-08-31
Use appstream.dart for packagekit icons and screenshots


Feichtmeier:
Unlike snapd, packagekit does not provide any icons or screenshots of the apps.
So to get those we need appstream.dart, which should be used in the package_service.dart here to fetch the metadata for X packages ad store them

IMO since snapd is so very fast with getting icons and screenshots I would not slow this app down by using appstream for snaps - Opinons ?  @oSoMoN @jpnurmi @Jupi007 

Depends on https://github.com/canonical/appstream.dart/issues/15

Jupi007:
Maybe a stupid question, but can't you just conditionally fetch the app icon: false for snap, true for other?

oSoMoN:
I haven't looked in the implementation details, but from your description of the problem I would indeed recommend using appstream only for packages where the icons/screenshots are not available otherwise (i.e. not snaps).

PR #160, opened 2022-09-02
Show progress of packagekit installations/uninstallations

### LINKED ISSUE #153: Show progress of packagekit installations/uninstallations
when installing a new app no progress is shown. Only the busy wheel appears

Feichtmeier:
Bildschirmfoto vom 2022-09-02 10-15-18

@alculete 
the dialogs will be merged to one layout in the future (so the packagekit dialog looks as fancy as the snap dialog... currently we do not have images yet as #128 needs to happen) , but this should fix the issue for now :)  

Fixes #153


alculete:
OK, looks good

Issue #91, opened 2022-08-22
Design discussion: explore page


Feichtmeier:
(click to enlarge)

## Explore Page

| Light | Dark |
|-|-|
|![grafik](https://user-images.githubusercontent.com/15329494/185951846-e768963a-cdfb-4c09-bf80-6364f5230d14.png)|![grafik](https://user-images.githubusercontent.com/15329494/185951951-41636428-d446-4020-b7ac-97d4a1137f49.png)|

## Large window

![grafik](https://user-images.githubusercontent.com/15329494/191794821-e9a65bb6-b5d3-413c-aa3a-ad8070efd4cf.png)

## Small window

![grafik](https://user-images.githubusercontent.com/15329494/191794748-74bd30a8-f1ba-4f3c-9217-70d0c555c908.png)

## Search

![explore_search_01](https://user-images.githubusercontent.com/15329494/185956381-8d1235fb-5f14-43c9-817e-26b901e2f52b.png)
![explore_search_02](https://user-images.githubusercontent.com/15329494/185956385-13ebbfea-30f4-425b-9335-d171ec459d32.png)

mivoligo:
Great job guys with the app!
Just started looking at these beautiful screenshots. I hope you don't mind if I add some feedback from time to time :slightly_smiling_face: 

So here goes the first one:
As long as I remember I always wanted to have a "Start" or "Open" button in a software center on an app's page. It would be useful for example after installing a new app. Here I can see there is such a button but IMO is not enough exposed and I thought it was just a button to go to the program's website. So personally I'd love to see a nice big button (as big as "Install") somewhere next to the app's title/icon row.

mivoligo:
Second one:
What if go fancy on an app's thumbnail and add one more line of description and even a screenshot?

mivoligo:
I'd also vote for removing the "show more" buttons for Description and details with the version and such. Instead just show them to the user without extra clicking.

madsrh:
@mivoligo You can edit your comment rather than create a new post 😉 

Great idea with the "Open/Start/Run/Launch/..." button 👍 

> What if go fancy on an app's thumbnail and add one more line of description and even a screenshot?

I agree that it would be great if the entire app tag line would fit. Most of them are short (like mailstring, notepadqq, slack) and that it properly the length we should aim for.
[...]

PR #205, opened 2022-09-14
Add verification status to account

### LINKED ISSUE #162: Add verification status to account
There is no way to verify if the software from that account is verified.

Feichtmeier:
grafik

CC @madsrh @mivoligo 

Things got crowded in there. So to fix this issue and clean up the header, I exchanged the semantics for the external link button with the homepage link. Combined with the verified icon if the publisher is verified.

Toolip
grafik

I moved the contact and website links (as they are two separate things sometimes, sometimes the same) to the description expandable:

grafik

Sorry for the german :P 

Any idea where to move the "open app" button now and how it should look?

Fixes #162

Ref #91


Feichtmeier:
How about clicking the app icon opens the app? :)

grafik

Feichtmeier:
Damn Paul, this is the solution! :+1: :D 

grafik

I like it

Jupi007:
Looks very good :D

madsrh:
What Paul said 😅

Looks great 🎉

mivoligo:
Really nice :)
Not sure how it's now with the scrolling. It would be great if those buttons were always visible.

Feichtmeier:
> 

The buttons at the bottom are always visible
The part that is scrollable is the middle (called it "content")

kenvandine:
The open button won't work in the snap yet.  We have a branch of snapd that adds app launch from another snap capability, but that will take a while to land, it's huge.  Can you hide the open button if $SNAP is set in the environment?

Feichtmeier:
Should be possible to bring this top down from main, yes

Issue #221, opened 2022-09-16
Crash on org.gtk.Application.Activate


kenvandine:
The app properly registers on the session bus as a singleton, but when invoked a 2nd time the running instance crashes.

kenvandine:
Thread 1 "software" received signal SIGSEGV, Segmentation fault.
0x00007ffff592bdc4 in hdy_window_mixin_add () from /home/ken/src/github/ubuntu-flutter-community/software/build/linux/x64/release/bundle/lib/libhandy_flutter.so
(gdb) thread apply all bt

Thread 28 (Thread 0x7fffa91ff6c0 (LWP 100759) "threaded-ml"):
#0  0x00007ffff5a78f3f in __GI___poll (fds=fds@entry=0x7fff740071a0, nfds=nfds@entry=2, timeout=timeout@entry=-1) at ../sysdeps/unix/sysv/linux/poll.c:29
#1  0x00007fffd045ebb5 in poll (__timeout=-1, __nfds=2, __fds=0x7fff740071a0) at /usr/include/x86_64-linux-gnu/bits/poll2.h:39
#2  poll_func (ufds=0x7fff740071a0, nfds=2, timeout=-1, userdata=0x82e080) at ../src/pulse/thread-mainloop.c:70
#3  0x00007fffd044830c in pa_mainloop_poll (m=m@entry=0x82df80) at ../src/pulse/mainloop.c:863
#4  0x00007fffd0452a7a in pa_mainloop_iterate (m=m@entry=0x82df80, block=block@entry=1, retval=retval@entry=0x0) at ../src/pulse/mainloop.c:945
#5  0x00007fffd0452b20 in pa_mainloop_run (m=0x82df80, retval=0x0) at ../src/pulse/mainloop.c:963
#6  0x00007fffd0462d3d in thread (userdata=0x82df30) at ../src/pulse/thread-mainloop.c:101
#7  0x00007fffb80a5733 in internal_thread_func (userdata=0x570050) at ../src/pulsecore/thread-posix.c:81
#8  0x00007ffff59f82e2 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:442
#9  0x00007ffff5a87240 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81

Thread 27 (Thread 0x7fffdc4376c0 (LWP 100750) "dart:io EventHa"):
[...]

PR #222, opened 2022-09-16
Activate existing instance

jpnurmi:
Close: #221


jpnurmi:
I'm unsure if we should add to `snapcraft.yaml`:
```yaml
common-id: io.snapcraft.Store
slots: [dbus-name]
```
The same changes were reverted from the installer but I think it was because it's classic: https://github.com/canonical/ubuntu-desktop-installer/pull/422/files

kenvandine (approved):
Thanks!

PR #239, opened 2022-09-20
UpdatePage: improve update count visualization

### LINKED ISSUE #238: Updating one package returns "everything is up to date"
When you only install one package the text implies that there are no more updates. It should just say something like "Update installed / install finished / done! / ..."

image

Feichtmeier:
Fixes #238

### DIFF
--- a/lib/services/package_service.dart
+++ b/lib/services/package_service.dart
@@ -349,7 +349,11 @@ class PackageService {
       setStatus(null);
       setProcessedId(null);
       setUpdatePercentage(null);
-      setUpdatesState(UpdatesState.noUpdates);
+      if (selectedUpdates.length == updates.length) {
+        setUpdatesState(UpdatesState.noUpdates);
+      } else {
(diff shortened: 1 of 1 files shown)


madsrh:
Yes, thanks! This works @Feichtmeier 

We would problery want some kind of notification that X-name package has been updated. Could we use [desktop_notifications](https://pub.dev/packages/desktop_notifications) or do you think it would be better to do it inside the app (like Nautilus does)?

Feichtmeier:
Yes we're already using desktop notifications for installation events

Something like 
X have been updated
But wait! There is more! Is not really need for  within the app
I think we can trust users to see that the list of updates shrinked

Feichtmeier:
Let's do this in another issue/pr ok @madsrh ?

madsrh:
> Let's do this in another issue/pr ok madsrh ?

OFC, I'm afk for work. Can you?

madsrh (approved):

PR #258, opened 2022-09-22
Navigate back on menu press

### LINKED ISSUE #254: Tabing on selected outer navigation should return to the first state of the page
When you're on an app page within Explore, you should be able to click "Explore" in the sidebar to return to the previous page (like the back button **<** does).

Similarly if you open the app page from "My apps" you can't click "My apps" in the sidebar to return. 

I've made a beautiful illustration below:
image

Feichtmeier:
While adding a Function(BuildContext) onTap to YaruPageItems was no problem, to use this to fix #254 we need a change to our tree.

The outer Navigator sits around the whole app (RED) while three inner navigaotrs sit around each page (ExplorePage, MyAppsPage, UpdatesPage) (GREEN)

grafik

grafik

This means popping the build context which is given from store_app.dart will pop the outer navigator and you end up with a white page :laughing: 

@jpnurmi any idea how to build the app in store_app.dart with for example defined route names or something so one could use popUntil... ?


jpnurmi:
@Feichtmeier We need to "reverse" the communication flow in one way or another. :) One possibility is to introduce a "controller" kind of object that is passed down to the children. That way they can listen to controller changes or even request page changes if desired. Another possibility is to make the ancestor an inherited widget so that children can access it from their own context and that way add a listener or pass a callback.

Feichtmeier:
Too many conflicts and also no solution in my head currently. Closing for now

PR #265, opened 2022-09-23
re-add-CI-for-main

Feichtmeier:
Was somehow disabled when the branch was renamed to main.


.github/workflows/analyze.yml
```diff
@@ -0,0 +1,22 @@
+name: flutter analyze
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  build:
```
jpnurmi, line 10: i'd suggest job names that describe what they do :)

.github/workflows/build-linux.yml
```diff
@@ -0,0 +1,20 @@
+name: flutter build linux
```
jpnurmi, line 1: consider a workflow as a "category" that can have multiple jobs related to that category. this sounds like a single job :)

.github/workflows/build-linux.yml
```diff
@@ -0,0 +1,20 @@
...
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: subosito/flutter-action@v1
+        with:
+          channel: "stable"
+      - run: sudo apt -y install git curl cmake meson make clang libgtk-3-dev pkg-config && mkdir -p ~/development && cd ~/development && git clone https://github.com/flutter/flutter.git -b stable && echo 'export PATH="$PATH:$HOME/development/flutter/bin"' >> ~/.bashrc && source ~/.bashrc
```
jpnurmi, line 19: - subosito/flutter-action is a github action that installs flutter for you - no need to manually install another flutter :)
- you can replace meson with ninja-build because flutter doesn't use meson

.github/workflows/analyze.yml
Feichtmeier: ah, okay, I hope this one fits better?

.github/workflows/build-linux.yml
Feichtmeier: Thanks for the review - great help!
Do the tests make sense like this?
analyze
build normally
build with flutter snap ?

jpnurmi: makes sense. other candidates are "flutter format" and "flutter test"

Feichtmeier: K 😌 I could indeed add format

For tests I would first need any test that makes sense

.github/workflows/analyze.yml
jpnurmi: sure, and if you want, the key can be `analyze:` so you don't need a separate name attribute

.github/workflows/format.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: subosito/flutter-action@v1
+        with:
+          channel: "stable"
+
+      # Consider passing '--fatal-infos' for slightly stricter analysis.
+      - name: format
+        run: flutter format
```
jpnurmi, line 23: `flutter format --set-exit-if-changed .` will make the check fail if there are wrongly formatted files

Feichtmeier: neat, ty!

.github/workflows/build-linux.yml
jpnurmi: having a CI check in place wouldn't hurt though because then it'll just work when pushing the first real test :)

jpnurmi (approved):
LGTM :+1:

.github/workflows/build-linux.yml
Feichtmeier: you are rright
added

PR #273, opened 2022-09-25
AppMedia: create a fullscreen carousel on tap

Feichtmeier:
@madsrh wanted a carousel within a carousel and I almost created a never ending recursion :D 
But he is right so here it is. Is this good

[Bildschirmaufzeichnung vom 25.09.2022, 11:46:03.webm](https://user-images.githubusercontent.com/15329494/192137543-8540ce10-9c0e-4d54-a644-a1fe70500f61.webm)

 @madsrh ?

### DIFF
--- a/lib/store_app/common/app_page.dart
+++ b/lib/store_app/common/app_page.dart
@@ -16,7 +16,6 @@
  */
 
 import 'package:flutter/material.dart';
-import 'package:flutter/services.dart';
 import 'package:software/store_app/common/app_data.dart';
 import 'package:software/store_app/common/app_description.dart';
 import 'package:software/store_app/common/app_header.dart';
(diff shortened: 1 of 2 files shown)


Jupi007:
Good idea :+1: 
I was thinking about add this feature to the carousel widget itself :D
Imo, this is an interesting feature which could be useful for other app.

madsrh:
Yes! 🥳 Looks very slick @Feichtmeier 👌

I haven't tested this yet. I'm afraid to ask, but can I use the keyboard? 🙈

Feichtmeier:
> Yes! 🥳 Looks very slick @Feichtmeier 👌
> 
> I haven't tested this yet. I'm afraid to ask, but can I use the keyboard? 🙈

Uhhh forgotten xD thanks, adding after lunch
Never be afraid please. Even if I say Jesus 😜😂

Feichtmeier:
@madsrh btw navigating with the touchpad already works as it is build into pageview/listviews/gridviews

However the default keyboard navigation which is activateable (outside of tabbing, which is always there) with defining the scrollview as primary is weird. It looks like you are switching focus :shrug: 

@Jupi007 you have functions to navigate back and forth in the carousel - could those functions be made public so one can use them in an outer keyboard listener?
Sure you could add the keyboard listener also into the carousel but then it would only work when your mouse is not outside of the carousel : | and you never know where users might be with there mouse

Feichtmeier:
https://user-images.githubusercontent.com/15329494/192141972-503067c3-2804-416f-a594-50a615f59aaa.mp4

Jupi007:
> @Jupi007 you have functions to navigate back and forth in the carousel - could those functions be made public so one can use them in an outer keyboard listener?

Well, as it is a statefull widget, you can't access to them even if we made them public.
So it would require to totally embedded the fullscreen mode inside it.

Feichtmeier:
Damn you are right. Well then we only have the option to do this and pray no user would move the mouse outside of the carousel in the full screen dialog.
Is there an option to somehow tell the carousel to be the primary scrollview to let flutter notice this and accept keyboard input?
That works with gridviews and listviews

Feichtmeier:
I revert the keyboard listener for this PR, let's do this in another one and or in yaruwidgets

Jupi007:
About the previous/next page control, I think that create a separate `YaruCarouselController` may solve the problem.
And about the keyboard, maybe use one of those flutter focus classes: grab the focus when the carousel is on fullscreen mode.

Issue #270, opened 2022-09-25
ExplorePage: Packagekit search is slower than snap search


soumyatheman:
The app search speed is very slow for debian packages, compare to the snap packages. ~~It takes almost a minute or more to load.~~

Feichtmeier:
Hi @soumyatheman 
I agree that it is slower than the snap search
But I can not say that it takes minutes
While the snap search is 1-2 seconds, packagekit search is like 2-5 seconds

[Bildschirmaufzeichnung vom 25.09.2022, 08:53:05.webm](https://user-images.githubusercontent.com/15329494/192132006-71af1c6c-b113-4e4f-98a5-3d0153ae547e.webm)

soumyatheman:
You're right sir, but is that fixable?

Feichtmeier:
Well well, we use packagekit.dart which "only" uses packagekit on the system, which uses dbus.
The deb search here is already much faster from *my* impression than in gnome software
It is ofc not ideal that the snap search is much faster than the packagekit search
But searching can't be really cached if you search online. I mean how do you want to predict what the user will type in?
I fear this is max search performance for packagekit :shrug: 
I personally think that 2-3 seconds for basically random input is quiet fast, even if snap search is kinda lightning fast

Eventually the author of packagekit.dart could say something more intelligent than me :)
CC @robert-ancell

soumyatheman:
I am sorry, I was in brave forum, discussing about the native messaging, that now works in firefox-snap very well. BTW, can you give me some detailed sources about dart programming and the learning curb for it.

jpnurmi:
A couple of commonly used techniques that may help:
- debounce searches to avoid spinning off a new search operation for every key store
- cancel a previous transaction that may be still ongoing to avoid piling up search operations

jpnurmi:
## Current
<details><summary>diff</summary>

```diff
diff --git a/lib/services/package_service.dart b/lib/services/package_service.dart
index 31e9b09..1228a42 100644
--- a/lib/services/package_service.dart
+++ b/lib/services/package_service.dart
@@ -604,6 +604,9 @@ class PackageService {
     Set<PackageKitFilter> filter = const {},
   }) async {
     if (searchQuery.isEmpty) return [];
+    print('searching "$searchQuery"...');
+    final sw = Stopwatch();
+    sw.start();
     final List<PackageKitPackageId> ids = [];
     final transaction = await _client.createTransaction();
     final completer = Completer();
[...]

PR #283, opened 2022-09-26
PackageService: cancel previous search if still ongoing

jpnurmi:
Ref: #270


Feichtmeier (approved):
Awesome! Checked out locally. Already feel-able performance boost! tyvm

Feichtmeier:
Awesome! Checked out locally. Already feel-able performance boost! tyvm

Issue #267, opened 2022-09-23
Failed to start Flutter on wayland 


Ethran:
After installing via snap: 
```
 sudo snap install snap-store --channel=preview/edge
 ```
 on Ubuntu 22.04 with all updates applied app is throwing error:
   ```
 ** (software:127930): WARNING **: 19:58:06.101: Failed to start Flutter renderer: No GL implementation is available
 ```
and displays empty window:
![image](https://user-images.githubusercontent.com/77555700/192031048-202fa481-2284-4107-bedd-0011370de875.png)

Issue does not appear using X11, there is one error, and everything is ok:
```
/usr/share/libdrm/amdgpu.ids: No such file or directory
```
Using version v0.1.3-alpha downloaded from github there is no errors thrown into terminal, and application is working ok. 
I also tested on Ubuntu 22.10 wayland,  with application installed via snap and it is working ok.

I am using laptop with dual graphics - integrated AMD and NVIDIA (ideapad 5 Pro-14ACN6 Laptop - Type 82L7)

**Comment by Feichtmeier** [comment]
Hi @Ethran 
Thanks for your report

So it could be assumed that the combination of the AMD GPU driver inside 22.04 and snaps using flutter could cause this error?

To be sure did you have your integrated AMD graphics or the Nvidia graphics active when the snap version of this app had this error at start?

**Comment by Ethran** [comment]
I was using AMD graphics to lunch application, turning off Nvidia graphics in bios does not help.

I don't know how to lunch it using Nvidia GPU:
I have tried lunching it using dedicated Nvidia graphics: right click on application and option "lunch using discrete graphics card". Same result - empty window. But checking in nvidia-smi it shows: "No running processes found". 


### It might be related:
Sometimes during update it throws warning about AMD GPU. Here are those warning running prime-select commend: 

```
wiktor@wiktor-IdeaPad-5-Pro-14ACN6:~$ sudo prime-select nvidia
[sudo] password for wiktor: 
Info: selecting the nvidia profile
Deleting /lib/modprobe.d/nvidia-runtimepm.conf
Updating the initramfs. Please wait for the operation to complete:
/W: Possible missing firmware /lib/firmware/amdgpu/yellow_carp_gpu_info.bin for module amdgpu
W: Possible missing firmware /lib/firmware/amdgpu/vangogh_gpu_info.bin for module amdgpu
\W: Possible missing firmware /lib/firmware/amdgpu/cyan_skillfish_rlc.bin for module amdgpu
[...]

PR #285, opened 2022-09-26
Added layout to make amdgpu.ids available at the expected location

kenvandine:
Added layout to make amdgpu.ids available at the expected location, attempting to fix issue #267


Feichtmeier:
Thanks! Let's see if this helps 😃

PR #307, opened 2022-09-29
Translation of texts in Spanish and button correction.

Cubel89:
Added new Spanish translations and corrected the "Show more" button. The text was written without going through the translation system. Now it makes use of l10n.


Feichtmeier (approved):
Gracias!

PR #354, opened 2022-10-04
AppPage: use custom back button

Feichtmeier:


jpnurmi:
Maybe we could even have this as `YaruBackButton` in yaru_widgets?

Feichtmeier:
> Maybe we could even have this as `YaruBackButton` in yaru_widgets?

Either this or maybe some 

YaruIconButton versions

.back
.toggle

?

jpnurmi:
sounds good! :+1:

PR #378, opened 2022-10-10
Initial tests.

### LINKED ISSUE #377: No meaningful tests
We should aim at adding tests for the existing models/services/widgets, and ensure that newly added code comes with tests too.

oSoMoN:
Those are far from complete, this is just to bootstrap writing tests for the project.

Fixes #377.


jpnurmi:
@Feichtmeier do you have anything against removing pubspec.lock?

jpnurmi:
Did I? They recommend having it for apps but I never understood the point of having a lock file that gets regenerated as soon as you run pub get, which is implicitly done by flutter run or build. When are the locked versions actually used? :sweat_smile:

Feichtmeier:
Ok, so we remove it but aim to go back to fixed normal vesions in pubspec.yaml as soon as we got the new yaru widgets version out?

jpnurmi:
Yes, it's a special situation with yaru_widgets right now. Normally, we'd be using published versions.

test/services/package_service_test.dart
```diff
@@ -0,0 +1,390 @@
...
+import 'package:ubuntu_service/ubuntu_service.dart';
+
+import 'package_service_test.mocks.dart';
+
+class MockTransaction extends MockPackageKitTransaction {
+  final StreamController<PackageKitEvent> controller =
+      StreamController<PackageKitEvent>.broadcast();
+
+  MockTransaction() {
+    when(events).thenAnswer((_) => controller.stream);
```
jpnurmi, line 22: not sure if it makes any practical difference here but it's not recommended to mock own responses

https://pub.dev/packages/mockito#best-practices

jpnurmi (approved):
LGTM :+1:

test/services/package_service_test.dart
oSoMoN: Thanks! I realized that the MockTransaction class wasn't even needed, so I removed it altogether.

PR #385, opened 2022-10-11
StoreModel: allow all types of networks to count as online

### LINKED ISSUE #361: Store is offline on VPN connection
When connected to a VPN `connectivity_plus` falsely reports `ConnectionType.mobile` due to a [bug](https://github.com/fluttercommunity/plus_plugins/issues/1166) I just reported - this causes `StoreModel` to think we're offline since it requires either `ConnectionType.wireless` or `ConnectionType.ethernet`.

Feichtmeier:
@d-loose could you test it with your connections?

Fixes #361

### DIFF
--- a/lib/store_app/store_model.dart
+++ b/lib/store_app/store_model.dart
@@ -95,9 +95,7 @@ class StoreModel extends SafeChangeNotifier {
     });
   }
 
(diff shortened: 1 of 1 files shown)


d-loose:
Works :+1:

Issue #387, opened 2022-10-13
Level up tests coverage


oSoMoN:
As of 2022-10-13, it's at 17%, that's not good enough to guarantee code quality.

d-loose:
Current coverage is at around 68%. More UI tests will follow :)

PR #388, opened 2022-10-13
Unit tests for SnapService.

oSoMoN:
#387


jpnurmi (commented):
there are a few occurrences of:
```dart
//verify(mockSnapdClient.loadAuthorization()).called(1);
```

is it expected to be called every time? or is it basically `// TODO: verifyNever(...)`?

oSoMoN:
> is it expected to be called every time? or is it basically `// TODO: verifyNever(...)`?

That's because the current implementation of the service isn't consistent in where/when it calls `loadAuthorization()`, and I'm not sure why. My expectation is the one that's commented out (i.e. it should be called, but it isn't). I want to get to the bottom of this, but I think it's better suited for a follow-up PR.

jpnurmi (approved):
Ok! LGTM :+1:

PR #397, opened 2022-10-18
Restore pubspec.lock

jpnurmi:
Snap builds on Launchpad started failing after `pubspec.lock` was removed in #378:
image
https://launchpad.net/~ubuntu-desktop/snap-store-desktop-next/+snap/snap-store-preview

I have confirmed that restoring the lock file helps: https://launchpad.net/~jpnurmi/+snap/snap-store-test/+build/1914464

I'd propose restoring the lock file until we figure this out.

P.S. The underlying issue is the same as https://github.com/canonical/ubuntu-desktop-installer/issues/1146

PR #401, opened 2022-10-19
Remove (again) pubspec.lock.

oSoMoN:
Add a more elegant workaround to the failing snap builds problem (see #397).


oSoMoN:
Note: this is still a workaround, because I couldn't get to the bottom of what's going on when building as root on Launchpad.

Despite failing with a non zero error code, the first invokation of `flutter pub get` appears to be fetching all the information it needs from pub.dev, and writing it correctly to `/root/.pub-cache/`.

jpnurmi (approved):
Phew, we don't need the lock file after all! :tada:

jpnurmi:
@oSoMoN did you manage to get remote-build working or how did you debug it?

PR #406, opened 2022-10-19
Basic integration tests.

oSoMoN:
These do end-to-end tests of the application, rather than testing the exact behaviour of selected widgets.
They should therefore be less sensitive to refactoring and visual changes, while allowing us to catch functional breakage early.


oSoMoN:
It looks like #216 is affecting the packager installer app, too (the integration test is failing because the app is localized in Danish, apparently).

jpnurmi:
Does forcing `LANGUAGE=en_US` help? Although, the language shouldn't matter because it's looking up localizations from the context. Danish tests pass for me locally:
```sh
LANGUAGE=dk_DK flutter test integration_test
```

lib/services/package_service.dart
```diff
@@ -472,6 +472,7 @@ class PackageService {
     await removeCompleter.future;
     if (_localId != null) {
       setIsInstalled(false);
+      await getDetails(packageId: _localId!);
```
jpnurmi, line 475: are these `package_service.dart` changes something that could be pushed separately coupled with unit tests?

.github/workflows/test.yml
```diff
@@ -7,21 +7,27 @@ on:
     branches: [main]
 
 jobs:
-  build:
-    name: test
+  coverage:
```
jpnurmi, line 10: +1 for this but it requires a repo settings update. leave to a separate change as it's not strictly related?

integration_test/software_test.dart
```diff
@@ -0,0 +1,129 @@
...
+
+        expect(aboutDialog, findsNothing);
+      });
+    });
+  });
+
+  group('Package Installer App', () {
+    testWidgets('Install local deb', (tester) async {
+      final localDeb =
+          File('integration_test/assets/hello_2.10-2ubuntu4_amd64.deb');
```
jpnurmi, line 54: this is awesome :star_struck:

lib/services/package_service.dart
oSoMoN: Good suggestion, I split it off as #410.

.github/workflows/test.yml
oSoMoN: Which setting would need updating?

integration_test/software_test.dart
```diff
@@ -0,0 +1,129 @@
...
+import '../test/test_utils.dart';
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  tearDown(resetAllServices);
+
+  group('Store App', () {
+    group('Settings page', () {
+      testWidgets('About dialog', (tester) async {
```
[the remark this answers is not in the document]
oSoMoN: Fair enough, I'll remove it when I figure out why the "Install local deb" test fails.
I'm keeping it for now, because the overhead is minimal (~3 seconds), and having two separate tests run in succession has allowed me to catch problems earlier while working on this branch.

.github/workflows/test.yml
jpnurmi: Sorry for being so vague. :) I noticed earlier that the CI was waiting for a required "test" check so I assumed it was because of this

image

jpnurmi: it's in "branches" -> "branch protection rules" for `main`

integration_test/software_test.dart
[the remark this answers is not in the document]
jpnurmi: makes sense! and fine by me to keep it for now. maybe we can replace it with a local .snap installation in the future if it will be supported

.github/workflows/test.yml
oSoMoN: Thanks for the tip! I restored the name of the test for now.

oSoMoN:
@jpnurmi this is all green now, only waiting for your green light.

jpnurmi (approved):
Very nice! LGTM :+1:

PR #410, opened 2022-10-20
Fix package service to update the properties of a local package when it has been installed/removed.

oSoMoN:


oSoMoN:
Note that I plan to rework the PackageService class in depth to make it stateless, so this will be refactored, but this allows unblocking #406.

lib/services/package_service.dart
```diff
@@ -472,6 +474,7 @@ class PackageService {
     await removeCompleter.future;
     if (_localId != null) {
       setIsInstalled(false);
+      await getDetails(packageId: _localId!);
```
Feichtmeier, line 477: What's the reason to call getDetails here?

lib/services/package_service.dart
```diff
@@ -693,6 +704,7 @@ class PackageService {
     await installCompleter.future;
     if (_localId != null) {
       await isIdInstalled(id: _localId!);
+      await getDetails(packageId: _localId!);
```
Feichtmeier, line 707: getDetails should only work for packageIDs in the repo I believe to remember, has been some while since I worked on this

let me test this PR please

lib/services/package_service.dart
oSoMoN: To ensure that the details about the local package are updated. You can test this by invoking the package installer app with a local deb, installing it, then removing it. The value of the "Size" field should be updated after installing and removing, but it isn't.

Feichtmeier: Alright. It's a bit sub optimal that this service is used for both the package installer and when installing a deb from the repo in the explore page. I want to test it to see if the other packages work as before

oSoMoN: Sure! I won't merge this before you approve it.

Feichtmeier: ok tested out locally both the installer and the store app
store app works as before and your fix works

funnily I noticed that when you remove an installed deb app (for example gnome-calendar) the installed status is not updated
I think the code in isInstalled() is wrong (and the name is also wrong -.-)

Would you like to fix this also for store packages? Or a different PR?

oSoMoN: Let's fix this in another PR. Would you mind filing an issue and assigning it to me?

PR #418, opened 2022-10-21
Ensure transaction event stream subscriptions aren't leaked.

### LINKED ISSUE #411: PackageService leaks transaction event stream subscriptions
There's a recurring pattern in `PackageService` to listen to transaction events without ever closing the stream subscription:
```dart
final transaction = await _client.createTransaction();
transaction.events.listen(...);
...
```

I'd suggest enabling the [cancel_subscriptions](https://dart-lang.github.io/linter/lints/cancel_subscriptions.html) lint _and_ always handling the return value of `Stream.listen()`:
image

oSoMoN:
Fixes #411.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -7,6 +7,7 @@ analyzer:
 linter:
   rules:
     avoid_print: true
+    cancel_subscriptions: true
     prefer_single_quotes: true
     prefer_const_constructors: true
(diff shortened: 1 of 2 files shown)


jpnurmi (approved):
Nice, much cleaner than I thought it would be! LGTM :+1:

PR #441, opened 2022-10-26
PackageModel and PackageService refactoring.

oSoMoN:
* Merge PackageModel and PackageInstallerModel.
* Make PackageService accept PackageModel arguments for common package operations, and update their properties instead of exposing those on the service itself.
* Update existing tests, and add unit tests for PackageModel.


oSoMoN:
Note: `PackageService` was only partially refactored (only the methods and attributes specific to `PackageModel` and `PackageInstallerModel`). In a follow-up PR, I intend to apply similar changes to `UpdatesModel`.

Feichtmeier:
Nice idea!
THe only thing I am not really happy about is moving the model into the service dir
I would rather flatten the whole app hiearchy and remove this "store_app" "package_installer" dirs to lib
Then you can leave the model inside common/packagekit
WDYT?

Feichtmeier:
Ah one very important thing:

The idea behind all those streams was to be able to stay intune what happens in the update page, even if you switch around the views. With the streams you can switch the views and the newly created models are always up to date with the service

Is this still possible with this approach? @oSoMoN

Feichtmeier:
I saw you only removed the streams in the package model and not the updates model

https://github.com/ubuntu-flutter-community/software/blob/7326a0acdd6cbed9db219eaa39937ce460131bb0/lib/store_app/updates/updates_model.dart

This should be okay as long as you don't remove the streamsubs inside Updatemodel
As those are really needed to not lose the constantly changing values set while updates are running and people navigate around while updating

I would personally prefer if we keep the streams in all models related to packagekit as this would be the unified way to read out the service, even if those masses of subs might be a bit inconvenient

Eventually it would be best if we would first flatten the dir to

lib/
   services/
   common_widgets/
      packagekit/
      snap/
   explore/
   my_apps/
   updates/
   settings/

In the first pr.

Then in a second pr refactor the models if needed to be one model for packagekit.

oSoMoN:
> I saw you only removed the streams in the package model and not the updates model
> 
> https://github.com/ubuntu-flutter-community/software/blob/7326a0acdd6cbed9db219eaa39937ce460131bb0/lib/store_app/updates/updates_model.dart
> 
> This should be okay as long as you don't remove the streamsubs inside Updatemodel As those are really needed to not lose the constantly changing values set while updates are running and people navigate around while updating

Yes, for now I refactored only `PackageModel` and related methods in the service. I'll be careful not to break the existing functionality when touching `UpdatesModel` of course, which is why I left it for a follow-up PR, as this one is already big enough and not trivial to review (but I tested it extensively, and the unit tests together with the one integration test make me confident).

Note that using a stream on the service for property updates related to one given package was *potentially* broken anyway, consider this case: the old `PackageModel` (and `PackageInstallerModel`) would listen to `packagePercentage` events from the service, so assuming that we can have two packages being installed concurrently (maybe packagekit doesn't allow that, but that's an implementation detail without a strong guarantee), you would end up with two models getting mixed up percentage updates.

The same goes for other properties such as `description`, `changelog`, `isInstalled`…
That last one is actually less of a theoretical case, because I don't think packagekit prevents you from querying the installed state of several packages concurrently, and there is no guarantee on the ordering of the responses, so we might end up with an incorrect state for a given package.


> I would personally prefer if we keep the streams in all models related to packagekit as this would be the unified way to read out the service, even if those masses of subs might be a bit inconvenient

It might make sense to keep streams for properties related to package updates, where they are not specific to one given package, but truly a stateful property of the service itself. I haven't looked at that closely yet.


> Eventually it would be best if we would first flatten the dir to
> 
> ```
> lib/
>   -> services/
>   -> common_widgets/
>      ->-> packagekit/
>       ->->snap/
>   -> explore/
>   -> my_apps/
>   -> updates/
>   -> settings/
> ```
> 
> In the first pr.
> 
> Then in a second pr refactor the models if needed to be one model for packagekit but don't remove the streams please.

The re-structuring can be done at any point, and would be better done in a separate PR that does only that, I think.

Feichtmeier:
> The re-structuring can be done at any point, and would be better done in a separate PR that does only that, I think.

Yes but then the model should not be moved into the service package right now.

> It might make sense to keep streams for properties related to package updates, where they are not specific to one given package, but truly a stateful property of the service itself. I haven't looked at that closely yet.

Yes, there is no other way to keep track of what is happening while updating when we allow the user to switch views, which would be a bit weird to not allow, honestly. That is actually the only reason I introduced those streams/streamsubs.
(without using funny packages like [get](https://pub.dev/packages/get) :S :D )

Okay then we can add the stream-less getters in the packagemodel, but really need to keep the streamsubs in the updatemodel.

Yet it would be nice if you do not move the model into the service dir please

oSoMoN:
> Yet it would be nice if you do not move the model into the service dir please

Done, I moved it back to where it was.

Feichtmeier:
Checked out the pull request locally.
Could not find any bad behaviour but I have one more question
What would happen if a debian file is so large that it takes more than just some seconds to install it?
And then you navigate back from the page, then navigate back into the page?
For snaps I created a mechanism to track snapd changes and update it on model init (installing/installed/not installed)
I think with streams we would still have the option to track this installation/deinstallation progress no matter where we navigate to/from while installing/uninstalling

oSoMoN:
> What would happen if a debian file is so large that it takes more than just some seconds to install it?

We can't be sure until we fix #444. I'll take a stab at that (in a separate branch).

Feichtmeier:
Okay, so shall we still merge this one here?

oSoMoN:
> Okay, so shall we still merge this one here?

If you're okay with it, let's do it.

Issue #444, opened 2022-10-27
Installing/removing a deb package locks the PackageService, cannot search for packages until the operation is complete


oSoMoN:
Steps to reproduce:
 1. search for "0ad-data"
 2. open the result for the deb package with that name (it's a large package, installed size: 3296 MB)
 3. click the "Install" button, installation will take a while, even on a system with fast I/O
 4. click the back button to go back to the search results

Expected result: the results from the previous search are still displayed, and one can issue a new search query.
Actual result: a blank result page with a spinner that doesn't go away until the package is done installing, which on my machine can take ~ 1 minute.

oSoMoN:
There are actually two problems here:

1. The search results should be cached, so when going back from a package's details page to the search results the page should be displayed instantly
2. Searching for packages in the apt cache should be allowed while a package is being installed

Feichtmeier:
As I've brain stormed in the meeting for 2. it could maybe work to have two instances of a packagekit service and every instance receives a new channel to that packagekit dbus interface?
If this works we could eventually split the service into two or three services

oSoMoN:
If we can't make it work really asynchronously with one instance, yes.

oSoMoN:
> 1. The search results should be cached, so when going back from a package's details page to the search results the page should be displayed instantly

So in fact the results are cached, but what happens is that `ExploreModel` has a `selectedSnap` and `selectedPackage` observable properties that are updated when popping a package details page, resulting in the widget for the `ExplorePage` being rebuilt because the model has changed.

That bit can be fixed easily enough by removing these properties altogether, the model doesn't need to track which details page is currently being displayed.

Feichtmeier:
I think this fixed since #464

PR #447, opened 2022-10-28
Do not track the currently displayed app/snap in the models (ExploreModel, MyAppsModel).

oSoMoN:
This ensures that the corresponding pages are not rebuilt when popping a details page (#444).


Feichtmeier (approved):
Checked out locally! Works! Thank you very much

PR #454, opened 2022-10-31
remove the hide on close trick made with windowmanager

### LINKED ISSUE #453: remove the hide on close trick made with windowmanager

Feichtmeier:
https://user-images.githubusercontent.com/15329494/198978908-50ca04b1-5051-4e4e-b05f-b09738328451.mp4

Fixes #453

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -33,7 +33,6 @@ Future<void> main(List<String> args) async {
   WidgetsFlutterBinding.ensureInitialized();
   await windowManager.ensureInitialized();
   windowManager.setPreventClose(true);
-  windowManager.addListener(MyWindowListener());
 
   registerService<NotificationsClient>(
     NotificationsClient.new,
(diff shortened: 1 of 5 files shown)


Feichtmeier:
@oSoMoN could you eventually add some more information to the tests you've written?

Currently I make changes to classes and some tests I never created or have a food in crash and I need to look up why. 
I know tests are great etc etc but this is a bit heavy stop pedal for progressing with design iterations I make or issue fixing

For example now I inited the store model with the lastUpdateState field of the packageService
this field is often needed because vanilla dart streams do not safe the last value put into the stream and when you init a model, which normally happens all the time if it does not sit above all widgets, then you need a value for things you check. 
I added this to the model initialisation and the test failed because "there is no such method". 

```
Add a stub for this method using Mockito's 'when' API, or generate the MockPackageService mock with
the @GenerateNiceMocks annotation (see
https://pub.dev/documentation/mockito/latest/annotations/MockSpec-class.html).
```

Then I saw the packageServiceTest accessed the field directly so I couldn't just make the field private and add a getter :shrug: 

I can remove the line in this case because the store model sits above the whole store app and the value of updatesState should always be accessed after at least one value has been put into the stream - currently (but this may change as soon as the update check comes from outside of the app).
I decided to remove this line for now, because otherwise I would have to change the service to satisy this test / mock machinery thing.

oSoMoN:
@Feichtmeier this kind of issue is caused by the use of Mockito, which requires regenerating the mocks whenever the mocked models change (for example when a property is added/removed, etc, see the documentation at https://pub.dev/packages/mockito).

I agree this is inconvenient, and a hassle. JP suggested looking into [mocktail](https://pub.dev/packages/mocktail), which doesn't require generating mocks files, so I'll give it a try. If it works well enough, this should make our lives easier.

Feichtmeier:
ok :+1: Thanks for taking care of tests :bow: (I am lacking knowledge here)

oSoMoN:
Done in #456

PR #462, opened 2022-11-01
Fix local package installation by looking at the local path first.

### LINKED ISSUE #302: PackageInstaller needs to work in the snap
We do not discuss this very often :laughing: but the app also has the feature to install local debian files from the file manager
Looks like this... 
grafik

Install / Uninstall works fine

When I start the app from the snap-store command in the cli, packagekit gives an error:

grafik

```bash
frederik@carbon:~/Downloads$ snap-store google-chrome-stable_current_amd64.deb 
Unsupported modifier, resource creation failed.
XXX: resource creation failed
Unsupported modifier, resource creation failed.
XXX: resource creation failed
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: org.freedesktop.PackageKit.Transaction.NoSuchFile: File google-chrome-stable_current_amd64.deb is not found or unsupported
#0      DBusClient._callMethod (package:dbus/src/dbus_client.dart:1131)
<asynchronous suspension>
#1      DBusClient.callMethod (package:dbus/src/dbus_client.dart:608)
<asynchronous suspension>
#2      PackageKitTransaction.getDetailsLocal (package:packagekit/src/packagekit_client.dart:1030)
<asynchronous suspension>
#3      PackageService.getDetailsAboutLocalPackage (package:software/services/package_service.dart:661)
<asynchronous suspension>

Unsupported modifier, resource creation failed.
XXX: resource creation failed
```

Any idea why this happens in the snap but not with the local binary?
CC @kenvandine 

Classes used for this are here:
https://github.com/ubuntu-flutter-community/software/tree/main/lib/package_installer

Here:
https://github.com/ubuntu-flutter-community/software/blob/ed77278b0e417997eb8a2757ddd8adb4846d8bbb/lib/services/package_service.dart#L643-L687

And in main.dart
https://github.com/ubuntu-flutter-community/software/blob/main/lib/main.dart

oSoMoN:
Fixes #302.

### DIFF
--- a/lib/store_app/common/packagekit/package_model.dart
+++ b/lib/store_app/common/packagekit/package_model.dart
@@ -184,14 +184,14 @@ class PackageModel extends SafeChangeNotifier {
   }
 
   Future<void> install() async {
-    if (_packageId != null) {
-      return _service
-          .install(model: this)
-          .then(_updateDetails)
-          .then(_updatePercentage);
-    } else if (_path != null) {
+    if (_path != null) {
       return _service
           .installLocalFile(model: this)
+          .then(_updateDetails)
+          .then(_updatePercentage);
+    } else if (_packageId != null) {
+      return _service
+          .install(model: this)
           .then(_updateDetails)
           .then(_updatePercentage);
     }


Feichtmeier:
@oSoMoN good job, fixed ! :)

PR #473, opened 2022-11-02
Cancel a background updates check when opening the details page for a package.

### LINKED ISSUE #465: Cancel a background updates check when opening the details page for a package
This is related to #444.

Because packagekit.dart serializes requests, we cannot run concurrently an updates check and retrieve the details for a given package.
When a user searches for packages and opens the details page for a given package, we should cancel any ongoing updates check to retrieve the details for the package as fast as possible.

oSoMoN:
Fixes #465.

### DIFF
--- a/lib/services/package_service.dart
+++ b/lib/services/package_service.dart
@@ -46,6 +46,8 @@ class PackageService {
     _client.connect();
   }
 
+  PackageKitTransaction? _pendingUpdatesCheckTransaction;
+
   final Map<PackageKitPackageId, bool> _updates = {};
   List<PackageKitPackageId> get updates =>
       _updates.entries.map((e) => e.key).toList();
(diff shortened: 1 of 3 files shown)


Feichtmeier:
hmmmmm doesn't seem to work for me

https://user-images.githubusercontent.com/15329494/199493281-3427a09a-6695-4a80-9b11-9a17306a4e23.mp4
(sorry for the music)

oSoMoN:
That's because this change will cancel the currently running updates check only when opening the details page for a given package. I overlooked that it should be cancelled when opening the search/my apps pages as well. I'll work on this.

d-loose:
Has this become obsolete thanks to #567?

Feichtmeier:
@d-loose for package pages loaded with appstream infos yes. But not for example for a case where we check for updates and then someone clicks on a deb in the filemanager. THe app does nto break but if there are really many updates the AppLoading page is probably shown for a long time

d-loose:
@Feichtmeier but this PR resolves exactly that - it cancels an ongoing update in `PackageModel.init`

https://user-images.githubusercontent.com/113362648/211818847-736a2d4b-4a71-4102-8d64-d25a439e4a6f.mp4

Feichtmeier:
> @Feichtmeier but this PR resolves exactly that - it cancels an ongoing update in `PackageModel.init`
> 2023-01-11.14-30-26.mp4

then we need this PR I think :), or ?

d-loose:
Ah yes, of course - I thought I asked whether it needs additional work.. Since the issue you described earlier got fixed by #567 I think it's good to go as it is.

Feichtmeier:
There are bunch of conflicts though

grafik

Feichtmeier:
Merged in the other PR, thank you @oSoMoN !

Issue #481, opened 2022-11-02
Add terminal view with details while updating


Feichtmeier:

hsbasu:
linking duplicate issue #292

Feichtmeier:
Fixed with #489

PR #489, opened 2022-11-04
Feichtmeier/issue481

Feichtmeier:
Ref #481 

Needs to be re-implemented with a more light-weight widget


madsrh:
I can see that you've tweaked the colors. You don't want a background?

image

Feichtmeier:
I thought you don't want a background! XD

I need to rework this to be not a fully fledged terminal because this is not good for resources

PR #521, opened 2022-11-15
L10n add text line for refresh button on the updates page

### LINKED ISSUE #511: Confusing refresh translation line (Russian)
In Russian both `update` and `refresh` translates to the same word `обновить`, both of these terms are being used side by side on the updates page, which creates confusing situation for Russian-speaking people. Also, the same text line for `refresh` is being used on an installed snap app page.
To solve this problem, please decouple text line for refresh of a snap and refresh of a list of updates.

DefaultX-od:
fixes #511

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -64,6 +64,7 @@
   "deselectAll": "Deselect all",
   "update": "Update",
   "updateButton": "Update",
+  "refreshButton": "Check for updates",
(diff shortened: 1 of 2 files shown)


DefaultX-od:
The main goal is to create a new text line for refresh button on the updates page, so it could be translated differently from refresh button of a snap. That would solve the confusing situation explained in issue #511 . Please note that I don't have a strong preference for the text itself, so consider it as a placeholder.

PR #555, opened 2022-11-22
AppReviews: replace arrow font with real icons

Feichtmeier:
grafik

CC @Jupi007 if have time some time, maybe we can get thin yaru icons for the flag and those regular arrows with quil


Jupi007:
> CC @Jupi007 if have time some time, maybe we can get thin yaru icons for the flag and those regular arrows with quil

Ofc, I will do it 🙂

PR #569, opened 2022-11-25
Improve linting and code analysing

### LINKED ISSUE #568: Improve linting and code analysing
I think it would be nice to improve linting and code analysing by importing rules from `yaru_widgets`.
Some ones could be very interesting like `use_super_parameters`.

CC @Feichtmeier @jpnurmi

Jupi007:
I just ran `dart fix --apply`.
The only manual fix I did, is add `.call()` invocations on nullable, and replace `await` by `unawaited()` (only to the ones added by `dart fix`).

Fixes #568

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -12,3 +12,27 @@ linter:
     prefer_const_constructors: true
     prefer_const_declarations: true
     require_trailing_commas: true
+    always_declare_return_types: true
+    avoid_catches_without_on_clauses: true
+    avoid_equals_and_hash_code_on_mutable_classes: true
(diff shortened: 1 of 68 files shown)


jpnurmi:
I'll leave the decision to @Feichtmeier. :)

P.S. Perhaps this could be done in a bit smaller steps that are easier to digest? For example, start with the `use_super_parameters` alone because it's easy to agree on. :)

Jupi007:
Hmm I think you're right, looking at the diff, this is too huge.
So let's close and split this.

I propose to start with:

- `use_super_parameters`
- `sort_constructors_first`
- `sort_unnamed_constructors_first`

?

PR #571, opened 2022-11-26
One-to-one swipe back gesture for Apps page

### LINKED ISSUE #259: AppPage: use gesture detector for navigating back and forth with mouse keys
Don't know how these ones are called (mouse button 3 and 4, or 5?) but I use it frequently to navigate back and forth in browsers and file explorers. We need this in software too, should be possible when wrapping the whole app in a mouse listener

Edit: this needs to be implemented in flutter, currently only 1., 2. and 3. buttons are listenable
https://github.com/flutter/flutter/issues/115641

whiskeyPeak:
Fixes #259 

https://user-images.githubusercontent.com/73116038/204101612-3e812d82-f7fb-4d93-a68b-f767dd262462.mp4

Don't mind the flickering, that's an issue I'll report in a bit :sweat_smile:

### DIFF
--- a/lib/store_app/common/app_page/app_swipe_gesture.dart
+++ b/lib/store_app/common/app_page/app_swipe_gesture.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 import 'package:software/store_app/common/constants.dart';
+import 'package:yaru_icons/yaru_icons.dart';
 
 class BackGesture extends StatefulWidget {
   final Widget child;
@@ -80,19 +81,22 @@ class _BackGestureState extends State<BackGesture> {
             left: xPosition,
             child: Visibility(
               visible: isVisible,
-              child: Container(
-                decoration: const BoxDecoration(
-                  color: Color.fromRGBO(255, 127, 80, 1),
-                  shape: BoxShape.circle,
+              child: OutlinedButton(
+                onPressed: () {},
+                style: OutlinedButton.styleFrom(
+                  backgroundColor: Colors.grey[100],
+                  shape: RoundedRectangleBorder(
+                    borderRadius: BorderRadius.circular(width),
+                  ),
+                  minimumSize: Size(width, height),
                 ),
-                width: width,
-                height: height,
                 child: const Icon(
-                  Icons.arrow_back_rounded,
+                  YaruIcons.go_previous,
+                  size: 20,
                 ),
               ),
             ),
-          ),
+          )
         ],
       ),
     );
--- a/lib/store_app/common/app_page/app_swipe_gesture.dart
+++ b/lib/store_app/common/app_page/app_swipe_gesture.dart
@@ -45,13 +45,10 @@ class _BackGestureState extends State<BackGesture> {
     }
   }
 
-  void onPanStart(DragStartDetails details) {
+  void onPanStart(DragStartDetails details, BoxConstraints constraints) {
     currentExtent = 0;
     xPosition = 0 - _kButtonSize;
-    yPosition =
-        (MediaQueryData.fromWindow(WidgetsBinding.instance.window).size.height -
-                _kButtonSize) /
-            2;
+    yPosition = (constraints.maxHeight - _kButtonSize) / 2;
     setState(() {
       isVisible = true;
     });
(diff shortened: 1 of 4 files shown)


Jupi007 (commented):
Great work!

But I think that all the gesture detection logic should be moved to a separate widget.
This would avoid code duplication 🙂

Also, in term of design, I think that the button should look like a simple outlined button.
What do you think @madsrh?

whiskeyPeak:
> Great work!

Thank you!

> But I think that all the gesture detection logic should be moved to a separate widget. This would avoid code duplication slightly_smiling_face

Ah yes definitely! I'll create a new file in the app_page folder.

lib/store_app/common/app_page/app_swipe_gesture.dart
```diff
@@ -0,0 +1,99 @@
+import 'package:flutter/material.dart';
+import 'package:software/store_app/common/constants.dart';
+
+class BackGesture extends StatefulWidget {
+  final body;
```
Jupi007, line 5: I think this property should be named `child` to fit Flutter naming conventions.

whiskeyPeak: done!

whiskeyPeak:
> Also, in term of design, I think that the button should look like a simple outlined button. What do you think @madsrh?

The outlined button requires an onPressed argument, which has no real use here.

Jupi007:
> The outlined button requires an onPressed argument, which has no real use here.

Well, that's not a problem as you can use an empty callback `() {}` 🙂

Feichtmeier:
((( the flickering comes from vscode not using wayland if you start flutter debug/run )))

I would personally prefer a monochrome, outlined, OSD looking button, too
Thank you for this work

whiskeyPeak:
Screenshot from 2022-11-26 20-06-38

Something like this?

Jupi007:
> Something like this?

 Yes it is better, but I would prefer a circle button 🙂

whiskeyPeak:
> Yes it is better, but I would prefer a circle button slightly_smiling_face

Done!

> ((( the flickering comes from vscode not using wayland if you start flutter debug/run )))

Hmm, I'm currently running the app on the Cinnamon DE which uses X11 and run it via Android Studio. This is an issue I've also seen with the default flutter app template using the libhandy window package. The flickering mainly seems to occur when trackpad gestures are occurring. 

It also fails to draw the close, expand and minimise buttons in the title bar.
@Feichtmeier

Feichtmeier:
Hm that's a specific Linux mint issue then with either flutter or with flutter and libhandy let's take this out of the software repo then

Feichtmeier:
Could you show a screenshot of the button ?

whiskeyPeak:
light theme:

Screenshot from 2022-11-27 10-01-38


dark theme:

Screenshot from 2022-11-27 10-01-59

Jupi007:
Thanks for the screenshots @whiskeyPeak 

I feel that the button is not perfectly round. To fix that, you can remove the button padding.
Also, use a 20px icon is not ideal, because it looks a bit blurry.
They are designed to be displayed at 24px 🙂

whiskeyPeak:
Screenshot from 2022-11-27 10-29-39

> I feel that the button is not perfectly round. To fix that, you can remove the button padding.
Also, use a 20px icon is not ideal, because it looks a bit blurry.

Done

lib/store_app/common/app_page/app_swipe_gesture.dart
```diff
@@ -0,0 +1,108 @@
...
+                          ? Colors.grey[100]
+                          : Colors.grey[900],
+                  shape: RoundedRectangleBorder(
+                    borderRadius: BorderRadius.circular(width),
+                  ),
+                  minimumSize: Size(width, height),
+                ),
+                child: const Icon(
+                  YaruIcons.go_previous,
+                  size: 24,
```
Jupi007, line 99:
```diff
-                  YaruIcons.go_previous,
-                  size: 24,
+                  YaruIcons.go_previous
```
This can be removed, as `24` is the default size :)

whiskeyPeak: Oops. Fixed

lib/store_app/common/app_page/app_swipe_gesture.dart
```diff
@@ -0,0 +1,108 @@
...
+        xPosition -= -details.delta.dx * 0.2;
+      });
+    }
+  }
+
+  void onPanStart(DragStartDetails details) {
+    currentExtent = 0;
+    xPosition = 0 - width;
+    yPosition =
+        (MediaQueryData.fromWindow(WidgetsBinding.instance.window).size.height -
```
Jupi007, line 53: I think it would be better to use a `LayoutBuilder`, instead of a media query, to grab the widget height, and not the one of the full window.

Jupi007 (commented):
I added a few more comments :)

lib/store_app/common/constants.dart
```diff
@@ -17,6 +17,7 @@
 
 import 'package:flutter/material.dart';
 
+const kMaxExtent = 800.0;
```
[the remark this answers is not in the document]
whiskeyPeak: done

lib/store_app/common/app_page/app_swipe_gesture.dart
```diff
@@ -0,0 +1,108 @@
...
+    required this.child,
+  });
+
+  @override
+  State<BackGesture> createState() => _BackGestureState();
+}
+
+class _BackGestureState extends State<BackGesture> {
+  final double width = 50;
+  final double height = 50;
```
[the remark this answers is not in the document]
whiskeyPeak: done

lib/store_app/common/app_page/app_swipe_gesture.dart
whiskeyPeak: Does this change solve a particular problem though? I looked into the LayoutBuillder widget and I am having a hard time figuring out how to incorporate the constraints into the onPanStart function. IMO I'd also rather not add any unnecessary widgets we don't need in the build method.

Jupi007: You can do it like this:

```dart
void onPanStart(DragStartDetails details, BoxConstraints constraints) {
    // ...
}

// ...

@override
    return LayoutBuilder(
      builder: (context, constraints) {
        return GestureDetector(
          onPanStart: (details) => onPanStart(details, constraints),
          // ...
```

> Does this change solve a particular problem though?

I don't think the current situation could have a problem, but that's just a good practice to resolve that kind of thing in the context of the widget itself and not the whole window 🙂
And to be a bit perfectionist, use the layout builder makes the button perfectly centered in the page, because the full window height take the app bar in account 😉

Feichtmeier:
Update from main please

Jupi007:
@Feichtmeier Please, wait for my latest change comment to be resolved before to merge this 🙂

lib/store_app/common/app_page/app_swipe_gesture.dart
whiskeyPeak: Ahhh I see, thanks for the code! Done

whiskeyPeak:
> Update from main please

Done, PTAL!

Feichtmeier:
Tested locally! Awesome work @whiskeyPeak ! Thanks @Jupi007 for the review

Jupi007:
Thanks a lot @whiskeyPeak 👍

PR #573, opened 2022-11-27
Explorepage: selecting no package format gives error

### LINKED ISSUE #572: Explorepage: selecting no package format gives error
In the popup you can set a checkmark for either Snaps or Debs, but you can also select non which gives an error.
A solution would be to add a thrid option for both package formats and use radiobuttons or use a menu with a checkmark 🤷‍♂️ 

image
image

error

Feichtmeier:
Fixes #572


madsrh:
Ahhh, yes. That's also a solution 😄  TY

Issue #582, opened 2022-11-30
Speed up start-up time


Feichtmeier:
https://user-images.githubusercontent.com/15329494/204810430-bdd4f930-c022-46da-b5e7-f9e1b1f958bf.mp4

Waiting for the store model initi to finish takes now longer. 

_Originally posted by @Feichtmeier in https://github.com/ubuntu-flutter-community/software/issues/567#issuecomment-1332164770_

CC @oSoMoN 

we could maybe just deactivate update search on start-up for now?

10 seconds startup is max IMO

Feichtmeier:
IMO this is closed with #585
we can re-open if anyone is unhappy with the startup performance

PR #585, opened 2022-11-30
Do not await service init in the explore page

Feichtmeier:
- Do not await service init in the explore page
- show the app asap and do not show a splash screen
-  Also use a stream to update explore page and friends

This was taking too long.

Ref #582


Feichtmeier:
@oSoMoN any veto?

PR #587, opened 2022-12-01
Add animation to back gesture

whiskeyPeak:
This adds an animation to the back gesture, both when the page is popped and not.

https://user-images.githubusercontent.com/73116038/205141351-587f1a65-0d90-4bb9-8c2a-3fa435ca9ca9.mp4


whiskeyPeak:
I'd make an issue but I feel like this change is simple enough that it doesn't require any :sweat_smile:

lib/store_app/common/app_page/app_swipe_gesture.dart
```diff
@@ -58,8 +60,14 @@ class _BackGestureState extends State<BackGesture> {
...
-    setState(() {
-      isVisible = false;
+    Timer.periodic(const Duration(milliseconds: 1), (timer) {
+      setState(() {
+        xPosition -= 1;
+        if (xPosition <= 0 - _kButtonSize) {
+          isVisible = false;
+          timer.cancel();
+        }
+      });
```
Jupi007, line 70: Use a timer for this is really not ideal and you should use an animation controller instead with the `SingleTickerProviderStateMixin` mixin.
It would be manually piloted in `onPanStart/Update` function, and the controller should be reversed in `onPanEnd`.

Jupi007 (changes requested):
Thanks a lot @whiskeyPeak, animated stuff are always nice 😀

lib/store_app/common/app_page/app_swipe_gesture.dart
whiskeyPeak: Ok, will give this a look!

whiskeyPeak:
@Jupi007 Using an animation controller now. PTAL!

lib/store_app/common/app_page/app_swipe_gesture.dart
```diff
@@ -16,19 +16,42 @@ class BackGesture extends StatefulWidget {
...
+    super.initState();
+    swipeController = AnimationController(
+      duration: const Duration(milliseconds: 500),
+      vsync: this,
+    );
+    swipeController.addListener(() {
+      setState(() {
+        xPosition -= (xPosition + _kButtonSize) * swipeController.value;
+      });
+    });
```
Jupi007, line 41: Add `setState` in a listener is a bad practice, you should prefer use an `AnimatedBuilder`.

Jupi007 (changes requested):
Thanks @whiskeyPeak 🙂
I added another comment.

lib/store_app/common/app_page/app_swipe_gesture.dart
whiskeyPeak: I initially implemented it like that but the build method looked somewhat convoluted with all the different return statements. Will change back though

whiskeyPeak:
It works but looks pretty weird imo

Jupi007:
That you can do, is keep the same logic as in the previous iteration with the controler listener which change `xPosition`, but without `setState`. 🙂

lib/store_app/common/app_page/app_swipe_gesture.dart
```diff
@@ -16,19 +16,37 @@ class BackGesture extends StatefulWidget {
   State<BackGesture> createState() => _BackGestureState();
 }
 
-class _BackGestureState extends State<BackGesture> {
+class _BackGestureState extends State<BackGesture>
+    with SingleTickerProviderStateMixin {
+  late AnimationController swipeController;
```
Jupi007, line 21: As this controller only controls the back animation, I think it would be named `swipeBackController`.

lib/store_app/common/app_page/app_swipe_gesture.dart
```diff
@@ -16,19 +16,37 @@ class BackGesture extends StatefulWidget {
...
 
   bool isVisible = false;
 
+  @override
+  void initState() {
+    super.initState();
+    swipeController = AnimationController(
+      duration: const Duration(milliseconds: 500),
+      vsync: this,
+    );
```
Jupi007, line 36: Use an `AnimationController` value as it seems weird to me, I think you should store `xPosition` as an `Animation<double>` 🤔

lib/store_app/common/app_page/app_swipe_gesture.dart
whiskeyPeak: done

lib/store_app/common/app_page/app_swipe_gesture.dart
whiskeyPeak: I don't understand. Are you saying I should use a Tween to animate the button from kMaxExtent to 0 - kButtonSize instead of using xPosition -= (xPosition + _kButtonSize) * swipeBackController.value?

Jupi007: After really thought about it, I think just use the `AnimationController` is fine 🙂
The goal is just to animate the back gesture.

whiskeyPeak: Ok, done!

lib/store_app/common/app_page/app_swipe_gesture.dart
```diff
@@ -16,19 +16,40 @@ class BackGesture extends StatefulWidget {
...
+  @override
+  void initState() {
+    super.initState();
+    swipeBackController = AnimationController(
+      duration: const Duration(milliseconds: 500),
+      vsync: this,
+    );
+    swipeBackController.addListener(() {
+      xPosition -= (xPosition + _kButtonSize) * swipeBackController.value;
+    });
```
Jupi007, line 39: Just a little nitpick 😉, you can use this shorter syntax:
```
swipeBackController = AnimationController()..addListener();
```

Jupi007 (changes requested):
Just a little nitpick, and then LGFM 🙂

Jupi007 (approved):
LGFM 👍

Feichtmeier (approved):

Issue #609, opened 2022-12-04
use `context.select()` instead of `context.watch()`


Feichtmeier:
There's something fishy because `Theme.of(context)` should work fine in a grid builder. I noticed that the build method gets called _a lot_ on startup, though. It might help to split models into smaller entities and use `context.select()` instead of `context.watch()` where appropriate to optimize rebuilds.

_Originally posted by @jpnurmi in https://github.com/ubuntu-flutter-community/software/issues/608#issuecomment-1336367039_

Feichtmeier:
What shouldnt help are all those model calls everywhere @jpnurmi 

Later I started to remove the model calls from all the sub widgets so app_page is API agnostic (app_page is in a pretty good shape for example). I need to do this too for all snap_and_friends_widget and just pass all those callbacks and values

jpnurmi:
:+1: for small standalone model-agnostic widgets wherever it makes sense!

Feichtmeier:
@d-loose related to your recent help here, the explore header still uses context.watch. I tried to switch to read for functions + select for values but then the changes were not applied in the UI anymore. Could you check if you manage to move this from watch to select + read ? I think this should be one of the last components using watch only

d-loose:
Yes, will do :+1:

PR #610, opened 2022-12-04
ExplorePage: use select() to avoid excessive rebuilds

jpnurmi:
This fixes the issue that the whole ExplorePage was built ~20 times on startup. ExplorePage doesn't care about all the snap-related changes that ExploreModel notifies of most of the time.

Ref: #609


Feichtmeier (approved):
Thank you!

PR #612, opened 2022-12-04
SectionGrid: use select() to optimize rebuilds

jpnurmi:
On startup, the section grid view item builder gets now called once for each item, so 27 times. Before the select() changes, it was called 21x27=567 times.

Ref: #609


Feichtmeier:
holy crap ... thank you!

PR #643, opened 2022-12-15
Added network-manager plug

### LINKED ISSUE #620: On systems without PackageKit, hide deb UI
On a system without PackageKit, it would be nice if software would gracefully degrade and remove all the UI elements that aren't necessary if only snaps are used. This occurs on Core Desktop, which doesn't support packages other than snaps.

The existence of PackageKit can be detected by the `DBusServiceUnknownException` exception thrown when calling `PackageKit.Client.connect()`.

kenvandine:
Added network-manager plug, not needed to autoconnect but necessary for systems running network-manager as a snap.  Fixes #620

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -65,6 +65,7 @@ apps:
       - desktop
       - desktop-legacy
       - network
+      - network-manager
       - network-manager-observe
       - opengl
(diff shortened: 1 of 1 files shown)


d-loose (approved):
LGTM :+1:

PR #650, opened 2022-12-16
Handle absence of PackageKit more gracefully in the UI

### LINKED ISSUE #620: On systems without PackageKit, hide deb UI

d-loose:
- catch `DBusServiceUnknownException` on `_client.connect()`
(introduces `dbus` dependency, until `PackageKitClient` gets its own exception)
- add an `isAvailable` property to `PackageService`
- hide packageKit specific UI elements if the service isn't available

fix #620

### DIFF
--- a/lib/app/common/app_format_popup.dart
+++ b/lib/app/common/app_format_popup.dart
@@ -55,11 +55,13 @@ class AppFormatPopup extends StatelessWidget {
 class MultiAppFormatPopup extends StatelessWidget {
   const MultiAppFormatPopup({
     super.key,
-    required this.appFormats,
+    required this.selectedAppFormats,
+    required this.enabledAppFormats,
     required this.onTap,
   });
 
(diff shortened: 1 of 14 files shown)


Feichtmeier:
Thank you @d-loose for working on this!

Found one issue: when you de-select debian packages from the dropdown in the explore header the entry disappears instead of being de-selected

d-loose:
Thanks for spotting this - should be fixed now :)

Feichtmeier:
Thanks @d-loose ! Works without issues :partying_face:

PR #656, opened 2022-12-18
Add dark mode option in settings page

### LINKED ISSUE #623: Add a dark theme toggle and accent colour button in the settings page.
Some distros don't have a system-wide dark mode implementation (eg: linux mint cinnamon) so the software store ends up in light mode with no way to modify it.

We should add an appearance section in the settings page where those 2 options are customizable. I was thinking it looking somewhat similar to the ubuntu flutter settings app?

Screenshot from 2022-12-07 00-16-29

@Feichtmeier Can you assign this to me if it's a feature we want to implement? Thank you :D

whiskeyPeak:
Adds radio switches to change theme from system, light and dark in the settings page.

https://user-images.githubusercontent.com/73116038/208270629-cca4ab88-3471-4ebb-9048-7c59f4718429.mp4

Code is still a little messy, ~~and doesn't support localisation yet.~~ and still needs to use the SharedPreferences package to preserve the state of the theme when the app is closed and opened again.

Fixes #623


lib/app/settings/settings_page.dart
```diff
@@ -83,6 +85,91 @@ class _SettingsPageState extends State<SettingsPage> {
...
+      margin: const EdgeInsets.only(
+        left: kYaruPagePadding,
+        top: kYaruPagePadding,
+        right: kYaruPagePadding,
+      ),
+      headline: Text(context.l10n.theme),
+      child: Column(
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          for (var i = 0; i < 3; ++i)
```
Feichtmeier, line 148: i < themes.length or just for(final theme in themes)

Feichtmeier (commented):
Thank you for working on this!

Overall solid solution except the shared preferences thing. It's not bad I'm just not sure what way we should use to safe settings after app close because there should be more to safe and I'm interested a solution for that. For example the app format dropdown

@jpnurmi sorry for getting you in, but what's the best way to safe settings in Linux flutter apps? GSettings?

whiskeyPeak:
> Overall solid solution except the shared preferences thing. It's not bad I'm just not sure what way we should use to safe settings after app close because there should be more to safe and I'm interested a solution for that. For example the app format dropdown

Hmm, I see what you mean. If the Gsetting idea doesn't pan out, it might be a good idea to have a class with all of the given "saved" variables and have other classes call upon them statically.

kenvandine:
I would rather avoid GSettings, take a look at what @jpnurmi did for workshops.

whiskeyPeak:
I had a quick look at how the Workshops app does it and it uses the shared_preferences package. Should I start implementing the functionality or are there additional design choices we have to take into consideration? Thanks! @Feichtmeier

jpnurmi:
See [ubuntu-flutter-community/jsettings.dart](https://github.com/ubuntu-flutter-community/jsettings.dart) :slightly_smiling_face:

Feichtmeier:
> I had a quick look at how the Workshops app does it and it uses the shared_preferences package. Should I start implementing the functionality or are there additional design choices we have to take into consideration? Thanks! @Feichtmeier

It's good for now, you can create a jsettings solution in a different PR if you have time

PR #659, opened 2022-12-19
Add an empty Snap page if no Snaps are installed on the system

### LINKED ISSUE #657: On systems with no Snaps installed, the Installed Page will load indefinitely.
There should be a "No Snaps installed" message coming up if no Snaps are installed. We could maybe do the same for the deb page? :grin:

https://user-images.githubusercontent.com/73116038/208285925-3dcdd2b9-1a48-4460-bcd4-28fbc08082fe.mp4

whiskeyPeak:
Added a utility function that returns true if the querying of snap packages is done and false if it isn't. The linked issue shows how the UI looked previously.

Also, I'm not really sure as to which icon to use to indicate that there are no Snap apps installed. Any ideas? @Jupi007 

https://user-images.githubusercontent.com/73116038/208337328-bf708d71-3ef1-49ef-ab63-b1fd193e469e.mp4

Fixes #657


Feichtmeier:
Very good idea - just maybe we could try to ask snapd.dart in the snap_service if snap is installed? If possible I would prefer this over just the timeout = snap is not installed thing

Jupi007:
> Also, I'm not really sure as to which icon to use to indicate that there are no Snap apps installed. Any ideas? @Jupi007

This would require a new "no snap" icon.
I will create one 🙂

whiskeyPeak:
> Very good idea - just maybe we could try to ask snapd.dart in the snap_service if snap is installed? If possible I would prefer this over just the timeout = snap is not installed thing

Will do!

> This would require a new "no snap" icon.
> I will create one slightly_smiling_face

Awesome, thanks :D

whiskeyPeak:
> Very good idea - just maybe we could try to ask snapd.dart in the snap_service if snap is installed? If possible I would prefer this over just the timeout = snap is not installed thing

So I gave this a further look and there is an edge case where a user has snapd installed on their systems without any snap applications. In this case, we'd fall right back into the original issue because the connection would be endlessly true. wdyt? @Feichtmeier

whiskeyPeak:
The software manager can currently be on 3 states: 
- no snapd installed
- snapd installed but with no Snaps installed
- Snaps installed

When snapd is not installed on the system, a "snapd is not installed" message will be displayed in the explore and installed pages for snaps. You can still search for deb packages if you select them via the popup menu.

https://user-images.githubusercontent.com/73116038/209433466-c2f50c21-a903-4d5d-aec7-0da10221bd66.mp4

When snapd is installed but without any Snaps, the installed page shows "No Snaps installed". The explore page works as ususal.

https://user-images.githubusercontent.com/73116038/209433510-e9eb9697-5fba-4aec-a0df-c58c7b26086b.mp4

When Snaps are installed, everything works as usual. 

https://user-images.githubusercontent.com/73116038/209433461-599dbc18-ed36-4b11-ba9c-aca7e12828fe.mp4

I'm currently wondering if we should make debian packages the default in the installed and explore page if snapd is not installed, since a lot of debian based distros don't make use of Snaps (and so don't even have snapd installed).

What do you think? @Feichtmeier @Jupi007 @madsrh

Jupi007:
This looks very good @whiskeyPeak :+1: 

I'm currently working on the icons.

> Also, Merry Christmas! :D

Merry Christmas too :D

Jupi007:
@whiskeyPeak Icons are ready: https://github.com/ubuntu/yaru_icons.dart/pull/112

whiskeyPeak:
> @whiskeyPeak Icons are ready: [ubuntu/yaru_icons.dart#112](https://github.com/ubuntu/yaru_icons.dart/pull/112)
> 
> Imo, once this is published, you should use `snapcraft_error` when snapd is missing, and `no_package_snap` icon when there is no snap installed.

Nice! Thanks for the work :D

madsrh:
Excellent work @whiskeyPeak 🎉 

In the first screen recording:

> When snapd is not installed on the system, a "snapd is not installed" message will be displayed in the explore and installed pages for snaps. **You can still search for deb packages if you select them via the popup menu.**

But if only snap packages are selected but snapd isn't installed, like in the video, it shouldn't say "we couldn't find...", then we want to disable the search field or just keep the "Snapd isn't installed" message IMO.

> I'm currently wondering if we should make debian packages the default in the installed and explore page if snapd is not installed,...

👍 Sound like a good idea **IF** snapd isn't installed. 🤔 One could argue that if Canonical wants more people to use snaps, it would make sense to expose them to the "You don't have snapd installed, click here..." page first, but I would find that a bit annoying on a pure deb system 😁 

> Maybe add a link of how install snapd

ofc 💯

Feichtmeier:
Thanks @whiskeyPeak 

Maybe you missed my review comment, but I think we aren't done yet

grafik

Feichtmeier:
One additional thing:

- do distros exist that have the snap-store installed but not snapd? There is not release of this app outside of the snap-store and it is specifically designed to work with snaps. It also works better with snaps because the API is richer and more specific

whiskeyPeak:
> Thanks @whiskeyPeak
> 
> Maybe you missed my review comment, but I think we aren't done yet
> 
> grafik

Huh weird, I don't see that message on the TL.
The code paths that lead to the _getSync methods will always throw a SocketException if snapd is not installed because there is nothing to facilitate a connection with. More throw statements could be added in the snapd.dart library to make absolutely sure that we're looking at the correct exception thought :D. 
I tested with my machine not being connected to the internet but with snapd installed and still didn't receive a SocketException so I'm quite confident that those are correlated. 

I could move the socket exception further down the tree to reduce the likelihood of something else creating a socket exception, though a better fix would be to add a throw statement in the _getSync function (which resides in the snapd.dart package).

> do distros exist that have the snap-store installed but not snapd? There is not release of this app outside of the snap-store and it is specifically designed to work with snaps. It also works better with snaps because the API is richer and more specific

I think this change is more in line with users (linux mint, debian) who'd download the software centre via github, so they wouldn't have either snapd nor the snap-store installed.

> But only if snap packages are selected but snapd isn't installed, like in the video, it shouldn't say "we couldn't find..." when searching, then we want to disable the search field or just keep the "Snapd isn't installed" message IMO.

Fixed, now the "snapd is not installed" message is always shown.

> +1 Sound like a good idea IF snapd isn't installed. thinking One could argue that if Canonical wants more people to use snaps, it would make sense to expose them to the "You don't have snapd installed, click here..." page first, but I would find that a bit annoying on a pure deb system grin

Yeah, I think adding the link is a good idea. I was just wondering how much we'd want to push for the user to install snapd on their system vs allowing pure debian users to use the store as well.

Feichtmeier:
> I think this change is more in line with users (linux mint, debian) who'd download the software centre via github, so they wouldn't have either snapd nor the snap-store installed.

I don't do github release extracts anymore. The snap-store snap is the only release we do

> I tested with my machine not being connected to the internet but with snapd installed and still didn't receive a SocketException so I'm quite confident that those are correlated.

Sure, but in general a socket exception can be anything. So this assumption can be wrong "SocketException" -> snapd is not installed 😸 
DOn't get me wrong, thankful for this work and totally pro for correctly displaying issues in the UI, but we should make it really precise

whiskeyPeak:
> Sure, but in general a socket exception can be anything. So this assumption can be wrong "SocketException" -> snapd is not installed 😸 
> DOn't get me wrong, thankful for this work and totally pro for correctly displaying issues in the UI, but we should make it really precise

Nw, completely understand what you mean! I'll get back to this once I find a better solution for the snapd problem. I'll close this pr in the meantime.

Feichtmeier:
Ok 👌 😊
Maybe we can talk to Robert on the snapd.dart repo what he thinks it's the best for verifying snapd is installed or not installed?

whiskeyPeak:
Alternatively, we could run a bash command via dart to determine whether snapd is installed on the system?

```
import 'dart:io';

void main() {
    Process.run('bash', ['-c', 'dpkg -l | grep snapd']).then((result) {
        stdout.write(result.stdout);
    });    

}
```
Parse the result into a string and compare the first 2 char to whatever characters we want: https://askubuntu.com/questions/18804/what-do-the-various-dpkg-flags-like-ii-rc-mean

Issue #561, opened 2022-11-23
Add an option to return to the other packageformat if it exists from the Snap/Packagepage


Feichtmeier:
Even if I do not like this idea semantically, I see no other room on the pages other than the dropdown for SnapPage

![grafik](https://user-images.githubusercontent.com/15329494/203487729-0865849f-66a3-47dd-9d3d-463e7aca7203.png)

For PackagePage a simple button "Switch to Snap" would be enough
![grafik](https://user-images.githubusercontent.com/15329494/203488051-19901776-9ac1-40ef-87bb-ffbc291d84ba.png)

madsrh:
Do you mean like this? or perhaps [with a label](https://user-images.githubusercontent.com/3986894/203498254-55123dbf-eea7-499f-a329-01df0f852601.jpg)

![change-format](https://user-images.githubusercontent.com/3986894/203498241-42d85f2a-d7bf-4f89-bcae-f0653650b752.jpg)

If possible, I think it would be best no to have a button on one page and a popup/dropdown on the other. Can we use a popup/dropdown on the PackagePage too?

![image](https://user-images.githubusercontent.com/3986894/203498164-6e431e13-1764-41bb-ba3f-bd9f4de2f1fb.png)

madsrh:
> Or just a normal button that forwards you to the deb page. So you can keep the navigation intact

Toggle buttons / [segmented buttons](https://m3.material.io/components/segmented-buttons/overview)

![image](https://user-images.githubusercontent.com/3986894/203848331-8681c9f5-7def-4538-88ea-ae17b025b443.png)

Feichtmeier:
Woops, it's not really closed unless you guys say so, but the functionality is there. But I started with an outlined button because this is much easier with the navigator because it is now

banner > snap + deb > click > snap page first > switch to deb button > deb page
                                                                                   
then
                               
                                                 appbanner  <   snap page   <    back button

Feichtmeier:
![grafik](https://user-images.githubusercontent.com/15329494/207702281-210b3f71-93fb-4c73-a2d7-10f795370a9f.png)

anasereijo:
> ![grafik](https://user-images.githubusercontent.com/15329494/207702281-210b3f71-93fb-4c73-a2d7-10f795370a9f.png)

[...]

PR #661, opened 2022-12-19
AppPage: add appformat popup

Feichtmeier:
https://user-images.githubusercontent.com/15329494/208445340-c05db3ec-749a-4510-a9d5-c386d35fef13.mp4

Ref #561


anasereijo:
I'm surprised to see the deb option back to the dropdown.
The option to switch between a snap to a deb shouldn't be hidden on that dropdown. 

First, that dropdown is related to snap channels, so we are hiding the deb option in a dropdown that is not related to the type of packages. It has to be clear to the user that they can install a snap and/or a deb (if both available).

We really should try to implement the switch button (or a similar solution if that doesn't work). This option has to be clear and well presented to the user.

Feichtmeier:
This is a very old pull request Ana ☺️

As I said, I'll add the button soon

PR #669, opened 2022-12-21
Add star developer icon for app banners

Feichtmeier:
grafik

### DIFF
--- a/lib/app/common/app_banner.dart
+++ b/lib/app/common/app_banner.dart
@@ -164,6 +164,25 @@ class SearchBannerSubtitle extends StatelessWidget {
                   color: light ? kGreenLight : kGreenDark,
                   size: 12,
                 ),
+              ),
+            if (appFinding.snap?.starredDeveloper == true)
+              Padding(
+                padding: const EdgeInsets.only(left: 5),
+                child: Stack(
+                  alignment: Alignment.center,
+                  children: const [
+                    Icon(
+                      Icons.circle,
+                      color: Colors.white,
+                      size: 9,
+                    ),
+                    Icon(
+                      Icons.stars,
+                      color: kStarDevColor,
+                      size: 12,
+                    ),
+                  ],
+                ),
               )
           ],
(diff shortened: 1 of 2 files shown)


anasereijo:
@Zoospora could you please have a look and check if the icon is the correct one and the overall design of it? thank you!

Zoospora:
Hey, of course!
I see this is shaping up really well. Congratulations on your hard work! 🤩

- First of all, I would like to say that the design team is still working on how to improve the understanding and visualisation of the different publishers. We hope to come up with some useful proposals. 

- Regarding the icons, I would like to mention that we at Vanilla tend to align them to the top line (I've attached a screenshot from Snapcraft) instead of centring them.

image
([https://snapcraft.io/search?category=development&page=2](https://snapcraft.io/search?category=development&page=2))

- I think the star idea is great, it helps to maintain consistency between products. On the other hand, I think that perhaps we should work on the orange colour so that it is accessible on the dark background. 

Let me explain: if you look at the image I have attached, the star is clearly visible on the white background, but on the dark background it disappears a little and it is difficult to distinguish it. In this respect, I have two different ideas:
1. Look for an orange within the range that makes a better contrast on the dark tone (creating an alternative version of the icon).
2.	Change the background colour of the cards. You could perhaps try the same background colour as the application (I don't know how it would look, it's just an exploration). 

- As for the texts, I thought I saw (correct me if I'm wrong) that in the default theme (white background), the text of the card description is quite light grey. Don't you think it would work better with a black text colour or whatever dark grey you use? Checking in Yaru's guide they also use a darker colour for the bodyText.

image
([https://ubuntu.github.io/yaru.dart/#/](https://ubuntu.github.io/yaru.dart/#/))

Again, these are all suggestions to what is already a great job! 
Thank you.

anasereijo:
@Zoospora this is how it looks with the light theme:

VirtualBox_Ubuntu Linux Clone_04_01_2023_11_59_08

Issue #663, opened 2022-12-19
AppStream service initialization slows down search


Feichtmeier:
Currently the AppstreamService is loaded asynchronously and without awaiting the future inside app model init. 
The page loads but the appstream service is not ready yet, slowing down the initial search a lot (after this initial lag, everything is fine though) :cry: 

https://user-images.githubusercontent.com/15329494/208499167-759465cc-558e-440d-bf6e-fd87f22d3d32.mp4

________

## Solution A - await AppStreamService.init (10 seconds splash screen on EVERY start):

https://user-images.githubusercontent.com/15329494/208498539-d133ff14-a5e1-4089-a50c-146c9db19113.mp4

## Solution B - do not select select debian packages initially as selected AppFormats:

https://user-images.githubusercontent.com/15329494/208497825-66d5aa3c-919d-4ec6-863c-e5b8a1fee7ef.mp4

## Solution C - make appstream service init faster (lol) and await the service

@jpnurmi @d-loose any ideas?

d-loose:
In principle I like Solution B, but it would be nice if debian packages were automatically added to the selected `AppFormat`s (and search results updated accordingly) as soon as `AppstreamService` is done with its initialization.
Since it takes quite long there should definitely be some indication to the user.

Semi-related: If you start typing a search query quickly enough, you end up on the `SearchPage` before `SnapService` has finished loading its sections and the page gets redrawn every time it loads the next one (causing the flickering in your screen recording).

Feichtmeier:
@d-loose hm hm, now the store starts, but then it takes 10-15 seconds to finish the UI because it is still waiting for the appstream service :cry: 


https://user-images.githubusercontent.com/15329494/210669579-3ad421c0-b410-4eee-a743-0c931b43dfac.mp4

Feichtmeier:
But this made me think of the whole page flow again since on the start page are only snaps anyways. I open a discussion

d-loose:
The problem is that `ExploreModel` is now initialized before `PackageService` finishes connecting to `PackageKitClient`. Since #650 `PackageService` provides a `serviceAvailable` attribute, which defaults to `false` and is asynchronously updated if the connection can be established without errors. But it currently lacks a way of communicating that its availability has changed.

Feichtmeier:
Ok I try to disconnected the pages more from each other and ping back when I am done?
We are currently also discussing a different page flow which needs a navigator

d-loose:
Ok, sounds good.
[...]

PR #683, opened 2023-01-03
Improve Search/Explore page initialization

d-loose:
* add `AppFormat.packageKit` to enabled and selected formats after `AppStreamService` is done with its initialization
* only subscribe to  `SnapService.sectionsChanged` in `ExploreModel` after `SnapService` is done with its initalization (reduces early page rebuilds)

https://user-images.githubusercontent.com/113362648/210390478-5936c552-d632-41bd-9711-ef6e19c53393.mp4

Any suggestions on how to indicate that `AppStreamService` is still loading in the background?

ref #663

### DIFF
--- a/lib/app/explore/explore_model.dart
+++ b/lib/app/explore/explore_model.dart
@@ -49,7 +49,7 @@ class ExploreModel extends SafeChangeNotifier {
     );
 
     if (_packageService.isAvailable) {
-      _appstreamService.init().then((value) {
+      _appstreamService.init().then((_) {
         _enabledAppFormats.add(AppFormat.packageKit);
         _selectedAppFormats.add(AppFormat.packageKit);
         notifyListeners();
--- a/lib/app/explore/explore_model.dart
+++ b/lib/app/explore/explore_model.dart
@@ -42,9 +42,11 @@ class ExploreModel extends SafeChangeNotifier {
 
   Future<void> init() async {
     _enabledAppFormats.add(AppFormat.snap);
-    _sectionsChangedSub =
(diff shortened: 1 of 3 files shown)


lib/app/explore/explore_model.dart
```diff
@@ -42,19 +42,23 @@ class ExploreModel extends SafeChangeNotifier {
...
-        _snapService.sectionsChanged.listen((_) => notifyListeners());
+    _selectedAppFormats.add(AppFormat.snap);
+    _snapService.initialized.then(
+      (value) => _sectionsChangedSub =
+          _snapService.sectionsChanged.listen((_) => notifyListeners()),
+    );
+
     if (_packageService.isAvailable) {
-      _enabledAppFormats.add(AppFormat.packageKit);
+      _appstreamService.init().then((value) {
```
d-loose, line 52: Since `init()` simply returns a Future from a private attribute it can be called again here, instead of adding an `initialized` property. Is this misleading/confusing?

Feichtmeier: @d-loose thanks for this, sorry for the late reply, was a bit offline

the important part is that the snap service is only initialized once in the app lifecycle so the list/maps of snaps found by section are stored, same with the installed apps etc
the rest is a matter of taste maybe :)

nitpick: i think you can remove the "value" from then : D and replace it with _

Feichtmeier (approved):

d-loose:
Thanks!
I fully support spending time offline :)

Feichtmeier:
No idea what is causing those integration test failures on both of your PRs but it seems unrelated to what you changed to me

PR #690, opened 2023-01-04
Improve hit behaviour of back gesture for PanedLayout page

whiskeyPeak:
Back gesture isn't triggered when the cursor is positioned in places where there is padding. This PR fixes that and also removes a redundant centre widget.

I should really add epileptic warnings before each of the videos I post here lol.

https://user-images.githubusercontent.com/73116038/210658814-9826b9f0-a9f9-4a81-853a-16ef3c7a4f25.mp4

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,4 +30,4 @@ jobs:
           sudo touch /etc/NetworkManager/conf.d/10-globally-managed-devices.conf
           sudo systemctl restart NetworkManager.service
       - run: sudo cp integration_test/assets/packagekit-ci.pkla /var/lib/polkit-1/localauthority/50-local.d/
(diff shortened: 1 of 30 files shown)


Feichtmeier:
> I should really add epileptic warnings before each of the videos I post here lol.

XD

tip: make the screen recording with OBS (snap works 🤷 )

Did you check if the center widget was needed for maybe different screen sizes? SOrry it is so long ago I worked on this 👴

whiskeyPeak:
> tip: make the screen recording with OBS (snap works shrug )

Ahh, I have the flickering regardless of whether I screen record or not. I'm starting to think it's a linux mint related issue.
> Did you check if the center widget was needed for maybe different screen sizes? SOrry it is so long ago I worked on this older_man

As things are right now, the padding on either side of the page for all of the layouts is always going to be the same. Same with the top and bottom padding, so the pages are all already centred. I tested all 3 layouts and they looked the exact same as before :).

PR #692, opened 2023-01-05
fix integration test

### LINKED ISSUE #627: PackageInstaller: fix integration test
Fix integration test after the package installer has moved into the now one and only software app with https://github.com/ubuntu-flutter-community/software/pull/625

The procedure to test should be the following:

- check that /usr/bin/hello does not exist yet
- load main with given path
- main with given path loads initial index 3, which is a package page with given path
- find the install button
- tap the install button
- /usr/bin/hello should exist now
- the install button is exhanged with a remove button
- find the remove button
- tap the remove button
- the install button should show up again
- /usr/bin/hello should not exist

I have problems with the pump method which produces changing results. Sometimes the app just crashes, then the install button is not found, then some overflow happens which does not happen in the real app

CC @jpnurmi

d-loose:
~~WIP~~

* use `pumpAndSettle` to wait until the UI is ready
* tap install button
* wait until uninstall button is present and enabled
* tap uninstall button
* wait until uninstall button is gone
* use `en_US.UTF-8` locale to avoid problems in the CI

fix #627

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,4 +30,4 @@ jobs:
           sudo touch /etc/NetworkManager/conf.d/10-globally-managed-devices.conf
           sudo systemctl restart NetworkManager.service
       - run: sudo cp integration_test/assets/packagekit-ci.pkla /var/lib/polkit-1/localauthority/50-local.d/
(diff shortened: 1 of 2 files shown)


Feichtmeier:
Awesome! :clinking_glasses:  Thank you very much! Was freaking me out, it worked locally but not in the CI

PR #712, opened 2023-01-08
Fix indefinite installed page

### LINKED ISSUE #657: On systems with no Snaps installed, the Installed Page will load indefinitely.

whiskeyPeak:
Doesn't use localisation until we figure out what kind of message we want to show. :D 

https://user-images.githubusercontent.com/73116038/211175419-871f268b-c7d4-46c2-b903-d92f6e937ec2.mp4

fixes #657


Feichtmeier (changes requested):

whiskeyPeak:
I was thinking that we could maybe rename it to "ResponsePage" and have it used wherever we should an empty page to the user with only an icon and message. This widget could be used:
- Explore page when the search can't find any correlated snaps/debs
- Installed page when no snaps are installed
- Installed page when no updates are available
- Update page when no updates are available

What do you think? @Feichtmeier

Feichtmeier:
Reading through the code again and thinking about it, this is actually not an "error" so ErrorPage does not really fit here. It is just "No snaps are installed". please change the name of the page to something fitting and move it to the "snaps" directory. The message can be inside the en.arb file with something like

"noSnapsInstalled": "There are no snap applications installed on your system."

lib/app/explore/start_page.dart
```diff
@@ -88,6 +89,9 @@ class _StartPageState extends State<StartPage> {
         controller: _controller,
         child: Column(
           children: const [
+            SizedBox(
+              height: kPagePadding - 5,
```
Feichtmeier, line 93: whats that? :)

Feichtmeier (changes requested):

lib/app/explore/start_page.dart
Feichtmeier: never mind, you merged master after and the diff was gone

Feichtmeier (approved):

lib/app/explore/start_page.dart
whiskeyPeak: Ahh yeah, I was wondering why it kept showing up lol

PR #714, opened 2023-01-08
Check for empty snapshot list

### LINKED ISSUE #713: Installed page shows empty page when update icon is selected
This is because the if statement only checks whether the snapshot is not null, so an empty list would be ignored here.

https://user-images.githubusercontent.com/73116038/211177321-a37e3ad7-1c52-47b7-bab5-0e5769b563a0.mp4

whiskeyPeak:
I was also getting some layout exceptions because of the Expanded widget so I removed all that logic. The page looks the exact same as before in all 3 layouts.

@Feichtmeier lmk if I'm missing something though!

https://user-images.githubusercontent.com/73116038/211177448-87cf6aed-623d-4957-9d71-892d50a35439.mp4

Fixes #713

### DIFF
--- a/lib/app/updates/no_updates_page.dart
+++ b/lib/app/updates/no_updates_page.dart
@@ -6,47 +6,31 @@ import 'package:yaru_icons/yaru_icons.dart';
 class NoUpdatesPage extends StatelessWidget {
   const NoUpdatesPage({
     Key? key,
-    this.expand = true,
   }) : super(key: key);
-
-  final bool expand;
 
   @override
   Widget build(BuildContext context) {
-    final page = Center(
-      child: Row(
-        mainAxisSize: MainAxisSize.min,
-        children: [
-          Column(
-            mainAxisSize: MainAxisSize.min,
-            children: [
-              YaruAnimatedOkIcon(
-                size: 90,
-                filled: true,
-                color: Theme.of(context).brightness == Brightness.light
-                    ? kGreenLight
-                    : kGreenDark,
-              ),
-              const SizedBox(
-                height: 20,
-              ),
-              Text(
-                context.l10n.noUpdates,
-                style: Theme.of(context).textTheme.headlineMedium,
-              ),
-              const SizedBox(
-                height: 100,
-              ),
-            ],
-          ),
-        ],
-      ),
+    return Column(
+      mainAxisSize: MainAxisSize.min,
+      children: [
+        YaruAnimatedOkIcon(
+          size: 90,
+          filled: true,
+          color: Theme.of(context).brightness == Brightness.light
+              ? kGreenLight
+              : kGreenDark,
+        ),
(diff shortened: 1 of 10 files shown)


Feichtmeier:
I'm pretty sure we need the expanded here to let the grid always get the full height. I'll double check later

Feichtmeier:
@whiskeyPeak there is indeed a bit inconsitency with expanded here and there, I think it would be more transparent for re-use if the expanded is around a widget and not inside. Maybe some artefacts from moving stuff around too much. Do you mind if I fix this in a different PR, you remove the expanded changes here and just focus on the no updates screen?
In general it is better to focus on one task in a PR - I am often lazy myself I know this feeling - but it is better to make another PR then :smile:

lib/app/updates/no_updates_page.dart
```diff
@@ -6,14 +6,11 @@ import 'package:yaru_icons/yaru_icons.dart';
...
     Key? key,
-    this.expand = true,
   }) : super(key: key);
 
-  final bool expand;
-
   @override
   Widget build(BuildContext context) {
-    final page = Center(
+    return Center(
```
Feichtmeier, line 13: Please remove the changes for "expanded", I fix this everywhere in a different PR

Feichtmeier (changes requested):

lib/app/updates/no_updates_page.dart
whiskeyPeak: Done!

whiskeyPeak:
After the latest rebase, i don't get the exception anymore but the message and icon aren't centred anymore.

Screenshot from 2023-01-08 16-10-24

Feichtmeier:
@whiskeyPeak wrap it inside Expanded -> Center -> Singlechildscrollview

whiskeyPeak:
I think it makes more sense to have the centre widget be part of the noUpdatesPage (we always want to the icon and message to be centred), have the singleChildScrollView be wrapped outside in cases where there's pixel overflow and just not use the expanded widget (i get a bunch of exceptions when it's wrapped around centre).

Feichtmeier:
those pages are sometimes part of a column and sometimes not. Thus sometimes they are expanded and sometimes not. Those scroll windows are needed for those tiny winy window sizes, the center is needed AROUND the scrolling window though. So we keep the noupates/updating pages are minimal as possible and leave it to the outer pages to place them. Thank you

Feichtmeier (approved):

whiskeyPeak:
> those pages are sometimes part of a column and sometimes not. Thus sometimes they are expanded and sometimes not. Those scroll windows are needed for those tiny winy window sizes, the center is needed AROUND the scrolling window though. So we keep the noupates/updating pages are minimal as possible and leave it to the outer pages to place them. Thank you

Yeah, I see what you mean. Done!

PR #721, opened 2023-01-09
ExploreHeader: use context.select

d-loose:
Use `context.select` in `ExploreHeader` to reduce unnecessary rebuilds.
Ref #609

### DIFF
--- a/lib/app/explore/explore_header.dart
+++ b/lib/app/explore/explore_header.dart
@@ -11,7 +11,13 @@ class ExploreHeader extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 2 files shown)


d-loose:
There are some subtleties that I also wasn't completely aware of (until now): in order for `context.select` to correctly cause a rebuild the selected value must be immutable - that's why I changed the getters to return a copy of the sets.

Feichtmeier:
aaah! this is specific for collections? Because elsewhere it worked

Feichtmeier:
I think I did not test this thoroughly enough. The selected snap category change does not rebuild the page anymore. I am not sure if this change was the reason though

PR #745, opened 2023-01-11
YaruWindowTitlebar: big layout changes

Feichtmeier:
- use yaruwindowtitlebar to save vertical screenspace
- adapt the layout to use yarumasterdetailpage
- add the searchbar to the titlebar
- adapt all pages containing an appbar to use yaruwindowtitlebar

grafik
grafik
grafik
grafik

grafik
grafik
grafik

### DIFF
--- a/lib/app/app.dart
+++ b/lib/app/app.dart
@@ -197,6 +197,7 @@ class __AppState extends State<_App> {
 
     return _initialized
         ? YaruMasterDetailPage(
+            layoutDelegate: const YaruMasterFixedPaneDelegate(paneWidth: 220),
             appBar: const YaruWindowTitleBar(
               isClosable: false,
               isMaximizable: false,
--- a/lib/app/app.dart
+++ b/lib/app/app.dart
@@ -146,13 +146,6 @@ class __AppState extends State<_App> {
   Widget build(BuildContext context) {
     final model = context.watch<AppModel>();
     model.setupNotifications(updatesAvailable: context.l10n.updateAvailable);
-    final width = MediaQuery.of(context).size.width;
-
-    final itemStyle = width > 800 && width < 1200
-        ? YaruNavigationRailStyle.labelled
-        : width > 1200
-            ? YaruNavigationRailStyle.labelledExtended
-            : YaruNavigationRailStyle.compact;
 
     final pageItems = [
       PageItem(
@@ -203,15 +196,20 @@ class __AppState extends State<_App> {
     ];
 
     return _initialized
-        ? YaruNavigationPage(
+        ? YaruMasterDetailPage(
+            appBar: const YaruWindowTitleBar(
+              isClosable: false,
+              isMaximizable: false,
+              isMinimizable: false,
+              isRestorable: false,
+            ),
             key: ValueKey(path),
             length: pageItems.length,
             initialIndex: _initialIndex,
-            itemBuilder: (context, index, selected) => YaruNavigationRailItem(
-              icon: pageItems[index].iconBuilder(context, selected),
-              label: pageItems[index].titleBuilder(context),
+            tileBuilder: (context, index, selected) => YaruMasterTile(
+              leading: pageItems[index].iconBuilder(context, selected),
+              title: pageItems[index].titleBuilder(context),
               // tooltip: pageItems[index].tooltipMessage,
-              style: itemStyle,
             ),
             pageBuilder: (context, index) => pageItems[index].builder(context),
           )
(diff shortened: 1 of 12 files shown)


Feichtmeier:
@d-loose test succeeds locally, no idea whats the issue here, I create a new issue for the test and merge this one

Feichtmeier:
I need to fix the layout for small windows first before this goes live, even if this looks too good to be true

anasereijo:
I don't think this was what @madsrh proposed (please correct me if I am wrong).

I really don't like seeing the search in the window bar. Makes it more confusing and visually unbalanced. Also, I don't think it saves that much space anyway. 
Why don't we try what @madsrh proposed instead?

anasereijo:
> I need to fix the layout for small windows first before this goes live, even if this looks too good to be true

Big changes like this shouldn't be pushed live, until it's get a green light from us all. Unless you need to push it live to test it?

Feichtmeier:
> > I need to fix the layout for small windows first before this goes live, even if this looks too good to be true
> 
> Big changes like this shouldn't be pushed live until it's get a green light from us all. Unless you need to push it live to test it?

Clickable titles are a very bad idea I am deeply sorry. I agree that this shouldn't go live until we find a design we can at least agree on though. This is not a browser, and no one ever made clickable titles in any app I have ever seen.
What I think can work is that we always show the searchbar, as I proposed in various issues and discussions

Feichtmeier:
Also as a note, even if mads proposed this I had a strong veto from the meeting to the discussion. Titles should *not* be something you click and then change to a searchbar. This is not a browser. It is misleading and ambiguous. What works is (as I said several times) that we always show a search bar instead.

This page navigation had the following idea:

Master | Detail

entry
entry
entry

the detail also shows on which page you are on. INSIDE the title. So you can't just exchange the title at some point while you are on the page. 

If something makes no sense, I think I have the best right to say it, since I spent almost half a year on designing and then implementing this application.

anasereijo:
Whatever the idea is, it needs proper (design) thinking as it's quite a big change in functionality and experience. I don't think this is a priority now, so I will be focused on moving on with the other pages. Of course, if @madsrh wants to explore some design options for this, he is more than welcome!

Thanks for trying this, though!

anasereijo:
> If something makes no sense, I think I have the best right to say it, since I spent almost half a year on designing and then implementing this application.

Of course you have all the right to express your opinion, @Feichtmeier. I don't think anyone said, or expressed, the opposite. The same applies to me, or to anyone who wants to get involved in this project. We all have different backgrounds and skills, and we are working towards the same goal: producing a new store with the best experience. 

I understand it can be frustrating sometimes to change things that have been implemented. This project is quite tricky because implementation started before UX and Visual Design, that's why it has been a lot of back and forth, but we have been taking that into consideration and we really appreciate your efforts and commitment.

Feichtmeier:
First of all, even if you might not  have notice, there was an additional titlebar always showing on top of the window taking up 50 pixel of vertical screenspace for nothing.

Thanks to jp we now have the option in flutter applications on linux to have CSD window controls inside the window without an additional titlebar on top.
This leads to the option to have widgets *inside* titlebars.
This leads to a design where you could show a searchbar always and search from anywhere to anywhere (not implemented).

What you two proposed was to make the title clickable like in a browser, which's result would be similar yet incredible easy to misunderstand.

This PR does not change much at the main navigation for now but saves vertical screen space and adapts this application to the rest of modern GTK4 (the other tooltkit on ubuntu). 

We can now work on simplifying the page flow and the entry point of the app.

We previously mixed up the "discover" "explore" experience windows11 and macOS succeed in, with a search experience. Let's please try to get this working now and in my opinion the next priority should be to get the discover / explore / start / entry / landing / however-you-call-the-first-page-after-start page to make it look eye-catching and like in a magazin. And not just show a list of categories.

Feichtmeier:
> implementation started before UX and Visual Design

not the design was happening before the first implementation by myself. Might be a weird situation for designers, but I am designing and developing. I have even read design books. Yeah, kinda rare, right

d-loose:
I think I found the issue: the `Scaffold` in `YaruLandscapeLayout` was recently replaced by `ScaffoldManager`, which doesn't provide a `Material` widget on its own, but needs a `Scaffold` ancestor. So not an issue with this PR but with yaru widgets :)

jpnurmi:
We're overusing `Scaffold` everywhere. The documentation says that it's best to avoid nesting them... :) For our use cases, often a basic header+body+footer layout with a material background would do, I suppose.

Feichtmeier:
I can change those scaffolds to column [ appbar, expanded ] if this is better

PR #748, opened 2023-01-12
Use the round back button as before

Feichtmeier:
grafik


jpnurmi:
One shape doesn't fit all use cases... Perhaps we could add a style argument to `YaruBack/IconButton`?

```dart
enum YaruBack/IconButtonStyle { round, square }
```

Jupi007:
> One shape doesn't fit all use cases... Perhaps we could add a style argument to YaruBack/IconButton?

@jpnurmi this sounds like a good idea :+1:

Feichtmeier:
@jpnurmi good idea :)

PR #753, opened 2023-01-12
Searchbar look

### LINKED ISSUE #752: Searchbar look
I would propose the searchbar to have a background and make it less long and move it into the middle of the title

Edit: screenshot is outdated, please look into #753 

@mads and/or @anasereijo do you agree with this?

Feichtmeier:
Edit: for screenshots please scroll down

Fixes #752

### DIFF
--- a/lib/app/explore/start_page.dart
+++ b/lib/app/explore/start_page.dart
@@ -85,8 +85,8 @@ class _StartPageState extends State<StartPage> {
             ? const CustomBackButton()
             : null,
         titleSpacing: 0,
-        centerTitle: false,
         title: SearchField(
+          autofocus: false,
           searchQuery: searchQuery,
           onChanged: setSearchQuery,
         ),
--- a/lib/app/common/search_field.dart
+++ b/lib/app/common/search_field.dart
@@ -70,12 +70,13 @@ class _SearchFieldState extends State<SearchField> {
         child: Center(
           child: SizedBox(
             width: 300,
-            height: 34.5,
+            height: 34,
             child: TextField(
               autofocus: widget.autofocus,
               controller: _controller,
               onChanged: widget.onChanged,
               textInputAction: TextInputAction.send,
+              textAlign: TextAlign.center,
               decoration: InputDecoration(
                 filled: true,
                 hintText: context.l10n.searchHint,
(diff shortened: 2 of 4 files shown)


madsrh:
@Feichtmeier the colors are good, but could you check the alignment? Looks to me like it's too close to the top. 

image

image

Feichtmeier:
@madsrh you are right, I think I can fix this by adjusting the text style and eventually also centering the text? The last bit is not directly related but IMO improves the whole thing:

grafik
grafik

madsrh:
It's better - maybe you went just a pixel or two too far down.

> ...eventually also centering the text?

As much as I like things to be in balance and centered, this text works best left aligned IMO

Feichtmeier:
Okay, what about this

grafik

Hover:
grafik

Focused
grafik

madsrh (approved):
Yes, thank you 👍 
Let's merge this one and if @anasereijo has additional comments we can always tweak this in another issue.

> What the hell 

Lol 🤣

Feichtmeier:
@madsrh 

A bit more centered. Works for the text field and the cursor but this text in flutter has some serious problems

grafik


Hover:
grafik


Focused
grafik

Feichtmeier:
@madsrh this won't be merged until ana approved it, since I enabled auto-merging when all requirements are met, and when you put two reviewers both are required. When you put 1 reviewer, 1 is enough since I've set the minimal number of reviewers two 1. So let's wait for ana since she said she had some ideas for the search bar too

Feichtmeier:
Uhm okay now the auto merge merged it :shrug: 

@anasereijo since mads approved it I hope this is not too much of a problem. If you want a different design or improvements we can change it next week :+1:

lib/app/explore/start_page.dart
```diff
@@ -85,8 +85,8 @@ class _StartPageState extends State<StartPage> {
             ? const CustomBackButton()
             : null,
         titleSpacing: 0,
-        centerTitle: false,
         title: SearchField(
+          autofocus: false,
```
d-loose, line 89: Is there any particular reason to disable the autofocus here?
It causes the search field to lose focus when the search string is deleted

https://user-images.githubusercontent.com/113362648/212667528-dcea694e-8128-4eea-93bf-2e6cdb93440c.mp4

Feichtmeier: Not, this is a mistake by me. We can re-enable it. Even if I'm unhappy with the current hierarchy of

ExploreModel -> ExplorePage -> Startpage + Search page 

Which is also the cause of the autofocus requirement.

It would be better and cleaner if we would instead use a navigator and move to search page when one sends the enter keyboard key

Then autofocus wouldn't be required

anasereijo:
I still need to look into this with more detail, so bear with me!
But one thing that caught my attention and might be a _quick thing to do_ (designers love saying this, right? hehe) is that the clear search icon (x) is missing. I think it would be nice to have it again as a quick way to clear the search and also to go back.

PR #754, opened 2023-01-12
Update yaru widgets ...

Feichtmeier:
.. for several bug fixes and improvements to YaruWindowTitlebar and friends. For details go to the repository


jpnurmi:
@Feichtmeier Not sure if this is affected but I didn't quite get the optimized window state right on the first attempt. I'm working on a fix:
- https://github.com/ubuntu/yaru_widgets.dart/pull/507

jpnurmi (approved):
LGTM :+1: 

There's another in the pipeline that allows you to use `YaruDetailPage` to keep the title bar in place during animations, but we can take that separately! So much going on! :sweat_smile:

PR #758, opened 2023-01-12
Fix API change: Bring back dialog close buttons

Feichtmeier:
This is a  needed fix for the api change of titlebars in yaru widgets to make the **close button** in dialogs return

grafik


jpnurmi (approved):
Sorry for the behavioral break. LGTM :+1:

Feichtmeier:
Nothing to be sorry about I just updated yaru widgets without adapting to it :) Thanks for the review

PR #720, opened 2023-01-09
Weblate merge conflict

seb128:
Unsure why it ended up locked but weblate states

```
Rebasing (1/6)
error: could not apply 58c6699... Translated using Weblate (Spanish)
Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".
Could not apply 58c6699... Translated using Weblate (Spanish)
Auto-merging lib/l10n/app_es.arb
CONFLICT (content): Merge conflict in lib/l10n/app_es.arb
```

The proposed change is basically doing what weblate suggests
$ git remote add weblate https://hosted.weblate.org/git/ubuntu-software/ubuntu-software/
$ git remote update weblate
$ git merge weblate/main 

which should allow the weblate side to pull again without error

PR #765, opened 2023-01-13
Translated using Weblate (Spanish)

seb128:
Currently translated at 100.0% (150 of 150 strings)

Translation: Ubuntu Software Store/ubuntu-software
Translate-URL: https://hosted.weblate.org/projects/ubuntu-software/ubuntu-software/es/


seb128:
The change in #720 wasn't enough to fix weblate because the project is configured with a rebase workflow, that one should do the trick and resolve the conflict

seb128:
Could someone review that one, it should be trivial and would allow to unblock translations which are currently locked

Feichtmeier (approved):
looking good :smile: 

Thank you @seb128 for taking the time!

seb128:
Thanks @Feichtmeier!

PR #767, opened 2023-01-13
Use `context.select` in SearchPage

d-loose:
Improvement of #756 to keep the spirit of #609 alive


Feichtmeier:
@d-loose did you test if the page 📄 is rebuild when you only change the snap category in the drop down with a not empty but not changed search query?

d-loose:
Yes :)

https://user-images.githubusercontent.com/113362648/212351339-fae8e228-4be8-4fc3-aef3-4b2c5e16e622.mp4

Feichtmeier (approved):

PR #768, opened 2023-01-13
Update yaru widgets for doubleclick->maximize on the first titlebar

Feichtmeier:
https://user-images.githubusercontent.com/15329494/212406443-a891e0c3-ced2-42b9-bbf9-8f2b53578157.mp4

- left bar has no window controls in landscape layout but has them in portrait layout
- left bar reacts on double click even if undecorated

Fixed in yaru widgets by @jpnurmi


Feichtmeier:
Include in #772

PR #781, opened 2023-01-15
Updates Page: swap tabs

Feichtmeier:
... for discoverability of system updates

grafik


madsrh (approved):
I think this works much better :pray: :muscle: Thanks you!

It would be a rare case that you would want to update your auto updating applications, right? If I was waiting for a new Firefox release, I think I might just search for Firefox and do the update from the appPage. I mean, if it's only snap that haven't audo updated yet, that list will properly never be very long.

PR #785, opened 2023-01-15
Updates: create a single page if packagekit is not available

### LINKED ISSUE #782: Updates: create a single page if packagekit is not available
Since this app is also targeting ubuntu core, which only has snap packages we should display the snap updates as updates and remove the tabbed view

grafik

Feichtmeier:
Fixes #782


d-loose (approved):
:+1:

PR #787, opened 2023-01-15
SnapUpdates: replace future builders with init/notify

### LINKED ISSUE #786: SnapUpdates: replace future builders with init/notify

Feichtmeier:
Fixes #786


d-loose (approved):
:+1:

PR #793, opened 2023-01-16
Show progress indicator when installing/removing deb packages

### LINKED ISSUE #783: Debian package installation: improve progress tracking on the AppPage
When installing a debian package currently no progress is shown on the app page. 

@d-loose could you help out with this one?

There are more issues, like the progress indication which currently only tracks snap installtion/updates, but this is a different issue.

d-loose:
https://user-images.githubusercontent.com/113362648/212661280-9c1232e7-3c10-4b55-ae9e-9ba2736e737f.mp4

Currently `PackageState` has only two states (`processing` and `ready`) - we should probably add more specific states in the future.

fix #783

### DIFF
--- a/lib/app/common/packagekit/package_controls.dart
+++ b/lib/app/common/packagekit/package_controls.dart
@@ -18,6 +18,7 @@
 import 'package:flutter/material.dart';
 import 'package:software/l10n/l10n.dart';
 import 'package:software/services/packagekit/package_state.dart';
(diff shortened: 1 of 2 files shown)


Feichtmeier (approved):
Thank you! Maybe in the future it would be nice to track the progress in a linear progress indicator but since this aligns the progress to snaps this is good 😊👍

Issue #788, opened 2023-01-15
AppIcons: always use a BasePlate?


Feichtmeier:
Even if the "store hygiene" could eventually make some store publishers change their mind and create high quality icons, we will never catch them all.

So how about presenting all app icons on a baseplate?

Here is a very horrible example
![grafik](https://user-images.githubusercontent.com/15329494/212562501-068c683a-17b9-4682-a632-ee0112eb68db.png)

With a baseplate
![grafik](https://user-images.githubusercontent.com/15329494/212562945-692e5302-04e4-4e32-8623-4f51ac7e21f1.png)

![grafik](https://user-images.githubusercontent.com/15329494/212562925-63dbb91f-400f-443d-b9e4-2d112b2cda89.png)
![grafik](https://user-images.githubusercontent.com/15329494/212563292-f56ec52b-f001-41cf-bdb0-0f71543856b0.png)
![grafik](https://user-images.githubusercontent.com/15329494/212563458-0f678976-1fca-4b72-b8b7-85ba6e971c5c.png)

Jupi007:
Looks like a great improvement.

Some like spotify shouldn't have a baseplate. So the best would be to do like android, and allow the app publisher to add metadata about the app icon.

madsrh:
I was expecting the return of the squircle 😅

I like this idea 👍

Jupi007:
> I was expecting the return of the squircle sweat_smile

We had the same thought 😆

Feichtmeier:
Wow, I think this was the first time I proposed something and it went right through you all 

![](https://media.giphy.com/media/zmRa0J4k30faE/giphy.gif)

Eventually this needs improvements let's see

PR #795, opened 2023-01-16
BasePlate and FallBackIcon improvements

Feichtmeier:
- use a rounded square form to avoid clipping for hard edge square forms, i.e. to avoid this:
grafik

- adjust padding to make them look bigger when they are very small
- always make the fallback icon light to avoid bulls-eye effect, to be clear, to avoid this:
grafik

- adjust fallback icon borders

After (dark):
grafik
grafik
grafik
grafik

After (light):
grafik
grafik
grafik
grafik

### DIFF
--- a/lib/app/common/app_banner.dart
+++ b/lib/app/common/app_banner.dart
@@ -82,12 +82,13 @@ class AppBanner extends StatelessWidget {
     var appIcon = Padding(
       padding: const EdgeInsets.only(bottom: 55, right: 5),
       child: SizedBox(
-        height: 45,
-        width: 45,
+        height: 40,
+        width: 40,
         child: BasePlate(
+          childPadding: 5,
           useBorder: true,
           hovered: false,
-          radius: 100,
+          radius: 8,
           child: AppIcon(
             iconUrl: iconUrl,
           ),
--- a/lib/app/common/app_icon.dart
+++ b/lib/app/common/app_icon.dart
@@ -19,7 +19,9 @@ import 'dart:math';
 
 import 'package:flutter/material.dart';
 import 'package:shimmer/shimmer.dart';
-import 'package:software/app/common/border_container.dart';
+import 'package:yaru_widgets/yaru_widgets.dart';
+
+const _borderColor = Color.fromARGB(255, 189, 189, 189);
 
 class AppIcon extends StatelessWidget {
   const AppIcon({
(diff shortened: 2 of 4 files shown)


madsrh (commented):
I love the return of the squircle 😸

In the light theme there's IMO too many lines now. I think it looks much cleaner without the gray border.



I think we should remove the border and the drop shadow. This way you'll only notice the baseplate when using the dark theme.

Is there room enough for the icon size to return to 45? They seems very small.

anasereijo:
> In the light theme there's IMO too many lines now. I think it looks much cleaner without the gray border.
> I think we should remove the border and the drop shadow. This way you'll only notice the baseplate when using the dark theme.

I agree, removing it makes the light theme look much cleaner 👍 

> Is there room enough for the icon size to return to 45? They seems very small.

Do you mean the actual icon inside the baseplate or the baseplate size? I actually quite like the size now (counting with the baseplate), I remember thinking it was way too big before? But happy to explore! We were thinking of making the app-name smaller too at some point, so hopefully it will all balance well together.

madsrh:
> Do you mean the actual icon inside the baseplate or the baseplate size? 

Sorry, I meant for the appbanners here: (this is 40px)
borders

anasereijo:
> The idea is very nice but then the alignment is wrong

That is a good point. It's more noticeable in the app-banners than on the app-page. 

What about a very light grey stroke?

windows example
Screenshot 2023-01-16 at 17 05 37

Feichtmeier:
> 🤦‍♂️ Damn, ofc! and only adding the baseplate in the dark theme is a hacky solution? I'm asking for a friend 😅

The only reason I thought a baseplate was needed was because there are horrible, horrible icons out there, without any background but transparency #788 

We could also remove the baseplate again if we think we should rather contact the publishers with the shity icons (no irony)

anasereijo:
This is actually a nice topic to bring up on my weekly design meeting (which is happening tomorrow). I will see what others think

Feichtmeier:
Ok! I try the softer border as a solution to get this PR merged then. Because the approach currently in main has problems, as shown in the opening post. So we need to do something about it for now

anasereijo:
> Ok! I try the softer border as a solution to get this PR merged then. Because the approach currently in main has problems, as shown in the opening post. So we need to do something about it for now

Makes sense! Go for it and we can the iterate a bit more

madsrh (approved):
Sorry, for getting sidetracked. Let's merge this now and circle back to it later.

Feichtmeier:
https://github.com/ubuntu-flutter-community/software/issues/802

PR #797, opened 2023-01-16
SnapUpdates: fix grid to be top aligned

Feichtmeier:
Bug Fix, The grid should not be centred. Regression from #787


d-loose (approved):
LGTM :+1: 
Sorry I didn't catch it in #787!

PR #809, opened 2023-01-17
improve package service tests

d-loose:
back to a bit more than 20% coverage \o/

Ref #387


Feichtmeier (approved):
As the tests pass and the coverage levels up: LGTM :)

PR #818, opened 2023-01-18
Search page optimization

d-loose:
* fix `SearchPage` rebuilds
* add 200ms grace period to `SearchField` (i.e. start search only 200ms after the user stops typing)

Before

https://user-images.githubusercontent.com/113362648/213253723-234088db-2111-4d82-bdbe-8ee533249fd5.mp4

After

https://user-images.githubusercontent.com/113362648/213253787-16148abd-9558-47ce-a6a3-793da477ef5b.mp4

### DIFF
--- a/lib/app/common/search_field.dart
+++ b/lib/app/common/search_field.dart
@@ -15,6 +15,8 @@
  *
  */
 
+import 'dart:async';
+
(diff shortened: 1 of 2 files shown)


Feichtmeier (approved):
THank you!

PR #820, opened 2023-01-18
Search-field moves between loading screen and start page.

### LINKED ISSUE #808: Search-field moves between loading screen and start page.
The buttons on the right are not drawn immediately, so the search-field ends up shifting to the left after they are displayed.

https://user-images.githubusercontent.com/73116038/212875380-f2572cf4-bfd5-481f-bd8f-f9857c236c1c.mp4

We can either:
- position the search-field further to the left during the loading phase
- find a way to render the buttons immediately at startup

Feichtmeier:
Fixes #808


d-loose (approved):
:+1:

PR #846, opened 2023-01-22
BorderContainer: use YaruBorderContainer

Feichtmeier:
BorderContainer: use YaruBorderContainer

No visual differences but this makes it a little easier to track properties and maintain the code.
Replacing all occurences of BorderContainer with YaruBordercontainer would also work but it would generate more lines of diff since the defaults for 90% of the constructor calls for software are defaulted in bordercontainer

### DIFF
--- a/lib/app/common/app_icon.dart
+++ b/lib/app/common/app_icon.dart
@@ -39,7 +39,7 @@ class AppIcon extends StatelessWidget {
   Widget build(BuildContext context) {
     final fallBackIcon = BorderContainer(
       borderColor: color,
-      containerPadding: EdgeInsets.zero,
+      padding: EdgeInsets.zero,
       borderRadius: 200,
       width: size,
       height: size,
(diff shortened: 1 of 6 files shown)


jpnurmi (commented):
package_page.dart and snap_page.dart were using "containerPadding" previously, where package_update_page.dart used "childPadding". now they are all using the same "padding" property. did "containerPadding" and "childPadding" get mixed up?

with `Container` and `YaruBorderContainer`:
- `margin` is outside the border
- `padding` is between the border and the child

it's the classic CSS box model:
image

Feichtmeier:
No but I realized that it is effectively doing the same thing.
Container says:

```
Empty space to inscribe inside the [decoration]. The [child], if any, is placed inside this padding.

This padding is in addition to any padding inherent in the [decoration]; see [Decoration.padding].
```

I don't remember why I was so keen on naming it childpadding but the reason might have been removed since I can not observe any visual differences

jpnurmi (approved):
ok :+1:

Feichtmeier:
When the new fallback icon is in we can eventually delete this file and move the defaults to app page :)

PR #848, opened 2023-01-22
Replace AppLoadingPage with each individual container loading state

### LINKED ISSUE #847: Replace AppLoadingPage with each individual container loading state
Each time we change the design in AppPage, we need to adapt AppLoading page and check the pixels. 
Instead let's foward the initialized bool to border containers and let them show a shimmer if false

Feichtmeier:
- Replace AppLoadingPage with each individual container loading state
- Several fixed to avoid jumping:
- wrap package controls spinner in a sized box plus padding
- give normal window header a fixed height
- set the normal window title font size always to 20
- found the publisher name and relased at actually in appstream api and added it to the additional informations

Fixes #847

TODO:

- [x] fix container size jumping at the top


d-loose (approved):
LGTM :+1:

PR #850, opened 2023-01-22
Update dialog text fixes

### LINKED ISSUE #849: Changelog dialog font issues
Text isn't visible in the light theme in the details section.

madsrh:
Okay, this does work and fixes some issues, but I'm sure there's a better way of doing this :sweat_smile: 

Fixes #849 
- text color and size in "package description"
- adds a URL tooltip
- even padding for the window content

I would have likes to also add padding between each expander and hide the entire line with Website when no url is provided :man_shrugging:


lib/app/updates/update_dialog.dart
```diff
@@ -185,10 +202,12 @@ class _UpdateDialogState extends State<UpdateDialog> {
         ),
         isExpanded: false,
         expandIcon: const Icon(YaruIcons.pan_end),
-        child: Padding(
-          padding: const EdgeInsets.only(right: 16.0),
-          child: Text(
-            model.description,
+        child: BorderContainer(
+          child: Padding(
```
Jupi007, line 206: The padding looks weird with the `BorderContainer` container, because it has already a padding, so the text is shifted to the right currently.

image

The `Padding` widget should be removed imo.

Feichtmeier:
run

```
dart fix --apply
```

to fix all the trailing comma issues without doing anything :)

lib/app/updates/update_dialog.dart
```diff
@@ -90,92 +91,108 @@ class _UpdateDialogState extends State<UpdateDialog> {
...
         ),
-        child: MarkdownBody(
-          data: model.changelog.length > 4000
-              ? '${model.changelog.substring(0, 4000)}\n\n ... ${context.l10n.changelogTooLong} ${model.url}'
-              : model.changelog,
-          shrinkWrap: true,
-          selectable: true,
-          onTapLink: (text, href, title) =>
-              href != null ? launchUrl(Uri.parse(href)) : null,
+        child: BorderContainer(
```
Jupi007, line 94: With a container, the expandable title is too close from the border imo:

Capture d’écran du 2023-01-22 20-18-17

So can you wrap each `BorderContainer`, with something like a bottom `Padding` of 16:

Capture d’écran du 2023-01-22 20-18-28
```dart
Padding(
  padding: const EdgeInsets.only(bottom: 16),
  child: BorderContainer(
    // ...
  ),
)

**Review by Jupi007 — CHANGES_REQUESTED** [review]
@madsrh I just take a look, and I noticed that the text color problem is related to `YaruTile` and so should be fixed in yaru_widgets.

So can you revert your changes related to this, and open an issue there? :)

I also added a few comments.
Feel free to ask help if my explanations are too obscure.

**Comment by Jupi007** [comment]
> run
> 
> ```
> dart fix --apply
> ```
> 
> to fix all the trailing comma issues without doing anything :)

In addition of formatting :)

```
dart fix --apply
dart format --apply
```

**Comment by Feichtmeier** [comment]
didnt know the last one :O does this fix the import?

**Comment by Feichtmeier** [comment]
((( Note: we should some day put a `final theme = Theme.of(context);` at the start of build at re-use it here. Quite many calls (by me! :D)

**Comment by Jupi007** [comment]
Sorry, looks like this command is just reporting errors in the code 😅

**Comment by madsrh** [comment]
Adding this did not work for me:

```
Padding(
  padding: const EdgeInsets.only(bottom: 16),
```

Can we fix this later? 😅

Feichtmeier:
No, you are trapped here for eternity 😅

I'll check tomorrow what's happening in here

Feichtmeier (approved):

lib/app/updates/update_dialog.dart
Feichtmeier: @madsrh do you plan to implement @Jupi007 change request? 😄

madsrh: Adding the padding didn't work, but no, I've found another issue so if it's okay, I would like to close this PR and make a new (simpler) one.

Feichtmeier: 😃 sure

PR #853, opened 2023-01-23
Show number of available updates and progress in launcher icon

### LINKED ISSUE #687: Show amount of updates and progress in the Ubuntu dock
It would be nice to use this unity7 ubuntu dock / plank API to show the amount of updates + progress bar on updating in the ubuntu dock

grafik

I have no idea how this works though :shrug: 

@3v1n0 maybe and idea? :)

d-loose:
https://user-images.githubusercontent.com/113362648/214027558-5e0e5424-fa7c-4a24-b73f-58affd57f551.mp4

It would be nice to show an overall percentage somehow, but for now all we have is the percentage that `PackageKitItemProgressEvent` provides.

fix #687

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -19,6 +19,7 @@ import 'package:connectivity_plus/connectivity_plus.dart';
 import 'package:desktop_notifications/desktop_notifications.dart';
 import 'package:flutter/material.dart';
 import 'package:gtk_application/gtk_application.dart';
+import 'package:launcher_entry/launcher_entry.dart';
 import 'package:packagekit/packagekit.dart';
 import 'package:snapd/snapd.dart';
 import 'package:software/app/app.dart';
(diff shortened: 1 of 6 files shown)


lib/main.dart
```diff
@@ -61,5 +62,11 @@ Future<void> main(List<String> args) async {
     dispose: (s) => s.dispose(),
   );
 
+  registerService<LauncherEntryService>(
+    (() => LauncherEntryService(
+          appUri: 'application://snap-store_snap-store.desktop',
```
d-loose, line 67: This is the `appUri` that works in the snap. It's a bit awkward to test locally: either change `appUri` to `application://software.desktop` and install the desktop file from the assets in `.local/share/applications`, or leave it as it is and instead rename the desktop file to snap-store_snap-store.desktop (the binary listed in `Exec` needs to be renamed to `snap-store_snap-store` in that case..)

Feichtmeier (approved):
🙏👍 thank you! Looks fine
We could someday simply count the updates and increase the progress after each package kit finish event. But this is something for a different pr

PR #862, opened 2023-01-23
AppPage and package updates in wide windows: limit adaptive padding

### LINKED ISSUE #861: AppPage in wide windows: remove adaptive padding
https://user-images.githubusercontent.com/3986894/214158165-404082f2-edac-441b-a9b6-8572806bc83b.mp4

Feichtmeier:
@madsrh but on a sidenote, I think we've spoken about this many many times. You need to apply the scaling if you are on a 4k screen :) You and no other designer will be able to design a layout that works on a 1:1 scaling on a 4k screen. I wonder why you always check those 4k 1:1 resolutions? :) 

grafik

Fixes #861

### DIFF
--- a/lib/app/app.dart
+++ b/lib/app/app.dart
@@ -93,7 +93,6 @@ class _App extends StatefulWidget {
 }
 
 class __AppState extends State<_App> {
-  int _updatesPageIndex = 0;
   bool _initialized = false;
   int _initialIndex = 0;
   String? debPath;
@@ -175,10 +174,9 @@ class __AppState extends State<_App> {
       ),
       PageItem(
         titleBuilder: UpdatesPage.createTitle,
-        builder: (context) => UpdatesPage(
+        builder: (context) => UpdatesPage.create(
+          context: context,
           windowWidth: width,
-          onTabTapped: (index) => setState(() => _updatesPageIndex = index),
-          tabIndex: _updatesPageIndex,
         ),
         iconBuilder: (context, selected) => UpdatesPage.createIcon(
           context: context,
--- a/lib/app/common/app_page/app_infos/additional_information.dart
+++ b/lib/app/common/app_page/app_infos/additional_information.dart
@@ -26,45 +26,48 @@ class AdditionalInformation extends StatelessWidget {
         context.l10n.additionalInformation,
         style: Theme.of(context).textTheme.headline6,
       ),
-      child: ScrollConfiguration(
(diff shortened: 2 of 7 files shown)


Feichtmeier:
We could check each indiviual section in wide mode (NOT ON A 4K SCREEN WITH 1:1 scaling) and see if we need to make edge cases like in the gallery
grafik

madsrh (commented):
I prefer the current master. It works perfect until +1800px

Would it be possible (with master) after a certain width to "freeze" the width of the green part and only add padding or ?

image

Feichtmeier:
yes

Feichtmeier:
but this wont stop you from using the appon a 4k scrren with 100% scaling ;)

Feichtmeier:
Instead of adding a limit to scaling we could add a limit to the window size. Which would ofc not change 200% scaling

madsrh:
> Instead of adding a limit to scaling we could add a limit to the window size. Which would ofc not change 200% scaling

You will know what the best solution is here 🤷‍♂️ What would happen if I maximized the window then? I think the padding solution is "simpler".

>  ...on a sidenote, I think we've spoken about this many many times...

We have indeed and sorry for bringing it up again, but even if it's an edge cast, I think we should plan for it. And no, I don't actually use my desktop in this 100% mode 😄 

It looks odd, but this Microsoft solution is better than overflow.
image

Feichtmeier:
We can fix the overflow. But we can not design for those edge cases. It's impossible and will affect the actual target resolutions and scales in a bad way. I'm not saying this is not an issue in just a bit .... "surprised" that we are here again... Looking at a 4k+ screen with 100% scaling. Windows is much better at 4k scaling most importantly because it immediately suggests a 150-200% scaling the moment it detects an enormous amount of pixels. Ubuntu doesn't do this and yeah.... Here we are. But this is not our problem at least not in this repo, seriously 😒

madsrh:
> We can fix the overflow. 

If it's a small fix that would ofc be good 👍 

Sorry for bringing it up again. Other than this edge case, I'm **very** happy with the current layout(s), so let's just close this 😸

Feichtmeier:
no it's an issue, we can fix it

Feichtmeier:
@madsrh 

grafik

please test on your times square screen :)

madsrh (approved):
YES!!! This work - Thank you 🙏

PR #870, opened 2023-01-25
Reload Explore page when clicking sidebar

### LINKED ISSUE #832: Reload Explore page when clicking sidebar
(sorry for the crappy title) Related to https://github.com/ubuntu-flutter-community/software/issues/254

If I search for something, and then press **Explore** in the sidebar, I would expect to return to the empty Explore page (as when the app launched). Currently, nothing happens.

image

d-loose:
Fixes #832 by making `AppModel` emit a signal when a selected page index is clicked again. `ExplorePage` then clears the search query.

I'm not sure if this isn't getting too convoluted - maybe we should consider a larger refactor of `ExplorePage` or improve the sidebar functionality in the yaru widgets instead.

P.S. all hell might break loose in the CI due to changes in flutter 3.7.0


lib/app/explore/explore_page.dart
```diff
@@ -75,11 +100,12 @@ class ExplorePage extends StatelessWidget {
     return Scaffold(
       appBar: YaruWindowTitleBar(
         title: SearchField(
+          key: ValueKey(showSearchPage),
```
d-loose, line 103: This makes sure the `SearchField` is replaced with an empty one, when the `searchQuery` is cleared.

Feichtmeier (approved):

PR #875, opened 2023-01-26
Replace xterm with LogView from ufp/ubuntu_widgets

jpnurmi:
LogView is compatible with Flutter 3.7.0 but also lighter because it's just a plain read-only TextField under the hood.

| Before | After |
|---|---|
| Screenshot from 2023-01-26 11-37-10 | Screenshot from 2023-01-26 11-50-59 |


Feichtmeier (approved):
TY!

PR #879, opened 2023-01-26
Flutter 3.7: code fixes

Feichtmeier:
- use yaru 0.5.0 and yaru_widgets from git ref which also depends on yaru 0.5.0 to not create any visual regression with the new flutter stabel version
- ran dart fix to change text theme name changes
- updated some icons to use onSurface color to not create visual regression


Feichtmeier:
sec need to fix CI

jpnurmi (approved):
LGTM :+1:

PR #892, opened 2023-01-27
Add clear icon to searchbar

### LINKED ISSUE #801: Add clear icon to searchbar
From @anasereijo https://github.com/ubuntu-flutter-community/software/pull/753#issuecomment-1384287838

> I still need to look into this with more detail, so bear with me! But one thing that caught my attention and might be a _quick thing to do_ (designers love saying this, right? hehe) is that the clear search icon (x) is missing. I think it would be nice to have it again as a quick way to clear the search and also to go back.

image

EDIT: clear icon only visible after the user starts typing

Feichtmeier:
Fixes regression, clear icon is back now if the field is not empty
grafik

Fixes #801


d-loose (approved):
LGTM :+1:

PR #894, opened 2023-01-27
Search with special characters gives bad query

### LINKED ISSUE #863: Search with special characters gives bad query
Poor title (again, sorry). Do we need an error message like the 🐣 ❓  or just allow special characters?
You can search with a **+** and **&** on _snapcraft.io_ although the results are the same as when you search without.

image

Feichtmeier:
image
image
image

Fixes #863

### DIFF
--- a/lib/app/explore/explore_model.dart
+++ b/lib/app/explore/explore_model.dart
@@ -128,9 +128,11 @@ class ExploreModel extends SafeChangeNotifier {
     if (searchQuery.isEmpty) {
       return [];
     } else {
+      final clearedSearchQuery =
+          searchQuery.replaceAll(RegExp('[^A-Za-z0-9]', unicode: true), '');
       try {
         return await _snapService.findSnapsByQuery(
-          searchQuery: searchQuery,
+          searchQuery: clearedSearchQuery,
           sectionName:
               selectedSection == SnapSection.all ? null : selectedSection.title,
         );


lib/app/explore/explore_model.dart
```diff
@@ -128,9 +128,11 @@ class ExploreModel extends SafeChangeNotifier {
     if (searchQuery.isEmpty) {
       return [];
     } else {
+      final clearedSearchQuery =
+          searchQuery.replaceAll(RegExp('[^A-Za-z0-9]', unicode: true), '');
```
d-loose, line 132: We also need to preserve whitespace, otherwise entering 'foo bar' will only give you results for 'foobar'

d-loose (changes requested):
Should we apply the same search string cleaning to `findAppstreamComponents` for the sake of consistency?

lib/app/explore/explore_model.dart
Feichtmeier: whats the regex for this?

Feichtmeier:
🤔 thinking about this again I think we shouldn't do this here at all. IMO this is something for snapd or snapd.dart
We just send strings

lib/app/explore/explore_model.dart
d-loose: I'd propose

```dart
replaceAll(RegExp(r'[^\w\s]+'), ' ', unicode: true)
```
to replace all non-alphanumeric and non-whitespace characters by a single space.

Feichtmeier: Problem is for example xournal++

Feichtmeier:
Can I close this @d-loose ? I think this is the wrong approach and the wrong repository

d-loose:
I'm trying to figure out where the problem is - dart's `Uri` class should take care of correctly encoding the query parameters when sending a request to the snapd API..
I'd leave it open until we have a fix (here or elsewhere) - the user definitely shouldn't see a 'Bad request' instead of search results.

PR #896, opened 2023-01-27
Move connectivity from AppModel into a small standalone unit

jpnurmi:
The connectivity status is not needed in AppModel. Having it in a small separate unit that only notifies connectivity status changes makes it cheap to watch anywhere connectivity status is needed. At the moment, the only place that needs the connectivity status is ExplorePage.

[connectivity.webm](https://user-images.githubusercontent.com/140617/215121727-58147503-43b9-4cc2-8d12-e9e77ecb1864.webm)


Feichtmeier (approved):
ty!

PR #897, opened 2023-01-27
AppBanner & PackagePage: fill all infos with data or fallback string

### LINKED ISSUE #852: PackagePage: fill all infos with data or fallback string
I recently found that `relasedAt` can be indeed received via appstream (sometimes).
The developer/publisher name can also be found sometimes.
We also got package kit left to eventually fill up more infos.
If neither works for getting the infos we should include a placeholder string like "Unknown" or "Not available".

@madsrh @anasereijo @Jupi007 any preference on the placeholder string?

Feichtmeier:
- appstream uses kind of random locale strings
- in package page we also have the package id that can be used for the license :+1: 
- appstream sometimes has released at info, fallback to unkown as discussed with the other designers :+1: 
- reliable language tags are "C" (haha so funny 90s geeks), "en_GB" and "en"
- fallback in appbanner is the name, since most of the publisher names would be "UNKNOWN" otherwise which looks really weird here
- released at needs improvements in terms of formatting, in a second PR after this
- changed most of the model getters to return nullable types, to correctly digest this in the UI, we should do this in a different PR for more getters, they should all be null if they could be null and we should never return default values in models anymore like '' or 0, or whatever

grafik
grafik
grafik
grafik

Fixes #852

### DIFF
--- a/lib/app/common/app_banner.dart
+++ b/lib/app/common/app_banner.dart
@@ -201,9 +201,10 @@ class SearchBannerSubtitle extends StatelessWidget {
     }
 
     if (appFinding.appstream != null && showPackageKit && !showSnap) {
-      publisherName = appFinding.appstream!.developerName['C'] ??
-          appFinding.appstream!.developerName['en'] ??
-          appFinding.appstream!.developerName['en_GB'] ??
+      publisherName = appFinding.appstream!.developerName[WidgetsBinding
+              .instance.window.locale.countryCode
+              ?.toLowerCase()] ??
+          appFinding.appstream!.developerName['C'] ??
           appFinding.appstream!.localizedName();
     }
 
--- a/lib/app/common/app_banner.dart
+++ b/lib/app/common/app_banner.dart
@@ -201,16 +201,10 @@ class SearchBannerSubtitle extends StatelessWidget {
     }
 
     if (appFinding.appstream != null && showPackageKit && !showSnap) {
-      if (appFinding.appstream!
-              .developerName[Localizations.localeOf(context).toLanguageTag()] !=
-          null) {
-        publisherName = appFinding.appstream!
-            .developerName[Localizations.localeOf(context).toLanguageTag()]!;
-      } else if (appFinding.appstream!.urls.isNotEmpty) {
-        publisherName = appFinding.appstream!.urls
-            .firstWhere((element) => element.url.isNotEmpty)
-            .url;
-      }
+      publisherName = appFinding.appstream!.developerName['C'] ??
+          appFinding.appstream!.developerName['en'] ??
+          appFinding.appstream!.developerName['en_GB'] ??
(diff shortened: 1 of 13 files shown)


Feichtmeier:
@d-loose 
integration test saved me of a NPE because I copy pasted stuff around :D 
:test_tube: :+1: :muscle:

lib/app/common/app_banner.dart
```diff
@@ -201,16 +201,10 @@ class SearchBannerSubtitle extends StatelessWidget {
...
-            .developerName[Localizations.localeOf(context).toLanguageTag()]!;
-      } else if (appFinding.appstream!.urls.isNotEmpty) {
-        publisherName = appFinding.appstream!.urls
-            .firstWhere((element) => element.url.isNotEmpty)
-            .url;
-      }
+      publisherName = appFinding.appstream!.developerName['C'] ??
+          appFinding.appstream!.developerName['en'] ??
+          appFinding.appstream!.developerName['en_GB'] ??
+          appFinding.appstream!.localizedName();
```
d-loose, line 207: Why are the locales prioritized this way? I'd expect to try the localized version first, and use English and 'C' as fallbacks

lib/app/common/packagekit/package_model.dart
```diff
@@ -62,11 +62,13 @@ class PackageModel extends AppModel {
...
-  String get title => appstream?.localizedName() ?? packageId?.name ?? '';
+  String? get title => appstream?.localizedName() ?? packageId?.name;
 
   String? get developerName {
-    return appstream?.developerName[
-        WidgetsBinding.instance.window.locale.countryCode?.toLowerCase()];
+    return appstream?.developerName['C'] ??
+        appstream?.developerName['en'] ??
+        appstream?.developerName['en_GB'] ??
+        appstream?.localizedName();
```
d-loose, line 71: Same as above

lib/app/common/app_banner.dart
Feichtmeier: 🆗👍
The localized versions didn't work except C, but you are right, we can at least pretend they work and maybe find out why the map is so often empty

Feichtmeier:
@d-loose works sometimes, but 'C' is a good fallback for those only having 'C' as an entry. 

grafik
grafik

Is this okay for you?

```dart
String? get developerName {
    final devName = appstream?.developerName[
            WidgetsBinding.instance.window.locale.countryCode?.toLowerCase()] ??
        appstream?.developerName['C'];

    return devName ?? appstream?.localizedName();
  }
```

d-loose:
As long as `appstream?.developerName[WidgetsBinding.instance.window.locale.countryCode?.toLowerCase()]` is tried first, I'm happy :)

d-loose (approved):
:+1:

PR #899, opened 2023-01-27
Remove unnecessary usage of shrinkwrap: true

whiskeyPeak:
As per the flutter documentation, making use of this property can lead to some performance issues due to the scrollView needing to be recomputed each time the scroll position changes. I tested this pretty thoroughly and didn't come across any issues.

The only place where I couldn't remove it was in the section_grid.dart and loading_banner_grid.dart files because they are nested inside a SingleChildScrollView. Optimally, we should make use of slivers in those cases but that would require a bit more refactoring.

Also fixes the bottom padding from 25 to 20 for the installed snaps and deb pages.

Before:
Screenshot from 2023-01-27 19-45-12

After:
Screenshot from 2023-01-27 19-42-57

@Feichtmeier Since the default value of shrinkWrap is false, should I just remove all of the ```shrinkWrap: false,``` instances?

### DIFF
--- a/lib/app/common/app_page/app_description.dart
+++ b/lib/app/common/app_page/app_description.dart
@@ -40,7 +40,7 @@ class AppDescription extends StatelessWidget {
         padding: const EdgeInsets.only(top: 20),
         child: MarkdownBody(
           data: description,
-          shrinkWrap: true,
+          shrinkWrap: false,
           selectable: true,
           onTapLink: (text, href, title) =>
               href != null ? launchUrl(Uri.parse(href)) : null,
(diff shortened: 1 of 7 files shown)


Feichtmeier:
Better wait with this until we got the be new explore page design

Feichtmeier:
However you can make a different PR with only the padding changes if you like :+1:

whiskeyPeak:
> Better wait with this until we got the be new explore page design

Yeah, I also encountered weird behaviour in some cases. Defo need to test this some more.

Feichtmeier:
@whiskeyPeak feel free to close if you do not plan to continue this branch

Issue #739, opened 2023-01-11
App-page Dependencies pop-up visual improvements 


anasereijo:
How it looks like:
<img width="552" alt="Screenshot 2023-01-11 at 12 08 28" src="https://user-images.githubusercontent.com/36884067/211802679-ae660bb2-40a3-45e1-9a59-9143a8bf7be7.png">

I was looking around Ubuntu OS, looking for dialogue pop-ups (not sure what you call them) with expand/collapse info, and found these 2 examples:

<img width="597" alt="Screenshot 2023-01-11 at 11 26 21" src="https://user-images.githubusercontent.com/36884067/211794662-91c5b48d-d9f3-4f6e-945a-545e548105bc.png">

<img width="481" alt="Screenshot 2023-01-11 at 11 35 50" src="https://user-images.githubusercontent.com/36884067/211796469-3a7fccbc-1c27-4c00-ad92-e7ae2c22f95a.png">

<img width="701" alt="Screenshot 2023-01-11 at 11 26 59" src="https://user-images.githubusercontent.com/36884067/211794794-470cbdf2-61fa-41f9-a9d1-9d3274e9f692.png">

I know it's a different design framework, but I think it would be nice to keep some consistency, at least on functionality. 
For instance, we could position the chevron before the text? Also, the text can be black and not as bold as per the examples.  Maybe also add a border when it expands.

I can mockup something if needed.

madsrh:
I'm not sure I understand @anasereijo but all the examples you've found, are going to be incorporated into Software eventually 🤞 (long term plan), so IMO we don't need to worry about consistency and matching this old design.

> ...we could position the chevron before the text...

The current position comes from [Material design](https://m1.material.io/components/expansion-panels.html)
Do you mean like this? 

<img width="506px" src="https://user-images.githubusercontent.com/3986894/211800126-d96d80dc-c80e-4753-bd6d-5796f81c8b0c.png">

anasereijo:
I meant in the popup. Unless we create a container (make it within a section with a border as per your examples) having the chevron on the right makes it look very floaty and not part of the link.

madsrh:
> having the chevron on the right makes it look very floaty and not part of the link.

I agree and I understand now 😸 
Can you create a mockup for this issue?

Jupi007:
Honestly, I really like the left aligned chevron idea :+1: 
This just need a few adaptation of the widget in yaru_widgets.

Feichtmeier:
[...]

PR #905, opened 2023-01-29
Small styling fixes for dependecies popup

madsrh:
Very small improvements towards #739

- moved expander to start
- removed text changing (feels weird that the text changes when expanded)
- added border

As you can see there's still some issues. Just a placeholder until some advanced help (not me) arrives implementing [this idea 🖼️  ](https://github.com/ubuntu-flutter-community/software/issues/739#issuecomment-1399258699)

### DIFF
--- a/lib/app/common/packagekit/package_page.dart
+++ b/lib/app/common/packagekit/package_page.dart
@@ -292,29 +292,30 @@ class _ShowDepsDialogState extends State<_ShowDepsDialog> {
             ),
           ),
           YaruExpandable(
+            expandButtonPosition: YaruExpandableButtonPosition.start,
             onChange: (isExpanded) => setState(() => _isExpanded = isExpanded),
             header: MouseRegion(
               cursor: SystemMouseCursors.click,
               child: Text(
-                _isExpanded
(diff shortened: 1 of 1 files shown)


Feichtmeier:
But this text was included in the mock up by @anasereijo 
Double check with her please

anasereijo:
heyy this is looking good, @madsrh 👍 

I think it's okay to leave the dropdown text as 'Dependencies' as long as it stays the same when expanded.

I have one suggestion: is it possible to make the list of dependencies a bit more compact (as per my mock-up)? It will help to reduce the amount of scrolling.

Thanks!

Feichtmeier (approved):
lgtm

anasereijo:
> I was thinking the same about the orange text (not just because it's orange), but it also changes when expanded. IMO just plain text would be fine. WDYT @Feichtmeier @anasereijo ? Should I add this to the PR?

I agree 💯

madsrh:
> I agree 💯

> lgtm

Now I'm confused 😀 @Feichtmeier was so fast. Do you want this in a new PR or did you merge because the current "lgtm"?

PR #908, opened 2023-01-30
Launcher entry: update counter and progress bar simultaneously

### LINKED ISSUE #687: Show amount of updates and progress in the Ubuntu dock

d-loose:
Fix #687


Feichtmeier (approved):
Looks good and makes sense to me if the value can be null

Issue #855, opened 2023-01-23
Searchbar on the apppage


Feichtmeier:
As a result of the meeting:

- make the search available on the AppPage

@d-loose could you eventually tackle this one?

madsrh:
> ... a suggestion dropdown for this or maybe a change from "search as you type" to "search on enter" is enough?

If a drop-down is possible that would be very slick 😎

Feichtmeier:
I think we need to discuss this again. The best would be if you would be there if we discuss this with @anasereijo and @madsrh 

Another way would be to move the search logic out of exploremodel one level higher in the AppModel and only pass the list of AppFindings to the lower pages if searched

anasereijo:
Agree. This needs a bit more thinking and discussion. We can talk about it on next week's 'work together' session? @d-loose I can add you to it if you are free! (Monday 5pm - UK)

d-loose:
@anasereijo can't promise I'll be there very long, but I'll try to join for half an hour or so.

PR #914, opened 2023-01-31
WIP: search field on `AppPage`

d-loose:
* Adds a search field linked to the `ExploreModel` to `AppPage`
* Only works if `AppPage` is reached from the `ExplorePage`
* Unclear: search behavior on `AppPage`s of locally installed packages (and pages reached through a `snap://` link or by opening a local .deb file). In those cases, starting a search for new packages in the `ExplorePage` would be confusing. Options:
  - Perform search in the given context (i.e. search for installed packages when viewing the `AppPage` of a locally installed package). Might also be confusing unless there is a clearer (visual) distinction between `AppPage`s for local packages and packages from the store / repository.
  - Only show the search field on `AppPages` packages that aren't installed

Ref #855

### DIFF
--- a/lib/app/app.dart
+++ b/lib/app/app.dart
@@ -39,6 +39,8 @@ import 'package:ubuntu_service/ubuntu_service.dart';
 import 'package:yaru/yaru.dart';
 import 'package:yaru_widgets/yaru_widgets.dart';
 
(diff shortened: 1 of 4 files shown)


d-loose:
Superseded by #936

PR #915, opened 2023-01-31
Refactor: separate app data and ratings

jpnurmi:
Decouple app data and associated ratings because the latter comes from a different (potentially slow and unreliable) data source. We must be able to present apps in the GUI without waiting for ratings to arrive.

Ref: #872


Feichtmeier (approved):
TY!

Issue #906, opened 2023-01-29
Add progress of installations (Snaps and Debs)


alculete:
I think having a way to know how much longer my apps will take to install would be a good idea. Some connections in some countries are not very fast, so knowing the progress in percentage of how the download is going would be great.
Also for snaps, include the information on plugs being downloaded so the user knows when the app has completed to install everything.

Feichtmeier:
@d-loose 

I often played around with this but never found a good solution. Snapd changes are divided into tasks. And you need to find out the time by

```dart
/// The time this change started.
final DateTime spawnTime;

/// The time this change completed.
final DateTime? readyTime;
```

Needs changes in snap model.

I think for package kit is this a bit easier.

@d-loose could you tackle this some day?

d-loose:
Why would we need those times? Both the snapd and packagekit service report a progress status for their actions(`PackageKitItemProgressEvent`, `SnapdTaskProgress`), which we currently don't display on the `AppPage`. Adding those together with some more useful information about what packagekit is currently doing (besides `processing` and being `ready`) would definitely be very useful. I can take care of that, but will probably need some design feedback.

Feichtmeier:
Snapd has the kind of change in there and I think I put this into the snap model. Again those are strings you need to check like currently in the snap model 

```dart
 String getChangeMessage({
    required BuildContext context,
    required String? changeKind,
  }) {
    switch (changeKind) {
      case 'install-snap':
        return context.l10n.installing;
      case 'remove-snap':
        return context.l10n.removing;
      case 'refresh-snap':
        return context.l10n.refreshing;
      case 'connect-snap':
        return context.l10n.changingPermissions;
      case 'disconnect-snap':
        return context.l10n.changingPermissions;
      default:
        return '';
    }
  }
  ```

which could be made more elegant like putting it as an extension on snapdchange or even directly into snapd.dart

d-loose:
Fixed by #916 #944

PR #916, opened 2023-01-31
Snap progress indicator

d-loose:
Adds a progress indicator for snap operations

[Screencast from 2023-01-31 13:13:51.webm](https://user-images.githubusercontent.com/113362648/215757099-02af8531-1d8f-457e-afbc-ffa15ba882fd.webm)

I'll add something similar for packagekit in a separate PR.

Ref #906

### DIFF
--- a/lib/snapd_change_x.dart
+++ b/lib/snapd_change_x.dart
@@ -1,0 +1,14 @@
+import 'package:snapd/snapd.dart';
+
+extension SnapdChangeX on SnapdChange {
+  double get progress {
+    var done = 0.0;
+    var total = 0.0;
+    for (final task in tasks) {
+      done += task.progress.done;
+      total += task.progress.total;
+    }
+
+    return done / total;
+  }
+}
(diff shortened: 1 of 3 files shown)


lib/snapd_change_x.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:snapd/snapd.dart';
+
+extension SnapdChangeX on SnapdChange {
+  double get progress {
```
Feichtmeier, line 4: ahh =) this was the part that I had missing!

lib/app/common/packagekit/package_page.dart
```diff
@@ -176,11 +176,6 @@ class _PackagePageState extends State<PackagePage> {
           );
 
     var controls = PackageControls(
```
Feichtmeier, line 178: why did you move this to a model call again? Just curious. Seems unrelated to me for this PR :thinking:

lib/snapd_change_x.dart
Feichtmeier: it's a good thing, works nicely!

lib/app/common/packagekit/package_page.dart
d-loose: Ah sorry, that's part of my packagekit experimentation - I'll exclude it from here :)

Feichtmeier (approved):
TY!

PR #924, opened 2023-01-31
Refactor: separate app model and reviews

jpnurmi:
Similarly to #915, reviews come from a different data source.

For now, it's the same fake review data but later on, we'll pass an ODRS service to the model that provides access to real data.

Ref: #872


lib/app/common/snap/snap_page.dart
```diff
@@ -92,17 +100,24 @@ class SnapPage extends StatefulWidget {
...
   bool initialized = false;
 
+  String get _ratingId => widget.snap.ratingId;
+  String get _ratingVersion => widget.snap.version;
+
   @override
   void initState() {
     super.initState();
-    context.read<SnapModel>().init().then((value) => initialized = true);
+    WidgetsBinding.instance.addPostFrameCallback((_) {
```
Feichtmeier, line 109: I think I have read somewhere that you can now just call `context.mounted` :thinking: or did I dream this?

Feichtmeier (changes requested):

lib/app/common/snap/snap_page.dart
Feichtmeier: so you don't safe any code here with this? you still need the postFrameCallBack ?

jpnurmi: it's the first line inside the post-frame callback:
```
if (!mounted) return;
```
it would not be safe to access the `State.context` property without this. the use case for `BuildContext.mounted` is different.

Feichtmeier (approved):

PR #926, opened 2023-01-31
Detach `ExploreModel` from `StartPage` and just pass a list of snaps and the `SnapSection` to the constructor

### LINKED ISSUE #925: Detach `ExploreModel` from `StartPage` and just pass a list of snaps and the `SnapSection` to the constructor
If we need either more "explorepages" or more snap sections + snap lists we can add this later if we want (after the explorepage re-design for example)

Feichtmeier:
Fixes #925


d-loose (approved):
LGTM :+1:

PR #931, opened 2023-01-31
ODRS ratings

jpnurmi:
NOTE: Only _ratings_ - not reviews. :)

| Explore | Snap | Deb |
|---|---|---|
| Screenshot from 2023-01-31 21-38-54 | Screenshot from 2023-01-31 21-39-14 | Screenshot from 2023-01-31 21-39-19 |

Ref: #872

### DIFF
--- a/lib/app/app.dart
+++ b/lib/app/app.dart
@@ -34,6 +34,7 @@ import 'package:software/app/settings/settings_page.dart';
 import 'package:software/app/updates/updates_page.dart';
 import 'package:software/l10n/l10n.dart';
 import 'package:software/services/appstream/appstream_service.dart';
+import 'package:software/services/odrs_service.dart';
 import 'package:software/services/packagekit/package_service.dart';
 import 'package:software/services/snap_service.dart';
 import 'package:ubuntu_service/ubuntu_service.dart';
@@ -57,7 +58,7 @@ class App extends StatelessWidget {
             create: (_) => ConnectivityNotifier(getService<Connectivity>()),
           ),
           ChangeNotifierProvider(
-            create: (_) => RatingModel(),
+            create: (_) => RatingModel(getService<OdrsService>()),
           ),
         ],
         child: const App(),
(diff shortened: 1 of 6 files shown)


Feichtmeier:
Amazomg @jpnurmi ! 💪 

Did you check the start-up / search performance with this new service?

jpnurmi:
The ODRS server can be sometimes a bit slow and unstable, and the API is not ideal because it only allows downloading them all as a bundle. Therefore, ratings are cached into `~/.cache/software/odrs/ratings.json` and thanks to the earlier changes they are now entirely decoupled from the app data to avoid blocking the GUI.

You can test it by inserting an artificial delay like this:
```diff
diff --git a/lib/services/odrs_service.dart b/lib/services/odrs_service.dart
index eeed16f..2fd9861 100644
--- a/lib/services/odrs_service.dart
+++ b/lib/services/odrs_service.dart
@@ -38,6 +38,7 @@ class OdrsService {
     }
     if (!exists || await file.expired(_kCachePeriod)) {
       try {
+        await Future.delayed(const Duration(seconds: 5));
         final ratings = await _client.getRatings();
         yield ratings;
         await file.writeRatings(ratings);
```
On the first startup, the ratings appear after the delay but on the second startup, they're read from the cache.

[Screencast from 2023-01-31 22-13-39.webm](https://user-images.githubusercontent.com/140617/215885214-1037dd21-18fc-4b68-a777-50b3235959cf.webm)

Feichtmeier (approved):
OKay nice, since they are decoupled and do not steal frames from anywhere I am fine with the first start delay 👍 
Thanks again! 💪

jpnurmi:
There's caching on two levels. First of all, we need a cache file to be able to restore the last known ratings as quickly as possible. The cache file has an expiry time of 1 hour. If the timestamp is older than that when asking for ratings, it returns the last known ratings immediately anyway and then re-fetches new ones. Secondly, we don't want to access the cache file repeatedly so the model keeps its own cache with a bit shorter expiry time.

PR #933, opened 2023-02-01
ODRS reviews

jpnurmi:
This PR implements the _fetching_ of reviews from ODRS. (Up/down-voting, reporting, and submitting reviews are not included in this.)

image

Ref: #872


Feichtmeier (approved):
Awesome works! Tested locally :smile: 

(((( I think I messed up the padding when switching the container, will fix after this ))))

PR #934, opened 2023-02-01
AppReviewsDialog: improve padding

Feichtmeier:
|Before|After|
|-|-|
|image|image|

### DIFF
--- a/lib/app/common/app_page/app_reviews.dart
+++ b/lib/app/common/app_page/app_reviews.dart
@@ -162,40 +162,38 @@ class _ReviewDetailsDialog extends StatelessWidget {
           ? []
           : userReviews!
               .map(
-                (e) => Padding(
-                  padding: const EdgeInsets.only(
+                (e) => BorderContainer(
+                  margin: const EdgeInsets.only(
                     left: kYaruPagePadding,
                     right: kYaruPagePadding,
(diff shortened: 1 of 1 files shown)


lib/app/common/app_page/app_reviews.dart
```diff
@@ -162,32 +162,40 @@ class _ReviewDetailsDialog extends StatelessWidget {
           ? []
           : userReviews!
               .map(
-                (e) => BorderContainer(
+                (e) => Padding(
```
jpnurmi, line 165: you can pass `margin` to avoid having to wrap Container or YaruBorderContainer with Padding

Feichtmeier: no, this is the old discussion :D padding in the container adds padding around the child :)
here we need both, around the child and around the container

Feichtmeier: image

Feichtmeier: eeeeeeeh margin :face_exhaling: ok sec

Feichtmeier: done :facepalm:

jpnurmi (approved):
LGTM :+1:

PR #936, opened 2023-02-01
Restructure UI and models

### LINKED ISSUE #921: make search available from App downwards

### LINKED ISSUE #922: move the searchfield to `_App` and pass the `searchResults` and the `ExploreHeader` from `AppModel` to `ExplorePage`->`SearchPage`

Feichtmeier:
- move search from explore model to app model
- pass search results down from app model
- select values from app model in search and explore page
- use the yaruwindowtitlebar above the yarunavigationpage
- remove yaruwindowtitlebars in the child pages

Fixes #922
Fixes #921

### DIFF
--- a/lib/app/common/search_field.dart
+++ b/lib/app/common/search_field.dart
@@ -28,10 +28,14 @@ class SearchField extends StatefulWidget {
     required this.onChanged,
     this.autofocus = true,
     this.hintText,
+    this.onSubmitted,
+    this.clear,
   });
 
   final String? searchQuery;
   final Function(String value) onChanged;
+  final void Function(String?)? onSubmitted;
+  final void Function()? clear;
   final bool autofocus;
   final String? hintText;
 
@@ -80,7 +84,11 @@ class _SearchFieldState extends State<SearchField> {
     return KeyboardListener(
       onKeyEvent: (value) {
         if (value.logicalKey == LogicalKeyboardKey.escape) {
-          _clear();
+          if (widget.clear != null) {
+            widget.clear!();
+          } else {
+            _clear();
+          }
         }
       },
       focusNode: FocusNode(),
@@ -94,6 +102,7 @@ class _SearchFieldState extends State<SearchField> {
               autofocus: widget.autofocus,
               controller: _controller,
               onChanged: onChanged,
+              onSubmitted: widget.onSubmitted,
               textInputAction: TextInputAction.send,
               decoration: InputDecoration(
                 filled: true,
@@ -112,7 +121,7 @@ class _SearchFieldState extends State<SearchField> {
                             child: Material(
                               color: Colors.transparent,
                               child: InkWell(
-                                onTap: _clear,
+                                onTap: widget.clear ?? _clear,
                                 child: Center(
                                   child: Icon(
                                     YaruIcons.edit_clear,
--- a/lib/app/app.dart
+++ b/lib/app/app.dart
@@ -27,6 +27,7 @@ import 'package:software/app/common/close_confirmation_dialog.dart';
 import 'package:software/app/common/connectivity_notifier.dart';
 import 'package:software/app/common/page_item.dart';
 import 'package:software/app/common/rating_model.dart';
+import 'package:software/app/common/search_field.dart';
 import 'package:software/app/explore/explore_page.dart';
 import 'package:software/app/installed/installed_page.dart';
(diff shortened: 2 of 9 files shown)


d-loose:
> add back button if any child page navigated until an app page (is this even possible?)
Can't we just add a back button to `AppPage`?
Not sure about the layout though - I guess we don't want to add an empty app bar with only a back button.

Feichtmeier:
> > add back button if any child page navigated until an app page (is this even possible?)
> Can't we just add a back button to `AppPage`?
> Not sure about the layout though - I guess we don't want to add an empty app bar with only a back button.
> 
> 

This will look weird tried it. If it's not possible I guess we have to. Depending what's more important for the other designers

lib/app/common/search_field.dart
```diff
@@ -28,10 +28,14 @@ class SearchField extends StatefulWidget {
...
     this.autofocus = true,
     this.hintText,
+    this.onSubmitted,
+    this.clear,
   });
 
   final String? searchQuery;
   final Function(String value) onChanged;
+  final void Function(String?)? onSubmitted;
+  final void Function()? clear;
```
d-loose, line 38: nitpick: can we name this `onClear` for the sake of consistency?

Feichtmeier: Absolutely 🤗

d-loose:
Ok, apart from the back button, which is just a design issue, I think I'm done with it :)

Feichtmeier:
> Ok, apart from the back button, which is just a design issue, I think I'm done with it :)

Thanks for the help

The back button... How could this work in theory? If it's inside app and either explore or installed page would use SnapPage.push
?

d-loose:
Just add a button to `AppPage` that does `Navigator.of(context).maybePop()`? Or am I missing something?

d-loose:
Ah, now I see the issue - I thought we could add it anywhere :D

d-loose:
I don't see a simple way of doing this without exposing the navigator key of `YaruNavigationPage`.

Feichtmeier:
> I don't see a simple way of doing this without exposing the navigator key of `YaruNavigationPage`.

Yeah I feared this is not easy

But you think it's possible? 
We should rework our yaruwidgets Navi pages to inherit a navigator IMO, some day (not now), since this use case of nested navigation won't be rare

d-loose:
If we allow `Yaru*Page` to accept a `NavigatorKey` as a parameter (and it then builds its navigator with that key), we can use it to access and manipulate the navigator state from outside, no?

Feichtmeier:
> If we allow `Yaru*Page` to accept a `NavigatorKey` as a parameter (and it then builds its navigator with that key), we can use it to access and manipulate the navigator state from outside, no?
> Not sure if that's great design though.

I don't think there is a different way to get global search without moving the titleBar into _App
So if we want the global from anywhere to anywhere search I think this is needed 

We can double check again with ana and mads ofc, but I'm sure the they wanted it before you make the work

Feichtmeier:
I just realized that I can just put exploremodel into the multiprovider of app -.-
so half of this PR is basically pointless and just clutters AppModel
hmmm

d-loose:
Ah, that's what I did in #914 - I thought you wanted to follow a different approach because of those unified search plans.

Feichtmeier:
> Ah, that's what I did in #914 - I thought you wanted to follow a different approach because of those unified search plans.

No I've just forgotten that multiproviders exist ...
Tested it locally. This totally works with explore model sitting inside the multiprovider of appmodel. Explore model is there to... explore/search. So I guess this is fine
Do you see any downside of this small change? Then I would make a PR with this change only

Feichtmeier:
https://github.com/ubuntu-flutter-community/software/pull/945

Ok then I can in the end split this PR into smaller parts :+1: :smile_cat:

Feichtmeier:
@d-loose I would say let's wait for ana and mads meeting on monday because I don't how the search behaviour should work and if we really need it when we do #960 
since settings is a dialogue now and if #960 is done we only have 1 "not-explore page" page, which would be something called "library" or "collection" or whatever name we come up with

Are the commits you done here easily re-doable in the future? Then I close this PR here since resolving the conflicts is annoying and I could better start in a fresh branch since all the changes except the move of the search bar inside the tree is done in little PRs parallel to this or before

d-loose:
:+1:
I'm sure we can cherry-pick some of the commits to salvage a bit of the work done here.

PR #938, opened 2023-02-01
AppModel: use context select

Feichtmeier:
Ref #609 

Still works as expected (badges ans spinner)
grafik
grafik


d-loose (approved):
:+1:

PR #944, opened 2023-02-02
Packagekit progress indicator

d-loose:
PackageKit equivalent of #916.
Depends on https://github.com/canonical/packagekit.dart/pull/26.

[Screencast from 2023-02-02 15:56:06.webm](https://user-images.githubusercontent.com/113362648/216359188-9f6adcd7-b5b1-401e-8625-df6fddc11037.webm)

Ref #906

### DIFF
--- a/lib/app/collection/collection_page.dart
+++ b/lib/app/collection/collection_page.dart
@@ -422,10 +422,6 @@ class _PackageTile extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final packageState = context.select((PackageModel m) => m.packageState);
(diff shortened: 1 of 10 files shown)


Feichtmeier (approved):
Works! Thank you! :muscle: 
grafik
grafik
Eventually we need listening to the installed packages stream here and rebuild the list after it was removed
grafik
But this can be done elsewhere :+1:

d-loose:
> Eventually we need listening to the installed packages stream here and rebuild the list after it was removed

This worked after #983 but maybe changes in #989 broke it again.
Maybe we should add some widget tests :>

PR #950, opened 2023-02-03
Show deb dependency sizes

### LINKED ISSUE #943: Show size of dependencies when installing debian packages
Some small packages might pull in a lot of large dependencies - we should provide the total download size.
image

d-loose:
image

Close #943

### DIFF
--- a/lib/services/packagekit/package_service.dart
+++ b/lib/services/packagekit/package_service.dart
@@ -678,13 +678,13 @@ class PackageService {
   Future<void> getDependencies({
     required PackageModel model,
   }) async {
-    Map<PackageKitPackageId, PackageKitInfo> dependencies = {};
+    Map<PackageKitPackageId, PackageKitInfo> dependencyInfos = {};
     if (model.packageId == null) return;
     final dependsOnTransaction = await _client.createTransaction();
     final dependsOnCompleter = Completer();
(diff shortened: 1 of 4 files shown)


lib/services/packagekit/package_service.dart
```diff
@@ -694,9 +694,29 @@ class PackageService {
...
+            info: dependencyInfos[event.packageId] ?? PackageKitInfo.unknown,
+            size: event.size,
+          ),
+        );
+      } else if (event is PackageKitFinishedEvent) {
+        getDetailsCompleter.complete();
+      }
+    });
+    await getDetailsTransaction.getDetails(dependencyInfos.keys);
+    await getDetailsCompleter.future;
```
Feichtmeier, line 718: You could cancel the sub here. I've forgotten this when I added deps

Feichtmeier (changes requested):

lib/services/packagekit/package_service.dart
d-loose: Right, good point!

Feichtmeier: Sometimes "I have my moments" 😂

Feichtmeier (approved):

PR #953, opened 2023-02-03
Settings: move to trailing & open in dialog

Feichtmeier:
I think it would be better to open the settings page without the need to leave the current page.

Any veto @anasereijo @madsrh @Jupi007 @d-loose  or @jpnurmi ? :smile_cat: 

Edit: if we merge this I will change the other dialogs on this page to be inside the settings dialog instead of opening a new one

grafik
grafik

### DIFF
--- a/lib/app/app.dart
+++ b/lib/app/app.dart
@@ -240,14 +240,7 @@ class __AppState extends State<_App> {
               style: itemStyle,
               onTap: () => showDialog(
                 context: context,
-                builder: (context) => AlertDialog(
-                  title: YaruDialogTitleBar(
-                    title: SettingsPage.createTitle(context),
-                  ),
-                  titlePadding: EdgeInsets.zero,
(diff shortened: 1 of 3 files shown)


Jupi007:
Imo, it is a good idea 👍

madsrh:
Much better! That page always felt a bit disconnected from the others 👍

Feichtmeier:
Cool! 4 Thumbs up :sob: :star: 

Navigating to repo settings and about (minus licenses) is now in a navigator:

grafik

Sources: ->

grafik

About: ->

grafik

I think this is enough for one PR

Feichtmeier:
woops, one tiny thing, background color for the darktheme :P 

grafik

madsrh (approved):

whiskeyPeak:
Wow, this looks so much better!

PR #963, opened 2023-02-04
Changed font size for app_header

madsrh:
Use the same font weight as expander headlines. Not sure if using bold font here is the correct solution. WDYT? @Feichtmeier @Jupi007 

Fixes https://github.com/ubuntu-flutter-community/software/issues/962 

Also slightly smaller fontsize for the publishername.

image

### DIFF
--- a/lib/app/common/app_page/app_header.dart
+++ b/lib/app/common/app_page/app_header.dart
@@ -65,7 +65,7 @@ class BannerAppHeader extends StatelessWidget {
                   children: [
                     Text(
                       appData.title,
-                      style: Theme.of(context).textTheme.titleLarge,
+                      style: theme.textTheme.titleLarge,
                     ),
                     PublisherName(
                       height: 14,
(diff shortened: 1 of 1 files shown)


lib/app/common/app_page/app_header.dart
```diff
@@ -65,13 +65,10 @@ class BannerAppHeader extends StatelessWidget {
                   children: [
                     Text(
                       appData.title,
-                      style: theme.textTheme.displaySmall!.copyWith(
-                        fontSize: 20,
-                        color: theme.colorScheme.onSurface,
-                      ),
+                      style: Theme.of(context).textTheme.titleLarge,
```
Feichtmeier, line 68: Instead of `Theme.of(context)` use the `theme` variable. Then the analyse warning goes away

Feichtmeier (changes requested):

Feichtmeier (approved):

PR #965, opened 2023-02-04
LicensePage tweaks

madsrh:
Small tweaks for license page:

- Removed duplicate back button
- Use full window height (removed SizedBox)

Before:
image

After:
image


Feichtmeier:
I used the sizedbox because the license page leaks over the edge
grafik

you could alternatively wrap it in a ClipRRect

You also got an analyse warning :)

Feichtmeier (approved):

PR #968, opened 2023-02-05
SettingsModel: clean up

Feichtmeier:
- make properties nullable
- make const private to not show up in code completion elsewhere


Jupi007 (approved):
LGFM 👍

PR #970, opened 2023-02-05
SnapUpdates: refresh all should not ask for password before every update

### LINKED ISSUE #969: SnapUpdates: refresh all should not ask for password before every update

Feichtmeier:
https://user-images.githubusercontent.com/15329494/216808762-4eec5bf0-2107-4d85-aeec-61021ded83e4.mp4

Fixes #969

### DIFF
--- a/lib/app/updates/snap_updates_model.dart
+++ b/lib/app/updates/snap_updates_model.dart
@@ -76,8 +76,7 @@ class SnapUpdatesModel extends SafeChangeNotifier {
     required String doneMessage,
   }) async {
     await _snapService.authorize();
-    if (_snapsWithUpdates == null ||
-        _snapsWithUpdates != null && _snapsWithUpdates!.isEmpty) return;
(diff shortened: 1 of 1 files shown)


d-loose:
I haven't encountered this issue before - what am I supposed to be looking for in your screen recording?
Do I understand correctly that calling `_snapService.refresh()` concurrently on all snap with updates will trigger a password prompt for all of them, but awaiting the first call and doing the remaining ones afterwards only triggers it once?

Feichtmeier:
> I haven't encountered this issue before - what am I supposed to be looking for in your screen recording? Do I understand correctly that calling `_snapService.refresh()` concurrently on all snap with updates will trigger a password prompt for all of them, but awaiting the first call and doing the remaining ones afterwards only triggers it once?

yes that's what's happening. this is a trick as I am not sure how it could be done differently

d-loose (approved):
LGTM :+1:

PR #973, opened 2023-02-05
AppPage Gallery Dialog: use appData.name as title

Feichtmeier:
grafik

Sorry, forgotten to change the title :)


Jupi007 (approved):
Indeed, I also didn't noticed it... 😅

PR #974, opened 2023-02-05
Merge updates and installed pages

### LINKED ISSUE #960: merge updates and installed
Actually, I had another look just now, and could see what happens when we have updates available. It will just display the packages with updates available, and clicking back on it will display all the installed apps again. Makes me wonder if we should merge updates and installed?

https://user-images.githubusercontent.com/36884067/215578875-b7fb649b-d97e-456a-8d51-576986957031.mov

_Originally posted by @anasereijo in https://github.com/ubuntu-flutter-community/software/issues/913#issuecomment-1409232696_

Feichtmeier:
Except design acknowledgement and design improvements, the following things need to be sorted out before this can go live:

- [x] add deb page
- [x] make snap updating more reliable on page changes

Snaps
grafik
grafik

Packages
grafik
grafik
grafik

(note: navigation is not consistent in this screenshots because I did not remove installed/updates items yet for less merge conflicts)

Fixes #960


Feichtmeier:
@d-loose 

to test this you need to

```dart
PageItem(
        titleBuilder: CollectionPage.createTitle,
        builder: (context) => CollectionPage.create(context),
        iconBuilder: (c, s) => CollectionPage.createIcon(
          context: c,
          selected: s,
          badgeCount: badgeCount,
          processing: processing,
          updateCount: updateAmount,
          updateProcessing: updatesProcessing,
        ),
      ),
      ```

File: lib/app/collection/collection_page.dart [hunk 1]
```diff
@@ -0,0 +1,359 @@
...
+    final searchQuery = context.select((CollectionModel m) => m.searchQuery);
+
+    final checkingForSnapUpdates =
+        context.select((CollectionModel m) => m.checkingForSnapUpdates);
+
+    final snaps = searchQuery == null || searchQuery.isEmpty
+        ? installedSnaps.entries
+        : installedSnaps.entries.where(
+            (element) => element.key.name.contains(searchQuery),
+          );
```
d-loose, line 211: this logic should probably be in the collection model

d-loose:
`simple_snap_*.dart` have a lot of overlap with some existing code - could we somehow simplify what we already have and reuse it here?

Feichtmeier:
> `simple_snap_*.dart` have a lot of overlap with some existing code - could we somehow simplify what we already have and reuse it here?

Absolutely.  I just don't want to change too much of the existing pages for now

Feichtmeier:
The bigger model and controls have all the snap channel stuff which we don't need here which takes computation time in a big list. But generally I duplicated a lot of code here which needs to be trimmed

d-loose:
> Absolutely. I just don't want to change too much of the existing pages for now

How about we continue this in a separate branch where we remove the old pages and replace them by the new one? Adding new pages that are still under construction and not visible in the UI can add a lot of clutter that we'd need to clean up later.

lib/app/collection/collection_page.dart
Feichtmeier: Thanks! Done now
grafik

Feichtmeier:
> 

Not sure if I understand correctly. You want to merge this first or work in two PRs?

d-loose:
I wouldn't want to merge these 'invisible' changes, but rather work on them until they're ready (either in this PR branch or in a separate feature branch if it becomes too much). Keeping the current Installed/UpdatesPages side by side with the new CollectionPage seems to create too much clutter imo.

Feichtmeier:
> I wouldn't want to merge these 'invisible' changes, but rather work on them until they're ready (either in this PR branch or in a separate feature branch if it becomes too much). Keeping the current Installed/UpdatesPages side by side with the new CollectionPage seems to create too much clutter imo.

sure! then I remove them here, all features are in. OK?

lib/app/collection/collection_model.dart
```diff
@@ -183,7 +183,7 @@ class CollectionModel extends SafeChangeNotifier {
     if (!_packageService.isAvailable) {
       return [];
     } else {
-      if (searchQuery?.isEmpty == true) {
+      if (searchQuery == null || searchQuery?.isEmpty == true) {
```
d-loose, line 186:
```dart
if (searchQuery?.isEmpty ?? true)
```

d-loose:
> sure! then I remove them here, all features are in. OK?

That would be great - even better if you could find a way to simplify and re-use the snap model :)

Feichtmeier:
> > sure! then I remove them here, all features are in. OK?
> 
> That would be great - even better if you could find a way to simplify and re-use the snap model :)

the problem with the snapmodel is, that it is really bloated. I made it when I was relatively new to this API and half of the code in there is very bad and now we can do this better by either adding extensions on snapd.dart snaps or with the snap_utils.dart for all the channel stuff which is really complicated, thanks to an API only made with strings :D 

However I agree that this should be somehow synced. Can we do this in another PR? I removed all the unused code now and moved the bits used to collections dir

d-loose:
> However I agree that this should be somehow synced. Can we do this in another PR? I removed all the unused code now and moved the bits used to collections dir

Yes, de-bloating sounds great!

Feichtmeier:
> > However I agree that this should be somehow synced. Can we do this in another PR? I removed all the unused code now and moved the bits used to collections dir
> 
> Yes, de-bloating sounds great!

Cool!

All tests pass locally btw, but the CI is down today for flutter it seems
grafik

d-loose:
Yes, test failures are related to server outages.

Feichtmeier:
Whoopsie I have forgotten the remove button for the debs :P 

in now!
grafik

Plus debs are now selected first if available, as before in Updates. Was a request by @madsrh at that time

lib/app/collection/collection_model.dart
```diff
@@ -0,0 +1,239 @@
...
+  bool? _loadPackagesWithUpdates;
+  bool? get loadPackagesWithUpdates => _loadPackagesWithUpdates;
+  void setLoadPackagesWithUpdates(bool? value) {
+    if (value == null || value == _loadPackagesWithUpdates) return;
+    _loadPackagesWithUpdates = value;
+    notifyListeners();
+  }
+
+  final Set<PackageKitFilter> _packageKitFilters = {
+    PackageKitFilter.installed,
```
d-loose, line 218: user shouldn't be able to remove `PackageKitFilter.installed` here

Feichtmeier:
Nice mock-ups!

Yes we can do this now much easier. If this will be merged please create a dedicated issue

lib/app/collection/collection_model.dart
Feichtmeier: You can't, it's disabled in the pop-up

Feichtmeier: woops I had some funny flag that was not enabled :) now it is! Thanks!

Feichtmeier:
@madsrh so now you would rather display the snap page first?

d-loose (approved):
Thanks a lot, LGTM now :+1:

Feichtmeier:
Thank you for the thorough review :+1:

Issue #954, opened 2023-02-03
Add an easy way to return to installing/updating/removing snaps/debs


alculete:
It would be great to have a way to go back to the applications that are still installing. An option should be added to the menu when a new installation is in progress, so I can navigate away and then return to see the installation status.

Feichtmeier:
@alculete thanks!

We made some changes and merged updates and installed.

Then we have an idea ongoing for showing the updating section above the installed section like in @madsrh  mock-up here:

Feichtmeier:
I think this is fixed by the new collection page layout

alculete:
IHi @Feichtmeier  This bug has returned. Updating snaps are shown but new snaps installation are not.

PR #975, opened 2023-02-07
Collection: sort snaps by update and SnapSort

Feichtmeier:
grafik

- bring back snap sort popup
- always show updates at top (if any are there)

Ref #954


madsrh (approved):
Yes! this looks good!

Issue #868, opened 2023-01-24
Application shows a grey box when first opened, only resolved on relaunch or resize


local-optimum:
[Screencast from 24-01-23 13:43:01.webm](https://user-images.githubusercontent.com/55029393/214310791-05161bef-3833-4d75-ac43-212727f68158.webm)

Consistently, when I open the new application it displays a completely grey box, and only loads the inner assets when resizing the window or relaunching the app. This happens more than 50% of the time for me.

I have also had situations where the minimise, maximise, close buttons have disappeared though this is harder to reproduce.

Feichtmeier:
THat would be bad.I try to reproduce but this is not happening for me with the latest snap

https://user-images.githubusercontent.com/15329494/214312892-f2b971a8-6f6d-47f3-a64c-d228e08aa59e.mp4

@d-loose @madsrh could you try to reproduce this?

jpnurmi:
May be related to:
- https://github.com/flutter/engine/pull/38982

Feichtmeier:
I can't reproduce this on any of my computers though. COuld this be related to 22.10? @local-optimum what snap version and what ubuntu version to you use in this screencast?

Feichtmeier:
fresh lunar VM, I can not reproduce this issue

![image](https://user-images.githubusercontent.com/15329494/214599958-feaf69d3-c18b-4758-b5cb-64ce0730ef1c.png)

@local-optimum do you eventually use the noveau driver and an old nvidia card?

d-loose:
Couldn't reproduce it either so far

local-optimum:
I'm using an NVIDIA 1080 with the 525 proprietary driver. This isn't specific to this app, it happens to all Flutter apps on my machine...

Feichtmeier:
> I'm using an NVIDIA 1080 with the 525 proprietary driver. This isn't specific to this app, it happens to all Flutter apps on my machine...

Holy crap that's no good

All flutter snaps also from the store? Like the flutter gallery?

local-optimum:
Flutter gallery and QuickGUI I tried, both of them show the blank screen initially but eventually load the window contents (~1s), looks like only this app never displays. QuickGUI has in the past needed a resize but can't repro that atm.

Feichtmeier:
Okay 👍 thanks for investigating

Let's keep it opened here until we know what's happening

My tip is on the graphic card driver

local-optimum:
[...]

PR #978, opened 2023-02-07
Attempt to fix occasional blank window

jpnurmi:
- Use `HdyWindow` directly from libhandy instead of relying on the widget tree restructuring magic by `handy_window`
- Restore the original order for showing the window and view, as it is in the Flutter app template

Probably fixes: #868

### DIFF
--- a/linux/CMakeLists.txt
+++ b/linux/CMakeLists.txt
@@ -6,6 +6,8 @@ set(APPLICATION_ID "io.snapcraft.Store")
 
 cmake_policy(SET CMP0063 NEW)
 
+set(USE_LIBHANDY ON)
+
 set(CMAKE_INSTALL_RPATH "$ORIGIN/lib")
 
 # Configure build options.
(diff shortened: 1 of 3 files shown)


jpnurmi:
NOTE: Marked as draft because this depends on the following unpublished changes:
- https://github.com/leanflutter/window_manager/pull/297
- https://github.com/ubuntu-flutter-community/handy_window/pull/28

Feichtmeier:
Thanks for this huge and in depth work!

I guess we need someone who could reproduce this problem in the first place to test this. I never could so I can't review this.

Feichtmeier:
Or you could give me any tip how to reproduce this outside of Oliver's laptop :)

X11 session, plus? Slow connection to the initial loading takes so long that the window isn't loaded in time?

jpnurmi:
This is not X11-specific. I was able to reproduce this on Wayland. :) It's not easy to reproduce or verify because the bug is seldom and random but as @d-loose pointed out in #868, running in release mode may increase your chances to reproduce it.

Feichtmeier:
Well well, tried appr. 20 times from the console and never happened to me. I guess @d-loose is a great candidate to review this then :)

d-loose (approved):
LGTM :+1:

PR #979, opened 2023-02-07
PackageService: remove filter in isInstalled

### LINKED ISSUE #976: Package model wrongly reports deb package is installed after removing
[Screencast from 2023-02-07 11:28:50.webm](https://user-images.githubusercontent.com/113362648/217220245-161ee350-bac3-4db2-9e67-0c211f6ce761.webm)

Running `apt autoremove` to remove installed dependencies seems to resolve it.

d-loose:
Uninstalled packages need to be included in the search to correctly determine whether a package is installed - otherwise it will find installed packages with a similar name, e.g. leftover dependencies like [packagename]-data.

Fix #976

### DIFF
--- a/lib/services/packagekit/package_service.dart
+++ b/lib/services/packagekit/package_service.dart
@@ -484,7 +484,6 @@ class PackageService {
     });
     transaction.searchNames(
       [model.packageId!.name],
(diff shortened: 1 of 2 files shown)


Feichtmeier (approved):
ty

PR #981, opened 2023-02-07
PackageService: fix isInstalled

d-loose:
Sorry, this needs another fix:
with multiple matches it reports the 'installed' status of the last package event, even if the name doesn't match.


Feichtmeier (commented):
TY

Feichtmeier:
Eeehrg, sorry for not testing this more thoroughly : /

I think something went wrong here.

Installed list:
grafik

One tile selected:
grafik

Issue #980, opened 2023-02-07
Dependency removal


d-loose:
When removing a debian package we should provide the option to remove dependencies that were installed alongside the package. Probably a simple dialog would be sufficient.
This functionality is already provided by [PackageKit](https://www.freedesktop.org/software/PackageKit/gtk-doc/Transaction.html#Transaction.RemovePackages) and implemented in packagekit.dart.

PR #987, opened 2023-02-07
Add dependency removal dialog

d-loose:
Adds a simple dialog, that allows removing unneeded dependencies when removing a deb.
image

Possible improvements:
* Add this to the `CollectionPage` as well (@Feichtmeier I've seen that you've avoided calling `PackageModel.init` for the list tiles to reduce the number of unnecessary PackageKit queries. We'd need to know whether each of the packages have dependencies here, though. Maybe this could be done after clicking the 'remove' button - might need some refactoring.)
* Add more details about the dependencies to be removed. If I understand PackageKit correctly, it doesn't give us enough information about a package's dependencies to figure out whether or not they're needed by other packages as well, and thus wouldn't be automatically removed. Calling PackageKit's `RequiredBy` on each dependency seems like a lot of effort though.

Edit:
Thanks to the integration test I realize that showing this dialog for every package that has dependencies isn't very clever..

Ref #980

### DIFF
--- a/lib/app/common/packagekit/package_controls.dart
+++ b/lib/app/common/packagekit/package_controls.dart
@@ -29,7 +29,9 @@ class PackageControls extends StatelessWidget {
     required this.packageState,
     required this.versionChanged,
     this.hasDependencies,
-    this.showDeps,
+    this.hasMissingDependencies,
+    this.showDepsAndInstall,
+    this.showDepsAndRemove,
   });
 
   final bool? isInstalled;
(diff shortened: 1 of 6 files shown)


anasereijo:
Maybe we can display what dependencies will be removed, similarly to when we display the ones to be installed? With that [expand/collapse](https://github.com/ubuntu-flutter-community/software/issues/739) section?

d-loose:
I'm a bit pessimistic about this. PackageKit doesn't provide (easily accessible) information about the dependencies that would be automatically removed. Maybe a viable option could be introducing a setting that controls whether or not to remove them.
gnome-software [automatically removes dependencies](https://github.com/GNOME/gnome-software/blob/main/plugins/packagekit/gs-plugin-packagekit.c#L694) depending on a [flag set at build time](https://github.com/GNOME/gnome-software/blob/main/meson.build#L200).

PR #992, opened 2023-02-08
PackageService: fix isInstalled

d-loose:
third time's the charm..


Feichtmeier:
Works!

grafik

Did you check if it still it works that the package installer understands that you opened a newer package than you got installed? Example discord has a new package you downloaded and opened it with snap-store =)

d-loose:
> Did you check if it still it works that the package installer understands that you opened a newer package than you got installed? Example discord has a new package you downloaded and opened it with snap-store =)

Yes, just tested it :)

Feichtmeier (approved):

PR #997, opened 2023-02-08
CollectionModel: make _installedSnaps null at start

### LINKED ISSUE #994: CollectionModel: make _installedSnaps null at start
For a glance "no snaps installed" is shown until the list is filled. 
Again I made the mistake to init a prop in the model.... 
We need null values to have a "waiting"-"but not yet checked if really empty" state in the UI

Feichtmeier:
Fixes #994

### DIFF
--- a/lib/app/collection/collection_model.dart
+++ b/lib/app/collection/collection_model.dart
@@ -99,7 +99,7 @@ class CollectionModel extends SafeChangeNotifier {
   List<Snap>? get installedSnapsWithUpdates => _snapService.snapsWithUpdate;
 
   bool get snapUpdatesAvailable =>
-      (_installedSnaps?.isEmpty ?? false) ? false : true;
+      installedSnapsWithUpdates?.isNotEmpty ?? false;
 
   bool? _snapServiceIsBusy;
   bool? get snapServiceIsBusy => _snapServiceIsBusy;
--- a/lib/app/collection/collection_model.dart
+++ b/lib/app/collection/collection_model.dart
@@ -25,9 +25,7 @@ class CollectionModel extends SafeChangeNotifier {
   Future<void> init() async {
     _snapChangesSub = _snapService.snapChangesInserted.listen((_) async {
       if (_snapService.snapChanges.isEmpty) {
-        _installedSnaps.clear();
-        _loadInstalledSnaps();
-        snapServiceIsBusy = false;
+        await loadSnaps();
       } else {
         snapServiceIsBusy = true;
       }
@@ -48,11 +46,10 @@ class CollectionModel extends SafeChangeNotifier {
       notifyListeners();
     }
 
-    await _loadInstalledSnaps();
-    notifyListeners();
+    await loadSnaps();
 
     if (_snapService.snapChanges.isEmpty) {
-      await checkForSnapUpdates();
+      await loadSnaps();
     } else {
       checkingForSnapUpdates = false;
     }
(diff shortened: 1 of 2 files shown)


lib/app/collection/collection_model.dart
```diff
@@ -84,27 +81,25 @@ class CollectionModel extends SafeChangeNotifier {
...
 
-  List<Snap> get installedSnapsWithUpdates {
-    final entryList = _installedSnaps.entries.toList().where((e) => e.value);
-    return entryList.map((e) => e.key).toList();
-  }
+  List<Snap>? get installedSnapsWithUpdates => _snapService.snapsWithUpdate;
 
   bool get snapUpdatesAvailable =>
-      _installedSnaps.entries.where((e) => e.value == true).toList().isNotEmpty;
+      (_installedSnaps?.isEmpty ?? false) ? false : true;
```
d-loose, line 102: Shouldn't this check whether `installedSnapsWithUpdates` has entries (rather than `_installedSnaps`)?

Simplification:
```dart
bool get snapUpdatesAvailable => installedSnapsWithUpdates?.isNotEmpty ?? false;
```

Feichtmeier: woops! you are right! out of my intension this does not result in an error because if there ain't any snaps there ain't any updates xD but I change this ofc! TY

lib/app/collection/collection_model.dart
```diff
@@ -114,17 +109,21 @@ class CollectionModel extends SafeChangeNotifier {
     notifyListeners();
   }
 
-  List<Snap> get _snapsWithUpdates => _installedSnaps.entries
-      .where((e) => e.value == true)
-      .map((e) => e.key)
-      .toList();
+  List<Snap>? get _snapsWithUpdates => _snapService.snapsWithUpdate;
```
d-loose, line 112: `_snapService.snapsWithUpdate` isn't nullable, so `_snapsWithUpdates` shouldn't be either (also the names should match, or just use `_snapService.snapsWithUpdate` without the private getter)

lib/app/collection/collection_model.dart
```diff
@@ -114,17 +109,21 @@ class CollectionModel extends SafeChangeNotifier {
...
 
-  Future<void> _loadInstalledSnaps() async {
+  Future<void> loadSnaps() async {
+    checkingForSnapUpdates = true;
+    _installedSnaps = null;
+    _snapsWithUpdates?.clear();
```
d-loose, line 117: This calls `clear()` on a list that belongs to the snap service - is this necessary, or is there a better way?

lib/app/collection/collection_model.dart
```diff
@@ -48,11 +46,10 @@ class CollectionModel extends SafeChangeNotifier {
       notifyListeners();
     }
 
-    await _loadInstalledSnaps();
-    notifyListeners();
+    await loadSnaps();
 
     if (_snapService.snapChanges.isEmpty) {
-      await checkForSnapUpdates();
+      await loadSnaps();
```
d-loose, line 52: Why call `loadSnaps()` a second time here?

Feichtmeier: no idea, brain fart

lib/app/collection/collection_model.dart
Feichtmeier: Yes this should be inside the service but I tried to limit the changes to the model here. Can I do this in a second PR?

lib/app/collection/collection_model.dart
Feichtmeier: ehhh, another brainfart. Duplicate getter

lib/app/collection/collection_model.dart
Feichtmeier: the whole getter is not needed. I think I refactored too much around and ended up with those duplicates. Sorry and thanks for double checking! Removed it

d-loose (approved):
Thanks, LGTM now :+1+

PR #999, opened 2023-02-08
CollectionModel: remove unecessary list clearing

Feichtmeier:
Was abs. not needed. It's null on start that is correct :+1: 
Afterwards the service takes care


d-loose (approved):
:+1:

PR #1000, opened 2023-02-08
App: add more shortcut pageItems

Feichtmeier:
@local-optimum I think you mentioned the importance of "iot" and "server and cloud" somwhere (was it in the meeting?)

This PR adds two more shortcuts to the sidebar

- iot
- server and cloud

grafik
grafik

We can add more in the future if you like

### DIFF
--- a/lib/app/app.dart
+++ b/lib/app/app.dart
@@ -163,7 +163,7 @@ class __AppState extends State<_App> {
           .firstOrNull
           ?.substring(7);
       if (debPath != null || snapName != null) {
-        _initialIndex = 7;
+        _initialIndex = 6;
       }
     });
   }
@@ -191,6 +191,17 @@ class __AppState extends State<_App> {
           context: context,
           selected: selected,
           snapSection: SnapSection.all,
+        ),
+      ),
+      PageItem(
+        titleBuilder: (context) =>
+            ExplorePage.createTitle(context, SnapSection.productivity),
+        builder: (context) =>
+            const ExplorePage(section: SnapSection.productivity),
+        iconBuilder: (context, selected) => ExplorePage.createIcon(
+          context: context,
+          selected: selected,
+          snapSection: SnapSection.productivity,
         ),
       ),
       PageItem(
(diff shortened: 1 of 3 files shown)


local-optimum:
So I don't really think server and IoT are that relevant as a sidebar links. I think Productivity should go first as most of our most popular apps are things like Telegram and browsers. Then Development, then Games and I would suggest a 4th icon that represents 'categories' to take you to the full category list -  most stores seem to represent this by four mini-squares.

Feichtmeier:
@local-optimum thanks for the input

The "categories" item you mentioned should be basically "Explore". But this will be included in @anasereijo rework I think. Like a "endless" scrolling experience like in a magazine mode with category tiles included (at least this is how I think of it, without having a concrete mock-up). But ofc we could keep the explore page more like the endless scrolling experience and add the "Categories" item on-top of this list

You didn't mention art & design, but I think those apps are also quite popular among ubuntu users. Don't you think?

anasereijo:
>  I would suggest a 4th icon that represents 'categories' to take you to the full category list - most stores seem to represent this by four mini-squares.

@local-optimum we can do that, however, I think having it on 'Explore' could be enough – I see it as a page where users will literarily explore and discover everything we have to offer, including categories. On the Explore page, we will have a mix of curated 'collection' banners and a section showing all the categories available (linked to each category page).
I looked at what 🍏 and 🪟 and they both have a categories page, but IMO it's quite the same as the Explore page but focused just on categories. We can explore this of course, but just a thought!

> You didn't mention art & design, but I think those apps are also quite popular among ubuntu users. Don't you think?
@Feichtmeier Maybe that could be the 4th one?

madsrh:
I wonder why Server and Cloud is on this list. Aren't most of these applications without GUI, so would you really install them from the store? IDK, I think Development would be more useful. 

> art & design, but I think those apps are also quite popular among ubuntu users

+1


BTW, Congrats on the PR #1000 🎉 🍾

Feichtmeier:
((( Note: when one searched there is still the full list of all categories )))

Feichtmeier:
@local-optimum @anasereijo @madsrh 

grafik

Any idea on the icon?

grafik

grafik

?

On the discussion: how about this list of sidebar items now?

- explore (all categories)
- productivity (category)
- development (category)
- games (category)
- art & design (category)

anasereijo:
@Feichtmeier I think it will be better if we look at all iconography for categories in one issue/discussion, to see how it all looks together. And then we can then also check what @Zoospora thinks about it as she might have some suggestions.

But, as a quick 'decision', between the 3 options I prefer the one with the paper plane. I checked the current store, and have the following. Maybe something to consider when we look at all?
 
Screenshot 2023-02-09 at 11 17 30

anasereijo:
> On the discussion: how about this list of sidebar items now?
explore (all categories)
productivity (category)
development (category)
games (category)
art & design (category)

I think this is fine, unless @local-optimum wants to change the order to give a bit more 'importance' to some categories.

madsrh:
> explore (all categories)
productivity (category)
development (category)
games (category)
art & design (category)

Yes, I like this.

Feichtmeier:
grafik

@madsrh @anasereijo if you agree, please push the approve button ;)

anasereijo (approved):

Zoospora:
> @Feichtmeier I think it will be better if we look at all iconography for categories in one issue/discussion, to see how it all looks together. And then we can then also check what @Zoospora thinks about it as she might have some suggestions.
> 
> But, as a quick 'decision', between the 3 options I prefer the one with the paper plane. I checked the current store, and have the following. Maybe something to consider when we look at all?
> 
> Screenshot 2023-02-09 at 11 17 30

Hi, I think that the icon mentioned by @anasereijo , taking notes on paper can represent the category quite well. However, the clock doesn't seem to me to be so misguided. I would go for one of these two options, the one on the paper being the one I think best represents it. 

Thanks!!

PR #1004, opened 2023-02-08
CollectionPage: improve performance

Feichtmeier:
- service: use a second list of IDs only for checking the installedId vs updateId versions so the installed list does get overwritten when checking for updates
- collectionpage: replace columns with many elements with listview builders, this needs a second change with scroll controllers and lazy loading in the future
- re-add package filter popup

### DIFF
--- a/lib/services/packagekit/package_service.dart
+++ b/lib/services/packagekit/package_service.dart
@@ -77,13 +77,18 @@ class PackageService {
   final Map<String, PackageKitPackageId> _installedPackages = {};
   List<PackageKitPackageId> get installedPackages =>
       _installedPackages.entries.map((e) => e.value).toList();
-  PackageKitPackageId? getInstalledId(String name) => _installedPackages[name];
   final _installedPackagesController = StreamController<bool>.broadcast();
   Stream<bool> get installedPackagesChanged =>
       _installedPackagesController.stream;
   void setInstalledPackagesChanged(bool value) {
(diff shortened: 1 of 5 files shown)


Feichtmeier:
Note: we need a second change for when one de-selects "application" the list explodes with items. Will implement a lazy loading for the lists after this PR @d-loose

lib/services/packagekit/package_service.dart
```diff
@@ -385,6 +390,34 @@ class PackageService {
     );
   }
 
+  Future<void> getInstalledPackagesForUpdates({
```
d-loose, line 393: Could we refactor this a bit to avoid too much code duplication?
The only difference to `getInstalledPackages` is the map the results are stored in.

d-loose (changes requested):

lib/services/packagekit/package_service.dart
Feichtmeier: ok will do. I avoided to change the signature of this method but maybe I can get around with an optional argument without crashing the tests

d-loose: you could just add a helper method that accepts a parameter and keep the signatures of the public methods intact

Feichtmeier: @d-loose is this solution okay for you? (Parameter, see commit)

d-loose (approved):
:+1:

PR #1006, opened 2023-02-09
`StartPage`: show appstream data as soon as possible

d-loose:
[Screencast from 2023-02-09 16:06:59.webm](https://user-images.githubusercontent.com/113362648/217855766-30062088-2cea-44f7-a42e-f2bb1f872c80.webm)

### DIFF
--- a/lib/app/explore/explore_page.dart
+++ b/lib/app/explore/explore_page.dart
@@ -91,9 +91,8 @@ class _ExplorePageState extends State<ExplorePage> {
     final showSearchPage = context.select((ExploreModel m) => m.showSearchPage);
     final searchQuery = context.select((ExploreModel m) => m.searchQuery);
     final setSearchQuery = context.read<ExploreModel>().setSearchQuery;
-    final snaps = context.select((ExploreModel m) {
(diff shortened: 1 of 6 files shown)


lib/app/explore/explore_page.dart
```diff
@@ -91,9 +91,8 @@ class _ExplorePageState extends State<ExplorePage> {
     final showSearchPage = context.select((ExploreModel m) => m.showSearchPage);
     final searchQuery = context.select((ExploreModel m) => m.searchQuery);
     final setSearchQuery = context.read<ExploreModel>().setSearchQuery;
-    final snaps = context.select((ExploreModel m) {
-      return m.sectionNameToSnapsMap[widget.section];
-    });
+    final startPageApps = context.read<ExploreModel>().startPageApps;
+    context.select((ExploreModel m) => m.startPageAppsChanged);
```
jpnurmi, line 95: i don't have a good idea how to fix this without refactoring. perhaps we should leave it as a separate step or task?

d-loose: Fine with me :)

jpnurmi (approved):
should i stamp also :) :+1:

PR #1024, opened 2023-02-10
CollectionPage: show to top button

Feichtmeier:


Feichtmeier:
:-1: forgot to dispose, changing computers

lib/app/collection/collection_page.dart
```diff
@@ -178,20 +223,22 @@ class CollectionPage extends StatelessWidget {
...
-                  ),
-                ),
-              ),
-            )
+          Expanded(
+            child: Stack(
+              alignment: Alignment.bottomRight,
+              children: [
+                if (appFormat == AppFormat.snap)
+                  const _SnapList()
```
Jupi007, line 231: The scroll controller isn't passed to the `_SnapList` scroll view, and so the fap isn't showing up in the snap view.

Jupi007 (changes requested):
I noticed that, if there was not enough item (like when loading), it was put at the bottom:

image

lib/app/collection/collection_page.dart
```diff
@@ -178,20 +223,22 @@ class CollectionPage extends StatelessWidget {
...
+                else
+                  packageList,
+                if (_showFap)
+                  Padding(
+                    padding: const EdgeInsets.all(kYaruPagePadding),
+                    child: floatingActionButton,
+                  )
```
Jupi007, line 238: Maybe can you use a fade animation instead of append and remove the fap from the tree?

lib/app/collection/collection_page.dart
```diff
@@ -153,6 +181,23 @@ class CollectionPage extends StatelessWidget {
...
+          children: const [PackageUpdatesPage(), _InstalledPackagesList()],
+        ),
+      ),
+    );
+
+    final floatingActionButton = FloatingActionButton(
+      foregroundColor: theme.colorScheme.onInverseSurface,
+      backgroundColor: theme.colorScheme.inverseSurface,
+      shape: const CircleBorder(),
+      onPressed: () => _controller.jumpTo(0),
```
Jupi007, line 197: What do you think of use `animateTo`?
With 500ms and an `easeInOutCubic` curve?

lib/app/collection/collection_page.dart
Feichtmeier: In a separate pr

lib/app/collection/collection_page.dart
Jupi007: Oh, indeed sorry.
I didn't understood it like that :)

lib/app/collection/collection_page.dart
Jupi007: I suppose you would do (or not) this in a separate PR :)

Feichtmeier: 500 was too slow, took 200

Jupi007 (approved):
LGFM :+1:

PR #1029, opened 2023-02-10
Pubspec: Upgrade handy_window

### LINKED ISSUE #1023: Minimum window size can be changed to a size lower than what it is set to in my_application.cc
My laptop has a screen resolution of 1920x1200, so by setting that as the minimum window size, the software centre app should theoretically cover the entire screen on my laptop. This isn't the case.

https://user-images.githubusercontent.com/73116038/218001910-7b605a2e-fb1e-4d1e-9bea-aecaabe5b320.mp4

Feichtmeier:
Fixes #1023

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -32,7 +32,7 @@ dependencies:
   flutter_svg: ^1.1.0
   glib: 0.0.1
   gtk_application: ^0.0.3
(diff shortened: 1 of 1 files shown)


jpnurmi (approved):
LGTM :+1:

PR #1030, opened 2023-02-10
Revert "Pubspec: Upgrade handy_window"

Feichtmeier:
Reverts ubuntu-flutter-community/software#1029

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -32,7 +32,7 @@ dependencies:
   flutter_svg: ^1.1.0
   glib: 0.0.1
   gtk_application: ^0.0.3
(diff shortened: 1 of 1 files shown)


Feichtmeier:
for some reason, integration was no longer a required test for automerge after I played with the merge queue. Reverting this for now. then retrying and fixing the test

Feichtmeier:
it looks like the test succeeded in main. Pump and settle time out

d-loose:
> for some reason, integration was no longer a required test for automerge after I played with the merge queue.

I think it wasn't marked as a requirement before, and maybe it shouldn't be because it can always be a bit flaky due to connection issues and whatnot

PR #1031, opened 2023-02-10
CollectionPage: organize

### LINKED ISSUE #1025: CollectionPage: organize
This file is an absolute mess. It needs refactoring, moving to other files and cleaning up :broom: :1234:

Feichtmeier:
No visual changes, except that the FAB now also shows up for the snap collection if scrolled down

Fixes #1025

### DIFF
--- a/integration_test/software_test.dart
+++ b/integration_test/software_test.dart
@@ -49,7 +49,7 @@ void main() {
       await app.main([]);
       await tester.pumpUntil(
         find.byType(StartPage),
-        timeout: const Duration(seconds: 80),
+        timeout: const Duration(seconds: 120),
       );
       await tester.pumpAndSettle();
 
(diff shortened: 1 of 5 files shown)


Jupi007 (approved):
LGFM 👍
Good job!

Feichtmeier:
@d-loose really weird

when I run this test locally it does not enter the text either
but when I "intervene" in the test window and just enter one letter, after that the tester can insert test and the test continues after that without any intervention
I tried to understand what might have changed the hierarchy in the last commits. The only thing I found was the helper method _createExplorePageItem but after I reverted this the test had the same behaviour :shrug:

Feichtmeier:
The specific part of the test that is failing in this automated integratoin test is working if done by a human. Thus I will merge this now and maybe @d-loose could investigate when he is back :+1:

PR #1035, opened 2023-02-12
Release 0.2.9 into candidate

Feichtmeier:


Feichtmeier:
hello hello this is yaru 1 calling to canonical base anyone here?
what should happen with this? can we just merge to canidate when we think it is ready? @local-optimum @kenvandine @anasereijo ?!?!

Issue #887, opened 2023-01-27
Unnecessary mouse-over pop-ups in App details


local-optimum:
![image](https://user-images.githubusercontent.com/55029393/215069313-ce898e44-453b-426a-a208-ac90b4262a42.png)

The app screens have unnecessary mouseover descriptions duplicating already visible info, this is the case on the Rating/Confinement/Version bar and in the App details bar most noticeable but may be present elsewhere. We should do a pass on which elements may benefit from additional detail on mouseover (eg license or confinement) and provide additional detail or just remove them throughout.

madsrh:
Basically publisher names, version numbers and website urls can be very long and needs a tooltip. The others labels have tooltips for consistency. Wouldn't it seem odd if only some labels have a tooltip? I mean, for publisher, version and url the tooltip are there even is the string is very short. 
Unless you want to do something like: if string is longer than XX characters long, add tooltip 🤷‍♂️ 

The Inkscape version eg. is `1.2.2-b0a8486541-2022-12-01-cust..`. and isn't even fully displayed on the snapcraft website.

Feichtmeier:
Yep this is the reason those strings can be ultra long

PR #1038, opened 2023-02-12
Remove unnecessary toolips

madsrh:
@Feichtmeier @local-optimum @Jupi007 continuing to explore a solution to  [#887](https://github.com/ubuntu-flutter-community/software/issues/887) on how to avoid redundant information, could this solution work?

- removed unnecessary tooltips
- added a length threshold for Version and Publisher

### DIFF
--- a/lib/app/common/app_page/app_infos/app_infos.dart
+++ b/lib/app/common/app_page/app_infos/app_infos.dart
@@ -102,7 +102,6 @@ class RatingInfoFragment extends StatelessWidget {
 
     return AppInfoFragment(
       header: context.l10n.rating,
-      tooltipMessage: averageRating.toString(),
       child: Align(
         alignment: Alignment.center,
         child: Padding(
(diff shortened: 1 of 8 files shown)


Feichtmeier:
No we need these tooltips for when the information can't be read because the string is too long

madsrh:
> No we need these tooltips for when the information can't be read because the string is too long

I didn't remove all of them 🤔 you can always read the remaining strings. Eg. Date will never be longer than xx/xx/xxxx. I tried with the smallest possible window side.

Feichtmeier:
License can be super long

madsrh:
> License can be super long

Good point. There's even support for custom licenses.

This was the longest string I could find for testing 😁 

image

Feichtmeier:
grafik

works

madsrh:
> works

image

😮 I must have opened 100 different apps, looking for a license with a long name and you just happened to know this one?

Feichtmeier (approved):

Feichtmeier:
I knew this one :P  From testing this app through months I have some funny things to use for testing
Like intellij apps with channels

PR #1040, opened 2023-02-13
Package type visualization improvements

### LINKED ISSUE #1036: Package type visualisation improvements
Hey! How is it going?

@anasereijo  and I have been exploring for a few days different ideas on how to display the packages when we only have one. One of the options we've explored, and the one we've finally settled on for the proposal, is the idea that @Feichtmeier  proposed at the time. 

While it may also lead to think that it is clickable or not, we think it is the way that best maintains visual consistency between applications with only one package and applications with both. 

There are different design systems that account for this type of chips, where the left side corresponds to a "larger category" than the right side, which is usually clickable. So we think that, given the circumstances, this option is plausible and could work. 

As for the use of typography and colours, although we have basically followed Yaru's guide for this, we have tried to highlight a little one button over the other with the weight of the typography and the shade of grey used in each element. 

This is the proposal explored:

image

image

image

Thanks for everything! 😊

_Originally posted by @Zoospora in https://github.com/ubuntu-flutter-community/software/discussions/831#discussioncomment-4767982_

Feichtmeier:
Snap and Deb available:
grafik
grafik

Only snap but with multiple channels:
grafik

Only 1 channel and only snap:
grafik

Fixes #1036

### DIFF
--- a/lib/app/common/app_page/app_format_toggle_buttons.dart
+++ b/lib/app/common/app_page/app_format_toggle_buttons.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:software/app/common/app_format.dart';
 import 'package:software/l10n/l10n.dart';
 import 'package:yaru_icons/yaru_icons.dart';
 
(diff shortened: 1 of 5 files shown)


anasereijo:
@Feichtmeier what do we see when it's just a deb available ?

madsrh (approved):

Feichtmeier:
> @Feichtmeier what do we see when it's just a deb available ?

grafik

anasereijo:
I'd say let's go with this! We are doing some user interviews this week, so I am sure we will find out whether this works or not soon.

Feichtmeier:
oki

PR #1044, opened 2023-02-13
Bring section banner closer to design

Feichtmeier:
grafik
grafik
grafik

- smaller icons
- better alignment
- slogan + title ( we need better text here... )

ref:


Feichtmeier:
grafik
need to check tomorrow

madsrh (approved):

PR #1051, opened 2023-02-14
Correctly look for snap updates

### LINKED ISSUE #1042: CollectionPage: hide content snaps in updates
grafik

Feichtmeier:
It turns out that the issue was something totally different. I didn't check if the selected channel version is null.
Which I did now and then there is no version to compare to which means that there is no update available.
I first ran `snap refresh kde...." in the terminal and snapd said that there is no update available. After that I found the bug in the code

grafik

Fixes #1042

### DIFF
--- a/lib/app/common/snap/snap_utils.dart
+++ b/lib/app/common/snap/snap_utils.dart
@@ -12,7 +12,7 @@ bool isSnapUpdateAvailable({required Snap storeSnap, required Snap localSnap}) {
   );
   final trackingVersion = selectAbleChannels[tracking]?.version;
 
-  return trackingVersion != version;
+  return trackingVersion == null ? false : trackingVersion != version;
 }
 
 Map<String, SnapChannel> getSelectableChannels({required Snap? storeSnap}) {
--- a/lib/app/common/snap/snap_controls.dart
+++ b/lib/app/common/snap/snap_controls.dart
@@ -74,10 +74,10 @@ class SnapControls extends StatelessWidget {
                 ),
               if (model.snapIsInstalled)
                 ElevatedButton(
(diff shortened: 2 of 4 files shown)


lib/app/common/snap/snap_utils.dart
```diff
@@ -12,7 +12,7 @@ bool isSnapUpdateAvailable({required Snap storeSnap, required Snap localSnap}) {
   );
   final trackingVersion = selectAbleChannels[tracking]?.version;
 
-  return trackingVersion != version;
+  return trackingVersion == null ? false : trackingVersion != version;
```
d-loose, line 15: `trackingVersion != version` is enough here, since `version` can't be `null`

Feichtmeier: nope sadly not:
grafik

grafik

because the question is if the tracking channel is not null are the versions not equal, then there should be an update available

but if the tracking channel is null then there is no update available we could check, so we return false

Feichtmeier: because null is not equal to a not nullable string, so the function returns that there is an update available :)

d-loose: oh you are completely right - it's an inequality and I'm stupid :)

d-loose (approved):
LGTM :+1:

lib/app/common/snap/snap_utils.dart
Feichtmeier: no you are not!

np. this code is super weird. Also it is duplicated in snap_model but I fear to break anything so I did not replace it yet with snap-utils

PR #1052, opened 2023-02-14
Remove the package updating details

### LINKED ISSUE #1041: Remove the package updating details
IMHO this information is useless:

grafik

If we would be using apt directly someday then we could bring it back but who can have any plus in looking onto this weird packagekit output?

@madsrh @anasereijo ?

Feichtmeier:
grafik

Fixes #1041


d-loose (approved):
LGTM :+1:

PR #1056, opened 2023-02-14
SnapPage: add a share button

### LINKED ISSUE #1047: SnapPage: add a share button

Feichtmeier:
image

image

image

Fixes #1047

### DIFF
--- a/lib/app/common/app_page/app_header.dart
+++ b/lib/app/common/app_page/app_header.dart
@@ -18,6 +18,7 @@
 import 'package:flutter/material.dart';
 import 'package:software/app/common/app_data.dart';
 import 'package:software/app/common/app_page/publisher_name.dart';
+import 'package:yaru_icons/yaru_icons.dart';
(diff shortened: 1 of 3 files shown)


madsrh (approved):
Yes, share that snap! 😄 Some feedback you can consider:

Could you add a "share" tooltip? PR https://github.com/ubuntu-flutter-community/software/pull/1057

IMO the snackbar design is really very phonesentric and in a desktop app, the full width inverted surface feels massive.


I wonder if we could do something like this?

madsrh:
@Feichtmeier sorry for tagging you, I wasn't sure if you've missed this comment since the PR is closed, or if you just disagree 🐱 👆

Feichtmeier:
I like the color but I agree that we dont always need the full width

Actually M3 snackbars should look like this
grafik

I wonder if this is an issue in yaru.dart

madsrh:
> I like the color but I agree that we dont always need the full width

+1 we need that color to grab the users attention

PR #1065, opened 2023-02-15
Fix Title alignment

### LINKED ISSUE #895: Title alignment
I'm not sure what's best to have centered/aligned in relation to what but it would be nice to have a consistent alignment between pages. :)

image

Feichtmeier:
grafik

Fixes #895

### DIFF
--- a/lib/app/collection/collection_page.dart
+++ b/lib/app/collection/collection_page.dart
@@ -243,6 +243,7 @@ class _CollectionPageState extends State<CollectionPage> {
 
     return Scaffold(
       appBar: YaruWindowTitleBar(
+        leading: const SizedBox(width: kLeadingGap),
         title: SearchField(
           searchQuery: searchQuery ?? '',
           onChanged: setSearchQuery,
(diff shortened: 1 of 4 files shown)


lib/app/collection/collection_page.dart
```diff
@@ -243,6 +243,7 @@ class _CollectionPageState extends State<CollectionPage> {
 
     return Scaffold(
       appBar: YaruWindowTitleBar(
+        leading: const SizedBox(width: 40),
```
jpnurmi, line 246: what does 40 stand for, or where does it come from?

kYaruWindowControlSize is 24, YaruTitleBarTheme.buttonPadding defaults to 10, and YaruTitleBarTheme.buttonSpacing is 14.

if the value is chosen by the eye and cannot be calculated with these, perhaps it could be shared as a constant to avoid that it gets out of sync between the files?

Feichtmeier: it comes from the trailing  back  buttons having a 40 pixel width

jpnurmi: oh wait, 40 has nothing to do with the window controls, does it? is the meaning to match the size of the back button?

jpnurmi: it could be nice to express that in the code so that all three places use `kBackButtonSize` or something similar from `constants.dart` because it's not immediately obvious what that 40 is :)

jpnurmi (approved):

PR #1077, opened 2023-02-16
Find unlisted snaps by exact name

Feichtmeier:
CC @local-optimum 

grafik

The steam snap is unlisted. So now show it if you type in the exact name
grafik

But not if by any query
grafik

Though you should prbly set the steam snap to listed at some point :)

### DIFF
--- a/lib/app/explore/explore_model.dart
+++ b/lib/app/explore/explore_model.dart
@@ -134,6 +134,15 @@ class ExploreModel extends SafeChangeNotifier {
     }
   }
 
+  Future<Snap?> _findSnapByName(String name) async {
+    try {
+      return await _snapService.findSnapByName(name);
+    } on SnapdException catch (e) {
+      errorMessage = e.message.toString();
+      return null;
+    }
+  }
+
   Future<List<AppstreamComponent>> _findAppstreamComponents(
     String searchQuery,
   ) async =>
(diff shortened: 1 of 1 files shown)


d-loose (approved):
LGTM :+1:

local-optimum:
It's nearly ready to be listed ;)

kenvandine:
Actually, it is public but not in stable.  It's searchable via snapcraft.io, snap find, and gnome-software.

d-loose:
There's another parameter - `scope` - that isn't implemented in snapd.dart yet, and it does the trick:

```bash
$ curl -s --unix-socket /var/run/snapd.socket "http://localhost/v2/find?q=steam" | jq '.result[].name'
"electronplayer-portuguese"
"sl-moon127"
"sl-jiphex"
"sl-doismellburning"

$ curl -s --unix-socket /var/run/snapd.socket "http://localhost/v2/find?q=steam&scope=wide" | jq '.result[].name'
"steam"
"electronplayer-portuguese"
"sl-moon127"
"solus-runtime-gaming"
"sl-jiphex"
"m10e"
"sl-doismellburning"
"chad-stride"

```

d-loose:
Although, I can't seem to find 'workshops' or 'firmware-updater' using `scope=wide`

PR #1078, opened 2023-02-16
Make publisher name click search for publisher

### LINKED ISSUE #1016: Make publisher name click search for publisher
When clicking the publisher name, currently the users gets send to the app page. I would expect to be directed to the publishers page.

https://snapcraft.io/publisher/kde
_not_
https://snapcraft.io/kdenlive

I tried to fix this, but couldn't find the correct string in `app_data`? 🤷‍♂️ 

image

Feichtmeier:
https://user-images.githubusercontent.com/15329494/219559999-d48fc32a-0e03-4282-a5bb-300ad1bd0008.mp4

@d-loose aaaaaalmost :) 

1) make click on publisher search with the publisher username as query
2) don't do this when on the collectionpage

TODO:

- [ ] when searched from an apppage spawned from the search, the publisher search works but the searchquery is not correctly set. Any idea @d-loose ? :thinking: 

Fixes #1016

### DIFF
--- a/lib/app/common/app_page/app_page.dart
+++ b/lib/app/common/app_page/app_page.dart
@@ -201,8 +201,12 @@ class _AppPageState extends State<AppPage> {
     final onPublisherSearch =
         widget.enableSearch == false || !widget.initialized
             ? null
-            : () => Navigator.of(context)
-                .pop(searchByPublisher(widget.appData.publisherUsername));
+            : () async {
+                await searchByPublisher(widget.appData.publisherUsername);
+                if (context.mounted) {
+                  Navigator.of(context).pop();
+                }
+              };
 
     final normalWindowAppHeader = BorderContainer(
       initialized: widget.initialized,
--- a/lib/app/common/app_page/app_page.dart
+++ b/lib/app/common/app_page/app_page.dart
@@ -17,6 +17,7 @@
 
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
+import 'package:provider/provider.dart';
 import 'package:software/app/common/app_data.dart';
 import 'package:software/app/common/app_page/app_description.dart';
 import 'package:software/app/common/app_page/app_header.dart';
@@ -30,6 +31,7 @@ import 'package:software/app/common/border_container.dart';
 import 'package:software/app/common/custom_back_button.dart';
 import 'package:software/app/common/link.dart';
 import 'package:software/app/common/safe_network_image.dart';
+import 'package:software/app/explore/explore_model.dart';
 import 'package:software/l10n/l10n.dart';
 import 'package:yaru_icons/yaru_icons.dart';
 import 'package:yaru_widgets/yaru_widgets.dart';
@@ -55,6 +57,7 @@ class AppPage extends StatefulWidget {
     this.onVote,
     this.onFlag,
     this.initialized = false,
+    this.enableSearch = true,
   });
 
   final bool initialized;
@@ -64,6 +67,7 @@ class AppPage extends StatefulWidget {
   final Widget? controls;
   final Widget? subDescription;
   final bool appIsInstalled;
+  final bool enableSearch;
 
   final double? reviewRating;
   final String? review;
@@ -106,6 +110,9 @@ class _AppPageState extends State<AppPage> {
     final isWindowWide = windowWidth > 1200;
 
     final icon = widget.icon;
+
+    final searchByPublisher =
+        context.select((ExploreModel m) => m.searchByPublisher);
 
     final media = BorderContainer(
       initialized: widget.initialized,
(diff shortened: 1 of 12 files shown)


d-loose:
The `searchQuery` is set, but the `SearchField` on the `ExplorePage` isn't redrawn. The trick from [here](https://github.com/ubuntu-flutter-community/software/pull/914/files#diff-8a501980c8ad61f8a9da60054bd888a93dad5147e812f847704b3af08a89dfbe) should work.

Feichtmeier:
Nope :S 

But when one switches again to another explore page the search field is updated :S 


Uploading 2023-02-17 11-14-24.mp4…

d-loose:
> Nope :S

Should have tried it, I realize now that this doesn't work with the current structure anymore :/
The whole search field setup is still a bit messy, I haven't found a good solution yet

lib/app/explore/explore_page.dart
```diff
@@ -115,12 +115,14 @@ class _ExplorePageState extends State<ExplorePage> {
...
     final errorMessage = context.select((AppModel m) => m.errorMessage);
+    final isCurrent = ModalRoute.of(context)?.isCurrent ?? false;
 
     return Scaffold(
       appBar: YaruWindowTitleBar(
         leading: const SizedBox(width: kLeadingGap),
         title: SearchField(
-          key: ValueKey(showSearchPage),
+          autofocus: isCurrent,
+          key: ValueKey(showSearchPage != isCurrent),
```
d-loose, line 125: This seems to work:

```suggestion
          key: ValueKey(isCurrent),
```

d-loose: We also don't need the `autofocus` parameter here

Feichtmeier: sadly this doesn't help either. sometimes if you change the build method hot reload manages to make the impression that it works : D 
If you restart the app the effect is gone

Feichtmeier:
https://user-images.githubusercontent.com/15329494/219654751-a4e87da1-7726-4f78-bc7e-1e9c1325707c.mp4

I think I got it, please test @d-loose

lib/app/explore/explore_page.dart
```diff
@@ -120,7 +120,7 @@ class _ExplorePageState extends State<ExplorePage> {
       appBar: YaruWindowTitleBar(
         leading: const SizedBox(width: kLeadingGap),
         title: SearchField(
-          key: ValueKey(showSearchPage),
+          key: ObjectKey(ModalRoute.of(context)?.isCurrent ?? searchQuery),
```
d-loose, line 123: Very nice, this works well without causing any rebuild while the page is shown and the query changes

d-loose (approved):
LGTM :+1:

lib/app/explore/explore_page.dart
Feichtmeier: oh no... but the other thing that previously worked does not work anymore now :laughing: 
If you click a snap from the explore page without previously searching. The search still works, but now THIS searchquery stays empty :face_exhaling:

Feichtmeier: no this works. But the searchquery does not go away anymore if you clicked on the sidebar to navigate back (also known as "the designer shortcut")

d-loose: If only we had widget tests.. :laughing:

d-loose: Not sure if there's a better way to logically combine `showSearchPage` with the rest:
```suggestion
          key: ValueKey(
            '$showSearchPage${ModalRoute.of(context)?.isCurrent ?? searchQuery}',
          ),
```

Feichtmeier: Thanks, will test!

(Eventually we really need to come back to the tree restructuring. I think the instant search needs to go then want we need an autocomplete search field. )

Feichtmeier:
oh man... it worked for the first time and I was about to say "wooho it works". Then I refreshed the app aaaaand it's back to not changing the searchQuery if clicked on an apppage found by search :D

lib/app/common/app_page/app_page.dart
```diff
@@ -191,9 +198,16 @@ class _AppPageState extends State<AppPage> {
       Clipboard.setData(ClipboardData(text: appData.website));
     }
 
+    final onPublisherSearch =
+        widget.enableSearch == false || !widget.initialized
+            ? null
+            : () => Navigator.of(context)
+                .pop(searchByPublisher(widget.appData.publisherUsername));
```
d-loose, line 205: Now I realize that there's a race condition here - if the `searchQuery` isn't updated by `searchByPublisher` before the app page is popped and `ExplorePage` gets rebuild with `isCurrent=true`, it won't see the changed query and thus not rebuild the `SearchField`

Feichtmeier: so should I try to await the search before I pop the context?

d-loose, line 205: This adds a small delay, but otherwise WorksForMe™
```suggestion
    final onPublisherSearch =
        widget.enableSearch == false || !widget.initialized
            ? null
            : () async {
                await searchByPublisher(widget.appData.publisherUsername);
                if (context.mounted) {
                  Navigator.of(context).pop();
                }
              };
```

Feichtmeier:
okay now it really works

https://user-images.githubusercontent.com/15329494/219696469-cc8a523e-a4f4-4e3d-a1b8-e5c1f213c82b.mp4

PR #1089, opened 2023-02-17
Smaller fontsize for YaruExpanders

madsrh:
@anasereijo mentioned making the expander titles a bit smaller. It quickly became a mess, so @Jupi007 suggested creating a YaruExpandableTitle widget. 

Current master:
image

PR:
image

Updates:

### DIFF
--- a/lib/app/common/expandable_title.dart
+++ b/lib/app/common/expandable_title.dart
@@ -3,7 +3,12 @@ import 'package:flutter/material.dart';
 class YaruExpandableTitle extends StatelessWidget {
   final String title;
 
-  const YaruExpandableTitle(this.title);
+  final Key? titleKey;
+
+  const YaruExpandableTitle(
+    this.title, {
+    this.titleKey,
+  }) : super(key: titleKey);
 
   @override
   Widget build(BuildContext context) {
--- a/lib/app/common/expandable_title.dart
+++ b/lib/app/common/expandable_title.dart
@@ -1,0 +1,17 @@
+import 'package:flutter/material.dart';
+
+class YaruExpandableTitle extends StatelessWidget {
+  final String title;
+
+  const YaruExpandableTitle({required this.title});
+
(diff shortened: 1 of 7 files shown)


lib/app/common/expandable_title.dart
```diff
@@ -0,0 +1,22 @@
...
+
+class YaruExpandableTitle extends StatelessWidget {
+  final String title;
+
+  final Key? titleKey;
+
+  const YaruExpandableTitle(
+    this.title, {
+    this.titleKey,
+  }) : super(key: titleKey);
```
Jupi007, line 11: @madsrh here you just need to do:

```dart
const YaruExpandableTitle(
  this.title, {
  super.key
});
```

No need to create an intermediate `titleKey` property 🙂

Jupi007 (changes requested):

lib/app/common/expandable_title.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:flutter/material.dart';
+
+class YaruExpandableTitle extends StatelessWidget {
+  final String title;
+
+  const YaruExpandableTitle(this.title, {super.key});
```
Jupi007, line 6: Just had a thought, this will only be used on expandable containers, so what about renaming this widget into `ExpandableContainerTitle`? Because we have some other expandable who won't use this widget, and so I think a specific name would be better.

If you agree, you just need to press F2 after clicking on this (in VS code) to be able to rename all occurrences quickly 🙂

madsrh: As you can see, pressing F2 clearly didn't work for me 😆 Sorry for all the commits in the PR.

Jupi007 (approved):
LGFM 👍

But let's wait @Feichtmeier opinion before merging 🙂

Feichtmeier (approved):

PR #1094, opened 2023-02-18
Use primary color for links

madsrh:
Use primary color for links in apppage _description_ and update _changelog_

Before / After:

image

### DIFF
--- a/lib/app/collection/package_update_dialog.dart
+++ b/lib/app/collection/package_update_dialog.dart
@@ -108,6 +108,10 @@ class _PackageUpdateDialogState extends State<PackageUpdateDialog> {
               selectable: true,
               onTapLink: (text, href, title) =>
                   href != null ? launchUrl(Uri.parse(href)) : null,
+              styleSheet: MarkdownStyleSheet(
+                p: Theme.of(context).textTheme.bodyMedium,
+                a: TextStyle(color: Theme.of(context).primaryColor),
+              ),
             ),
           ),
         ),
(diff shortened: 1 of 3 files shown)


Feichtmeier:
generally yes, but ask the Anas because I saw links to be always blue in their mock-ups

madsrh:
> generally yes, but ask the Anas because I saw links to be always blue in their mock-ups

Sorry for not creating an issue, I figured this was easy peasy, but yeah let's check with @anasereijo 

 We can try with all blue as long as we just don't have two link colors like below I'm happy:

image

anasereijo:
Heyy,

> We can try with all blue as long as we just don't have two link colors

yup, I totally agree – it should be consistent. I've used the blue link on the wireframes, as it is what we use for links on Vanilla. However, orange is the default accent colour, so we should keep the orange as default. Apologies for the confusion!

I imagine that if the user selects another accent colour that it will change accordingly, right?

cheers

Feichtmeier (approved):

madsrh:
Well, there's still these links 🤷‍♂️ Should I include this in the PR as well. Do you have an opinion here @Feichtmeier ?


Current:


Orange:


Just plain text:

Feichtmeier:
no idea :) go for whatever you want the links to look like

madsrh:
> no idea :) go for whatever you want the links to look like

Damn 😸 Well as the title says, this PR is "Use primary color for links" so if we want blue, we can always revert these changes in another PR 👍

PR #1098, opened 2023-02-20
AlertDialog for removing snap

madsrh:
Fixes https://github.com/ubuntu-flutter-community/software/issues/988

I was unsure about the wording (not provided in the issue), so I've simply used already existing string.

image

### DIFF
--- a/lib/app/collection/package_update_dialog.dart
+++ b/lib/app/collection/package_update_dialog.dart
@@ -108,6 +108,10 @@ class _PackageUpdateDialogState extends State<PackageUpdateDialog> {
               selectable: true,
               onTapLink: (text, href, title) =>
                   href != null ? launchUrl(Uri.parse(href)) : null,
+              styleSheet: MarkdownStyleSheet(
+                p: Theme.of(context).textTheme.bodyMedium,
+                a: TextStyle(color: Theme.of(context).primaryColor),
+              ),
             ),
           ),
         ),
--- a/lib/app/common/app_page/app_description.dart
+++ b/lib/app/common/app_page/app_description.dart
@@ -46,6 +46,7 @@ class AppDescription extends StatelessWidget {
               href != null ? launchUrl(Uri.parse(href)) : null,
           styleSheet: MarkdownStyleSheet(
             p: Theme.of(context).textTheme.bodyMedium,
+            a: TextStyle(color: Theme.of(context).primaryColor),
           ),
         ),
       ),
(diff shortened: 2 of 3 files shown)


madsrh:
Yes! I've set a record in most failed tests 💪 🏆 

Sorry, I have no idea why. It works fine here and I've checked with `flutter analyze`

d-loose:
> Sorry, I have no idea why. It works fine here and I've checked with `flutter analyze`

It's my fault - sorry for the disturbance! See #1097

lib/app/collection/simple_snap_controls.dart
```diff
@@ -110,8 +110,36 @@ class SimpleSnapControls extends StatelessWidget {
...
+                          context: context,
+                          builder: (BuildContext context) {
+                            return AlertDialog(
+                              title: Text(context.l10n.confirm),
+                              content: Text(
+                                '${context.l10n.remove} ${model.snap.apps.first.name}?',
+                              ),
+                              actions: <Widget>[
+                                OutlinedButton(
+                                  child: const Text('Cancel'),
```
d-loose, line 125: Can we use the localization string here?

lib/app/collection/simple_snap_controls.dart
```diff
@@ -110,8 +110,36 @@ class SimpleSnapControls extends StatelessWidget {
...
+                                  onPressed: () {
+                                    Navigator.of(context).pop();
+                                  },
+                                ),
+                                ElevatedButton(
+                                  child: const Text('Remove'),
```
d-loose, line 131: Another localization string, please :)

lib/app/collection/simple_snap_controls.dart
```diff
@@ -110,8 +110,36 @@ class SimpleSnapControls extends StatelessWidget {
...
-                    enabled ? () => model.remove(context.l10n.done) : null,
+                onPressed: enabled
+                    ? () {
+                        showDialog(
+                          context: context,
+                          builder: (BuildContext context) {
+                            return AlertDialog(
+                              title: Text(context.l10n.confirm),
+                              content: Text(
+                                '${context.l10n.remove} ${model.snap.apps.first.name}?',
```
d-loose, line 121: I think it would be better to have a dedicated localization string for package removal with the app name as a parameter, because there're probably languages where "\<word for remove\> *app name*?" isn't correct.

d-loose (changes requested):

lib/app/collection/simple_snap_controls.dart
d-loose: I've opened a PR on your PR :)

d-loose (approved):
LGTM :+1:

Feichtmeier (approved):

anasereijo:
title: Remove "foobar2000"
text: Are you sure you want to remove this package?
buttons: they look nice, but I think Remove could be red (we do that in other products)

madsrh:
Yes! 

Now, with the help from Dennis, I can fix that. I'll adjust the PR

PR #1105, opened 2023-02-21
Bulk refresh

### LINKED ISSUE #1074: collection page: cancelling updates one by one
I just noticed that when I click on [update (4)] I get asked to insert my password, but if I cancel, I have to **cancel four times** - maybe because it's 4 updates? 

(not so good recording, but I can provide a better one if needed :D)

[Screencast from 16-02-23 11_49_53.webm](https://user-images.githubusercontent.com/36884067/219358670-11be87ff-57d8-4904-becc-c44b686f87c8.webm)

d-loose:
This makes use of `refreshMany` in snapd.dart to bulk refresh multiple snaps at once.
Note that it's necessary to change how snap changes are tracked in `SnapService`: a single `SnapChange` can affect multiple snaps, so using a `Map<Snap, SnapChange>` doesn't make sense any more. This also has an effect on the notifications - they now use 'Software Store' as `appName` rather than the name of the snap that has been changed, since there can be multiple now.

It would be great to test this with real-world snap updates before merging, even though I've added a unit test.

Fix #1074

### DIFF
--- a/lib/app/app_model.dart
+++ b/lib/app/app_model.dart
@@ -34,7 +34,7 @@ class AppModel extends SafeChangeNotifier {
   );
 
   final SnapService _snapService;
-  Map<Snap, SnapdChange> get snapChanges => _snapService.snapChanges;
+  Map<String, SnapdChange> get snapChanges => _snapService.snapChanges;
   StreamSubscription<bool>? _snapChangesSub;
 
   final AppstreamService _appstreamService;
--- a/lib/app/collection/collection_model.dart
+++ b/lib/app/collection/collection_model.dart
@@ -129,29 +129,13 @@ class CollectionModel extends SafeChangeNotifier {
   Future<void> refreshAllSnapsWithUpdates({
     required String doneMessage,
   }) async {
-    await _snapService.authorize();
     if (snapsWithUpdate.isEmpty) return;
-
-    final firstSnap = snapsWithUpdate.first;
-    _snapService
-        .refresh(
-      snap: firstSnap,
+    await _snapService.authorize();
+    await _snapService.refreshMany(
+      snaps: snapsWithUpdate,
       message: doneMessage,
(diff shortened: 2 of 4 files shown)


Feichtmeier:
will test when I got some snap updates and time ⏲️ 👍

d-loose:
> Do the individual tasks provide the snap that is being updated?

Yes, `snapChange.snapNames` contains the list of names of all snaps affected by this change.

I use this to get the `snapChange` for a given snap [here](https://github.com/ubuntu-flutter-community/software/pull/1105/files#diff-0fe053dbe94dc1c678349e896e0fd1b9dce889b66fe4d1672c4e093bf9345622R76-R77) (assuming a snap can only be changed by a single task at a time).

Feichtmeier:
Hmm but is it possible to track the progress for each individual snap here then? otherwise it looks a bit weird to show those X spinners with the same process :)

d-loose:
Hm, good question. While `SnapdChange` has a list of affected snaps (although not documented [here](https://snapcraft.io/docs/snapd-api#heading--changes), but implemented [here](https://github.com/snapcore/snapd/blob/9cf75d31bf4eb5d9ccd1f0c82149ef7819c8b117/daemon/api.go#L171)) the individual `Task`s don't explicitly provide the name of the individual snap this task is working on (I'm not even sure the change is split into individual tasks for each snap).
I can look into it, but for the moment we only get the progress for the entire `SnapdChange` that refreshes all snaps.

d-loose:
I don't think this is possible with the information snapd currently provides. Maybe we could show the progress of the update in a similar fashion as we do for debs?

Feichtmeier:
That's a shame :( Now I feel bad for this idea
I think it is very natural to see the update in each snap row. It is a common pattern around other store apps and things where you download a list of things

d-loose:
Superseded by #1130

PR #1108, opened 2023-02-22
Update from yaru

Feichtmeier:
A request from the installer team led to stronger borders and to popupmenu buttons (dropdowns) having the same border as outlined buttons

| Before | After |
| - | - |
|grafik|grafik|
|grafik|grafik|

Updating yaru and yaru_widgets here will thus also change those styles here.

Is this okay for you @anasereijo @Zoospora @madsrh ?

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -55,10 +55,10 @@ dependencies:
   url_launcher: ^6.1.2
   version: ^3.0.2
   xdg_icons: ^0.0.1
-  yaru: ^0.5.1
-  yaru_colors: ^0.1.3
+  yaru: ^0.5.4
+  yaru_colors: ^0.1.6
   yaru_icons: ^1.0.3
-  yaru_widgets: ^2.1.0
+  yaru_widgets: ^2.1.1
 
 dev_dependencies:
   build_runner: ^2.2.0
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -55,7 +55,7 @@ dependencies:
   url_launcher: ^6.1.2
   version: ^3.0.2
   xdg_icons: ^0.0.1
-  yaru: ^0.5.4
+  yaru: ^0.5.5
   yaru_colors: ^0.1.6
   yaru_icons: ^1.0.3
   yaru_widgets: ^2.1.1


Feichtmeier:
for reference https://github.com/ubuntu/yaru.dart/issues/244

madsrh:
Sorry, but I really think these are way too strong. I hope we can find some other solution.

PR: 
image

Current: image

Feichtmeier:
Could you eventually recommend a hex for dividers and outline on the yaru repo? Buttons, toggle buttons and the "dropdown" use outline. Dividers, yarubanners, yarusections and yarubordercontainer user divider colour

madsrh:
> Could you eventually recommend a hex for dividers and outline on the yaru repo? Buttons, toggle buttons and the "dropdown" use outline. Dividers, yarubanners, yarusections and yarubordercontainer user divider colour

That would be tough because I'm very happy with the current 🤔 Any ideas @Jupi007 ?

Feichtmeier:
Well then we need to talk to JP and elio about this I guess. Maybe we can get away with only increasing the border of those filled buttons I changed for them

Feichtmeier:
Updated the screenshots to use yaru 0.5.5 with again toned down borders. A bit stronger than in main but still not as extreme

madsrh:
> Updated the screenshots to use yaru 0.5.5 with again toned down borders. A bit stronger than in main but still not as extreme

I cannot see any change between the two screenshot 😄 Did you upload the same twice?

I will approve this one, but would prefer to keep the current

Jupi007:
> That would be tough because I'm very happy with the current thinking Any ideas @Jupi007 ?

I don't want to take a strong position on this very complex topic, because this need to be think globally, but maybe something intermediate between the current look (which can be a bit erased in some situations), and the initial proposal (which is really too much from my point of view) could fit? 🙂

> I cannot see any change between the two screenshot smile Did you upload the same twice?

I also can't notice any diff 🤔

Feichtmeier:
| Before | After |
| - | - |
|grafik|grafik|
|grafik|grafik|

Feichtmeier:
You were right, those were the same screenshots. Take a look again now =)

Jupi007:
I tested it and honestly, it is not that bad :)

madsrh:
I think my opinion is the same. We can merge this, but I don't think this is an improvement. Especially this update page is very border heavy and looks really too much when the borders gets brighter.

> I don't want to take a strong position on this very complex topic, because this need to be think globally, but maybe something intermediate between the current look (which can be a bit erased in some situations), and the initial proposal (which is really too much from my point of view) could fit? 🙂

Then you feel this would fit? But yeah, it is very complex 😵

Feichtmeier:
My problem is not that I like it but that other people requested this who seem to be responsible for the installer looks.

Okay full stop right now. :skull: 

I can not fight the fights between the installer team and the software team 


I am not sure if everyone here

1) understands the dependencies of the repositories we created over the past months and what part they play in the visuals we created
2) accepts the fact that there will never be a theme that fits for every situation and there will *always* be situations where we need to change things per-app

Would be awesome if you could talk to each other @jpnurmi @anasereijo @madsrh regarding those two points

Thanks!

elioqoshi:
I am repeating this for the xth time now considering you did not fail to repeat yourself @Feichtmeier.

If you do not agree with the software store direction, there is no immediate need to impact the installer decisions on the software direction. I did not do any executive decision on how the design should be, so please do not frame it as something we are forced to do. I am happy to continue discussing how the borders should be designed if we want to streamline it in cross-applications. 

> I am not sure if everyone here

> Would be awesome if you could talk to each other @jpnurmi @anasereijo @madsrh regarding those two points

I am saying this in the nicest way possible but please do change this patronizing tone in your comments as it's out of place in a professional environment where we are working towards the same goal. Just because we disagree on things does not mean we are not well-informed or don't know what we are doing. 

Disagreeing and debating on these directions is a natural part of the design process and there is no need to get frustrated, let's please be more patient with each-other.

Feichtmeier:
What patronising tone? I expressed the problems several times and it looked to me that the signal didn't find the receiver. 

I'm not sure what your goal is but the yaru team was very successful in creating a completely new UI set and style which we've successfully baked into different depositories. Plus I've successfully created this application with the help of Mads, Paul and canonical people.

On topic:

Since the border changes are rather tame now we can merge it and see if it grows upon us

elioqoshi:
> What patronising tone? I expressed the problems several times and it looked to me that the signal didn't find the receiver.

That's exactly how I feel. You are choosing to ignore what I am saying as well.

Again, I do not feel strongly about the borders, if you are not comfortable to merge this, don't merge it and we can find an alternative solution.

Feichtmeier:
That is from the previous version.
With the version in this PR it should look like this
grafik

Feichtmeier:
but the border is #444444 while the fill is #414141, need to check while they are not the same yaru.dart, independent of the color we pick

Feichtmeier:
it is the correct color (as it comes from 1 variable) in yaru
grafik

however we use the wrong color here now since divider and outline are currrently different in yaru

will fix it here independent of what values the two border vars will have

```dart
const snapLabel = SizedBox(
      height: 39,
      child: AppFormatLabel(appFormat: AppFormat.snap),
    );

    final snapLabelContainerCut = YaruBorderContainer(
      color: theme.dividerColor, <---------------------------- needs to be outline
      padding: const EdgeInsets.symmetric(horizontal: 5),
      borderRadius: const BorderRadius.only(
        topLeft: Radius.circular(kYaruButtonRadius),
        bottomLeft: Radius.circular(kYaruButtonRadius),
      ),
      child: snapLabel,
    );

    final snapLabelWithChannelButton = Row(
      mainAxisSize: MainAxisSize.min,
      children: [
        snapLabelContainerCut,
        OutlinedButtonTheme(
          data: OutlinedButtonThemeData(
            style: OutlinedButtonTheme.of(context).style?.copyWith(
                  shape: MaterialStateProperty.resolveWith(
                    (states) => const RoundedRectangleBorder(
                      borderRadius: BorderRadius.only(
                        topRight: Radius.circular(kYaruButtonRadius),
                        bottomRight: Radius.circular(kYaruButtonRadius),
                      ),
                    ),
                  ),
                ),
          ),
          child: const SnapChannelPopupButton(),
        )
      ],
    );
    ```

Jupi007:
Well... I would say, let's give a try to these softly brighter borders... but honestly I also prefer the current look, so 😅
I think the best decision would still to try this a little while, so can be sure if we like it (or really not).

PR #1113, opened 2023-02-23
Bump dependency on intl

robert-ancell:
```$ flutter pub get
Resolving dependencies... (4.6s)
Because software depends on flutter_localizations from sdk which depends on intl 0.18.0, intl 0.18.0 is required.
So, because software depends on intl ^0.17.0, version solving failed.
```


robert-ancell:
Ah, this is only an issue on flutter master.

PR #1117, opened 2023-02-23
SnapPage: use outline for snapchannel button

Feichtmeier:
grafik


madsrh (approved):

madsrh:
Check this out @Feichtmeier 

No fractional scaling (aka 100%)
image

Fractional scaling at 175%
image

Feichtmeier:
urgs that's weird. how does 200% look? might be the float vs int grid thing all over again so we need a repaint boundary

PR #1138, opened 2023-03-01
Dependency removal

### LINKED ISSUE #980: Dependency removal

d-loose:
Lists dependencies of installed packages on the app page and shows a dialog that provides the user with the option to remove unneeded dependencies, when removing a package.

[Screencast from 2023-03-01 13-45-29.webm](https://user-images.githubusercontent.com/113362648/222143285-316dc134-384a-42f0-b610-6dfd6a6aa372.webm)

Fix #980

### DIFF
--- a/lib/app/common/packagekit/dependency_dialogs.dart
+++ b/lib/app/common/packagekit/dependency_dialogs.dart
@@ -1,0 +1,198 @@
+import 'package:collection/collection.dart';
+import 'package:data_size/data_size.dart';
+import 'package:flutter/material.dart';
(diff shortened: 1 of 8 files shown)


Feichtmeier (approved):
tested and works! TY!

PR #1140, opened 2023-03-02
SectionBanner: fill with appfindings rather than snaps

### LINKED ISSUE #1118: SectionBanner: fill with appfindings rather than snaps
When clicking on for example gimp from a sectionbanner only the snap app format is found. This is because we only have a 3-list of snaps in there and not appfindings

d-loose:
Fix #1118


Feichtmeier (approved):
Awesome, thanks!

PR #1146, opened 2023-03-08
UpdatePage: switch appformat dropdown to toggle button

### LINKED ISSUE #1080: UpdatePage: switch appformat dropdown to toggle button
As mentioned in the updates page thread but breaking out here for easier tracking. It has been identified through user surveys (and personal experience) that the drop down package select makes it difficult for users to find the debian packages or see debian updates easily without fiddling with the drop downs.

We already have an excellent solution on the app pages themselves:
image

I propose that we change ALL snap/deb drop downs to use the same toggle approach as the above, as seen on:
- The search page
- The collection/manage page

And any future locations this might arise.

Feichtmeier:
Fixes #1080


d-loose (approved):
LGTM :+1:

PR #1149, opened 2023-03-10
ExplorePage: detangle and prepare for more pages

Feichtmeier:
- split startpage into GenericStartPage, ExploreAllPage and GamesStartPage
- remove parameters of GamesStartPage

### DIFF
--- a/lib/app/explore/start_page.dart
+++ b/lib/app/explore/start_page.dart
@@ -16,32 +16,34 @@
  */
 
 import 'package:flutter/material.dart';
+import 'package:provider/provider.dart';
 import 'package:shimmer/shimmer.dart';
 import 'package:software/app/common/app_banner.dart';
 import 'package:software/app/common/app_finding.dart';
 import 'package:software/app/common/constants.dart';
 import 'package:software/app/common/loading_banner_grid.dart';
 import 'package:software/app/common/snap/snap_section.dart';
+import 'package:software/app/explore/explore_model.dart';
 import 'package:software/app/explore/section_banner.dart';
 import 'package:software/app/explore/section_grid.dart';
 import 'package:software/snapx.dart';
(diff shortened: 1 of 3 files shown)


lib/app/explore/start_page.dart
```diff
@@ -63,32 +65,83 @@ class _StartPageState extends State<StartPage> {
...
+        padding: const EdgeInsets.only(top: 15),
+        child: Column(
+          children: const [
+            _LoadingSectionBanner(),
+            LoadingBannerGrid(),
+          ],
+        ),
+      );
+    }
+
```
d-loose, line 90: could we move some of this structure into a generic page skeleton, so we don't have to duplicate it for every page?

lib/app/explore/start_page.dart
```diff
@@ -101,45 +154,128 @@ class _TeaserPage extends StatelessWidget {
...
+      bannerApp3 = appsWithIcons.elementAt(2);
+    }
+    if (bannerApp == null || bannerApp2 == null || bannerApp3 == null) {
+      return SingleChildScrollView(
+        padding: const EdgeInsets.only(top: 15),
+        child: Column(
+          children: const [
+            _LoadingSectionBanner(),
+            LoadingBannerGrid(),
+          ],
```
d-loose, line 239: same as above

lib/app/explore/start_page.dart
Feichtmeier: We can do this for the generic page, but I don't know what the plans are for all pages.
I try to extract the loading thing though

Feichtmeier:
@d-loose Ok I made a different approach: apps are nullable now (both the list and the children) and the seciton banner and the grid decide on their own what to show (null ? -> loading..) - WDYT?

d-loose (approved):
LGTM :+1:

PR #1151, opened 2023-03-11
More readable publisherName for app_banners

madsrh:
Just a suggestion: a very small change, but I think this is easier to read. WDYT? @anasereijo @Jupi007 @Feichtmeier 

- removed _italic_
- used Hintcolor

I wonder if we should remove the italic from the app page also?

### Before:
image
image

### After:
image
image


anasereijo:
👍👍 I think it’s more readable and the colour is probably more accessible too (didn’t test it though)

Jupi007:
I like it 👍
And yeah, it would be better to also drop the italic font on the app page for consistency.

madsrh:
Removed italic from AppPage

image

Feichtmeier (approved):

PR #1152, opened 2023-03-11
Appformat popup font size

madsrh:
The two dropdowns or popups should have the same font size:

### Before:
image

### After:
image


Jupi007:
Much better like this 👍

Feichtmeier (approved):
cool!
I wonder if we could change this in yaru_widgets, too (after this)

PR #1155, opened 2023-03-12
Changes searchfield font size

madsrh:
Okay, please check this code - It took me all day and I must have visited every site on Google looking for a way to fix this 😅

I also made the clear icon a bit stronger. WDYT?

Ikke-navngivet-2

### Before
image

image

### After
image

image

### DIFF
--- a/lib/app/common/search_field.dart
+++ b/lib/app/common/search_field.dart
@@ -89,7 +89,7 @@ class _SearchFieldState extends State<SearchField> {
         child: Center(
           child: SizedBox(
             width: 280,
-            height: 30,
+            height: 34,
             child: TextField(
               style: theme.textTheme.bodyMedium,
               strutStyle: const StrutStyle(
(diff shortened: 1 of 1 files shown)


Jupi007 (approved):
Great, it looks much better!

Feichtmeier:
The font looks better but I think the text field height changes

madsrh:
> but I think the text field height changes

This is smaller, yes. Do you prefer the current height even with the smaller text @Feichtmeier ?

Feichtmeier:
i am not sure, I found it always very hard to find the correct size here. If we find a solution I would love some generic headerbar textfield sizes for all our apps : D
Would it be possible to have the font change without the size change?

madsrh:
> Would it be possible to have the font change without the size change?

Yes, I changed it deliberate. Like this?

image

Feichtmeier:
yeah maybe you are right. I think the textfield should be as tall as buttons, 40 pixels, just curious how does a button look alongside of this textfield?

madsrh:
image

Feichtmeier:
Let's give it a try and see if this sticks

Feichtmeier (approved):

PR #1159, opened 2023-03-14
Additional Information fixes

Feichtmeier:
- use horizontal 20 pixel spacing
- move the icon to the start
- publisherName.replaceAll(' ', '\u00A0')

[Bildschirmaufzeichnung vom 14.03.2023, 11:23:30.webm](https://user-images.githubusercontent.com/15329494/224971306-5b91aa1c-d30e-452e-9b8f-c2cbd63e27db.webm)

CC @anasereijo

### DIFF
--- a/lib/app/common/app_page/app_infos/additional_information.dart
+++ b/lib/app/common/app_page/app_infos/additional_information.dart
@@ -34,6 +34,7 @@ class AdditionalInformation extends StatelessWidget {
           gridDelegate: const SliverGridDelegateWithMaxCrossAxisExtent(
             maxCrossAxisExtent: 200,
             mainAxisExtent: 100,
+            crossAxisSpacing: kYaruPagePadding,
           ),
           children: [
             PublisherInfoFragment(
--- a/lib/app/common/app_page/app_infos/publisher_info_fragment.dart
+++ b/lib/app/common/app_page/app_infos/publisher_info_fragment.dart
@@ -28,7 +28,6 @@ class PublisherInfoFragment extends StatelessWidget {
     required this.publisherName,
     this.starDev = false,
     required this.website,
-    this.limitChildWidth = true,
     this.height = 14,
     this.enhanceChildText = false,
   });
@@ -37,7 +36,6 @@ class PublisherInfoFragment extends StatelessWidget {
   final bool starDev;
   final String publisherName;
   final String website;
-  final bool limitChildWidth;
   final double height;
   final bool enhanceChildText;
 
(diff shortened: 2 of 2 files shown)


madsrh:
Es funktiniert 🎉 

Should we also move the icons in the other places? 🙈 

You've properly set this review to auto-merge, so I'll wait for @anasereijo before approving 😄

anasereijo:
Let me bring this up with the snapcraft.io team, just to check if there is some reason why we shouldn't be doing this.
I've found the same issue on the website, so maybe this is a good opportunity to fix it on both sides!

Screenshot 2023-03-14 at 15 07 25

anasereijo:
Hello! I double-checked with some of my design colleagues if there is any guidance or reasons why we shouldn't do this, and it seems that it's okay. 👍🏻 I will also file an issue to change it on snapcraft.io for consistency.
> Should we also move the icons in the other places?

yes @madsrh is right - we have to change it everywhere now

One thing we talked about was the fact that, when too long, we truncate the publisher name – it would be better to wrap it instead, as it's too important information to be hiding. I can create a separate issue if needed.

anasereijo:
@Feichtmeier please do not proceed with this just yet 😅 there are some more ideas/concerns coming from internally

madsrh:
> a quick fix for this ...

That's fixes that issue, so not a quick fix... just a fix 😺 

That would be fine. Like this?

image

anasereijo:
> That's fixes that issue, so not a quick fix... just a fix 😺

😅 yeah, I mean quick as with this we don’t have to change the icon it everywhere, right?

Feichtmeier:
@anasereijo isn't this a bit uneven looking ?
grafik

How about ellipses if too long, icon to the right, but then have the tooltip on hover?
grafik

madsrh:
> How about ellipses if too long, icon to the right, but then have the tooltip on hover?

Isn't this how it works in the current master? 🤔 

> isn't this a bit uneven looking ?

Yeah, that is a bit wobbly 😅  Sorry, just a quick mockup. I was focused on the placement of the icon. You're right, it should ofc be:

image

Feichtmeier:
Sadly I didn't manage to stop the uneven height. This is the best I could do 
grafik

madsrh:
@Feichtmeier what about putting each fragment in a container and adding Alignment.topleft?

anasereijo:
@Feichtmeier oh that's a shame. @madsrh's [mockup](https://github.com/ubuntu-flutter-community/software/pull/1159#issuecomment-1474809149) is exactly what I had in mind.

Feichtmeier:
@anasereijo @madsrh I know this is not finished. But here is also the fix for breaking the publisher name witch every space. Even if it is not 100% like your mock-up it is still a small improvement over the current. Thus I am once again asking for merge :smile: 
(this PR will not auto-close any issue)

madsrh:
Sure 😊

madsrh (approved):

Feichtmeier:
the latest flutter show has a solution for this always top aligned but varying height problem with hose 2 OR 3 row fragments:
https://gist.github.com/craiglabenz/c6fc52e3e61f66c51f7a858115bfce51
https://www.youtube.com/watch?v=cq34RWXegM8

madsrh:
> the latest flutter show has a solution for this always top aligned but varying height problem with hose 2 OR 3 row fragments

Cool 😎
That video started out so newbie friendly but then... 🤯

PR #1160, opened 2023-03-15
Remove themeTile artifacts

whiskeyPeak:
You can see the white container bleed through due to anti-aliasing in dark mode.

Before:

Screenshot from 2023-03-15 05-35-07

After:

Screenshot from 2023-03-15 05-34-40


Feichtmeier (approved):
Ty 😊 there's also some new clipper in yaruwidgets but this can be done somewhen else

PR #1161, opened 2023-03-15
Make collection page responsive

### LINKED ISSUE #1128: Max width for Collections/Manage page
Current:
image

Expected:
image

Similar to:
image

whiskeyPeak:
Makes the collection page responsive when the window width >= 1200.

After:

https://user-images.githubusercontent.com/73116038/225227148-4ac53941-32db-43ea-871b-f335be65dc33.mp4

Fixes #1128

### DIFF
--- a/lib/app/collection/collection_page.dart
+++ b/lib/app/collection/collection_page.dart
@@ -235,8 +235,8 @@ class _CollectionPageState extends State<CollectionPage> {
                     child: (appFormat == AppFormat.snap)
                         ? const SnapCollection()
                         : PackageCollection(
-                      enabled: !checkingForPackageUpdates,
-                    ),
+                            enabled: !checkingForPackageUpdates,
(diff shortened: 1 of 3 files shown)


Feichtmeier:
Thanks @whiskeyPeak 
I had other plans for this page though. I thought maybe make use of expanded with different flex values and show more snap properties when the window is wider. Can't show anything but maybe you could imagine what I mean

whiskeyPeak:
> Thanks @whiskeyPeak I had other plans for this page though. I thought maybe make use of expanded with different flex values and show more snap properties when the window is wider. Can't show anything but maybe you could imagine what I mean

Ahhh I see what you mean, kinda like the explore page? 
We don't need variable padding for that.

madsrh:
> ...show more snap properties when the window is wider

That's a great idea 💡

PR #1163, opened 2023-03-16
Use same dateformat for ReleasedAt and InstalledAt

madsrh:
Here's another "death by a thousand paper cuts" fix 😅 Use the same date format for installed and released at date.

I would like to do something (UK/Eur) like `DateFormat('d. MMMM, y')` but that format isn't predefined, so I'm worried that localization won't work 🤷‍♂️

**Before**
image

**After**
image


Feichtmeier (approved):
Thanks 🙏 📄✂️👲🐉㊙️⛩️🥟

PR #1167, opened 2023-03-19
fix(UI): explore page snaps have lower case letters

### LINKED ISSUE #1156: Explorepage: Snaps have lowercase titles
Titles for snap packages on explore page is all lowercase, but on app page it's capitalized.
Is this because the apppage uses `appdata.title`, but the app_banner uses `appstream` and can we fix it?

image

image

image

shan-shaji:
Closes #1156

### DIFF
--- a/lib/app/explore/section_grid.dart
+++ b/lib/app/explore/section_grid.dart
@@ -66,7 +66,7 @@ class SectionGrid extends StatelessWidget {
 
         return AppBanner(
           appFinding: MapEntry<String, AppFinding>(
-            app!.snap!.name,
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Hi @shan-shaji thanks!

You only need to fill out this agreement there once and after that launchpad needs 1 day to verify your mail then the test should pass
once you did this you don't need to do this again

madsrh:
Works 🎉 Thank you @shan-shaji 😃 

image

Feichtmeier (approved):

Issue #1173, opened 2023-03-22
reviews: make the divider less adipose 


Feichtmeier:
![grafik](https://user-images.githubusercontent.com/15329494/226900130-a1c7ca8f-18fa-484c-b9cd-b20f45c7991a.png)

it is very fat

it is easy to fix (height + thickness 0 and wrap in padding)

PR #1180, opened 2023-03-22
Issue1173 - review divider

madsrh:
I couldn't make it work like you mentioned in #1173 @Feichtmeier but here's a solution. Looks better in the app than this screenshot.

Ikke-navngivet-1

### DIFF
--- a/lib/app/common/app_page/app_reviews.dart
+++ b/lib/app/common/app_page/app_reviews.dart
@@ -559,7 +559,7 @@ class _Review extends StatelessWidget {
         ),
         Container(
           height: 1,
-          margin: const EdgeInsets.symmetric(vertical: 30),
+          margin: const EdgeInsets.symmetric(vertical: 20),
           decoration: BoxDecoration(
             color: Theme.of(context).dividerColor,
             borderRadius: BorderRadius.circular(1),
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Hm no there is a strange trick with dividers

     Padding( 
       padding: EdgeInsets.only( 
         left: 20, 
         right: 20, 
         top: 20, 
         bottom: 20, 
       ), 
       child: const Divider( 
         height: 0, 
         thickness: 0,
       ), 
     );

(The values are probably wrong,I just copied this from the music app on my phone)

If this doesn't work try to make one of the 20s uneven, like 19 or something

It's related to this sometimes blurry rendering problem we've encountered before somewhere else

Feichtmeier:
Woops okay sorry. The trick is padding, 0 height, and don't add thickness :facepalm: 

grafik
grafik

2x
const Padding(
                padding: EdgeInsets.only(top: 30, bottom: 30),
                child: Divider(
                  height: 0,
                ),
              )

and

1x
const Padding(
          padding: EdgeInsets.only(top: 20, bottom: 20),
          child: Divider(
            height: 0,
          ),
        )

Feichtmeier:
Thanks for investing the time!

Feichtmeier (approved):

PR #1205, opened 2023-03-31
feat(UI): allow only full star rating

### LINKED ISSUE #1184: star ratings: allow full stars instead of half stars
We currently allow users to add half stars as ratings. I checked with a few other designers, and we agree that this is quite subjective for rating apps, so not very useful. Allowing just full star ratings should be enough. We will then have the average that can be decimal. It also seems to be pretty standard across other platforms. 

Mar-23-2023 11-06-49

shan-shaji:
Closes #1184


Feichtmeier (approved):
Ty

shan-shaji:
Hi @Feichtmeier, Can you please help me with the test? I couldn't find the reason for the test failure.

Feichtmeier:
> Hi @Feichtmeier, Can you please help me with the test? I couldn't find the reason for the test failure. 

Don't worry nothing you committed had an affect on the integration test
Those issues happen from time to time if the connection between the GitHub action container and the snap servers is slow
I'll retrigger the test until we get our green ✅, even though we could merge as is

madsrh:
It looks like some of your old work is in this PR:

> fix(UI): explore page snaps have lower case letters

I'm really horrible with Git, but shouldn't be rebase on the latest master @Feichtmeier ?

Feichtmeier:
yes, he based this branch on an old branch of him
however looking at the diff in this specific case it went okay
grafik
and since we will squash and merge anyways it does not really matter here

PR #1225, opened 2023-05-08
ExploreModel: return new map in getter

d-loose:
This makes sure we don't read entries from the `startPageApps` map while it's being changed asynchronously, leading to a `ConcurrentModificationError` (the cause for the most recent [integration test failures](https://github.com/ubuntu-flutter-community/software/actions/runs/4610387276/jobs/8148797561#step:11:316)).

### DIFF
--- a/lib/app/explore/explore_model.dart
+++ b/lib/app/explore/explore_model.dart
@@ -54,7 +54,7 @@ class ExploreModel extends SafeChangeNotifier {
         notifyListeners();
       }).then(
         (_) => Future.forEach<SnapSection>(
-          startPageApps.keys,
+          _startPageApps.keys,
           _loadStartPageAppstreamComponents,
         ),
       );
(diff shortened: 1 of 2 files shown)


d-loose:
After investigating a bit, it turns out both of my statements are wrong 🙃:
The `ConcurrentModificationError` seems to be caused by the explicit `_loadStartPageSnaps(SnapSection.all);` in `ExploreModel.init()`. But this exception doesn't cause the test failure (flutter seems to be fine with exceptions that occur after the test is done) - it's the following, which I can't reproduce anywhere outside the github CI:
```
The following StateError was thrown running a test (but after the test had completed):
Bad state: Client is closed

When the exception was thrown, this was the stack:
#0      _HttpClient._openUrl (dart:_http/http_impl.dart:2720:7)
#1      _HttpClient.getUrl (dart:_http/http_impl.dart:2609:48)
#2      SnapdClient._getSync (package:snapd/src/snapd_client.dart:1321:23)
#3      SnapdClient.getSnap (package:snapd/src/snapd_client.dart:1058:24)
#4      SnapService.findLocalSnap (package:software/services/snap_service.dart:116:33)
#5      SnapModel._findLocalSnap (package:software/app/common/snap/snap_model.dart:307:20)
#6      SnapModel.init (package:software/app/common/snap/snap_model.dart:43:24)
<asynchronous suspension>
```

PR #1226, opened 2023-05-09
Integration test: fix netplan

### LINKED ISSUE #1223: Fix integration test
@d-loose the test is failing and I don't know why. Could you eventually try to fix it when you have time?

d-loose:
Use a config file to use NetworkManager as the default renderer for netplan.
Setting the renderer via the CLI seems to not work anymore. 

Close #1223


Feichtmeier (approved):
Nice, ty!

PR #1234, opened 2023-05-18
Use the new switch expression

Feichtmeier:


lib/app/settings/settings_page.dart
```diff
@@ -154,28 +142,15 @@ class ThemeSection extends StatefulWidget {
...
-          {
-            App.themeNotifier.value = ThemeMode.dark;
-          }
-          break;
+  void onChanged(int index) {
+    for (var themeMode in ThemeMode.values) {
+      if (themeMode.index == index) {
+        setState(() {
+          App.themeNotifier.value = themeMode;
+        });
```
jpnurmi, line 150:
```dart
setState(() {
  App.themeNotifier.value = ThemeMode.values[index];
});
```

lib/app/common/app_format.dart
```diff
@@ -24,14 +24,10 @@ enum AppFormat {
...
   String localize(AppLocalizations l10n) {
-    switch (this) {
-      case AppFormat.snap:
-        return l10n.snapPackages;
-      case AppFormat.packageKit:
-        return l10n.debianPackages;
-      default:
-        return '';
-    }
+    return switch (this) {
```
jpnurmi, line 27: looks subjectively better than the floating indented switch above and below

jpnurmi (approved):
LGTM :+1:

lib/app/settings/settings_page.dart
Feichtmeier: Oh 🤯 you're right. Now it's merged. I'll do a continuous pr

lib/app/common/app_format.dart
Feichtmeier: ah you mean the return thing is better then the expression body because the lines are not that long this way?

PR #1248, opened 2023-05-30
fix: Installed snap list taking long to load

### LINKED ISSUE #1219: Installed snap list taking long to load
[Screencast from 2023-04-23 17-35-45.webm](https://user-images.githubusercontent.com/2129497/233849656-6429d0ba-2e6c-4bc9-bad5-4a222eba006e.webm)

Feichtmeier:
- let the installed snaps load as fast as possible in the UI then load the updates after
- forward the list from the service to the model, what was the reason for this unmodifyable list here @d-loose ? No critique, just curious
- make the list nullable, which gives one more state and does not leave the user with the "no snaps installed" message for some seconds while the list is loading
- to not disable the not-updates-rows while checking for updates, I see no reason why

Fixes #1219

[Bildschirmaufzeichnung vom 2023-05-30, 20-11-47.webm](https://github.com/ubuntu-flutter-community/software/assets/15329494/0701b6d6-990b-40c8-a489-779525cd2165)

### DIFF
--- a/lib/app/collection/snap_collection.dart
+++ b/lib/app/collection/snap_collection.dart
@@ -115,7 +115,7 @@ class SnapCollection extends StatelessWidget {
                                 : (i == installedSnaps.length - 1
                                     ? CollectionTilePosition.bottom
                                     : CollectionTilePosition.middle)),
-                        enabled: checkingForSnapUpdates == false,
+                        enabled: true,
                         onTap: () => SnapPage.push(
                           context: context,
                           snap: snap,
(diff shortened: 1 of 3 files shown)


lib/app/collection/snap_collection.dart
```diff
@@ -115,7 +115,7 @@ class SnapCollection extends StatelessWidget {
                                 : (i == installedSnaps.length - 1
                                     ? CollectionTilePosition.bottom
                                     : CollectionTilePosition.middle)),
-                        enabled: checkingForSnapUpdates == false,
+                        enabled: true,
```
d-loose, line 118: Nittiest of nitpicks: `!checkingForSnapUpdates`

Feichtmeier: oh, eventually it would be better to have a second boolean getter for the time the installed apps are being looked up? 
I just set this custom widget property to always true because I thought opening and removing installed snaps should not cause any trouble when looking for updates

d-loose: Ah sorry, my mind confused the added and removed lines.. I think it's fine to set it to `true` and decouple the widget from the loading state :+1:

PR #1249, opened 2023-05-31
Use cached_netowork_image for app banners and icons

### LINKED ISSUE #1116: Loading images takes a long time
When viewing the **Games** category, the image banners takes a while to load. The frustrating thing is that if I click another category (eg. **Art & Design**) and then back to **Games**, it has to reload the images again. Although far from that obvious, the same is also somewhat true for icons.

Can we somehow cache these images? I would imagine that these won't change that dramatically over time, so having them stored the next time the app is launched would be useful.

I tried fiddling with the image quality, but I couldn't make that change anything for better or worse.

image

dejvizelo:
[`cached_network_image`](https://pub.dev/packages/cached_network_image) makes it possible to cache app icons and banners so that the images are not fetched from their url every time, making image loading faster.

Close #1116

### DIFF
--- a/lib/app/common/app_banner.dart
+++ b/lib/app/common/app_banner.dart
@@ -15,6 +15,7 @@
  *
  */
 
+import 'package:cached_network_image/cached_network_image.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_rating_bar/flutter_rating_bar.dart';
 import 'package:provider/provider.dart';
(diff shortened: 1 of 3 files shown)


Feichtmeier:
@dejvizelo works and I like the fade in!

Without screen recording I also can not see any decrease in fps

[Bildschirmaufzeichnung vom 2023-06-01, 22-47-18.webm](https://github.com/ubuntu-flutter-community/software/assets/15329494/4067319f-f40c-428f-80cc-1117795c9ec0)

Feichtmeier (approved):

Feichtmeier:
The integration test is a bit fragile sometimes, I retriggered it

PR #1250, opened 2023-06-01
feat(review): validate input

jpnurmi:
Disable the _Submit_ button until the required conditions are met:
- rating is not optional
- name (user_display) is not optional
- title (summary) must be 2-70 characters
- review (description) must be 2-3000 characters

https://github.com/GNOME/odrs-web/blob/master/odrs/views_api.py

| Before | After (invalid) | After (valid) |
|---|---|---|
| image | Screenshot from 2023-06-01 22-06-27 | Screenshot from 2023-06-01 22-06-52 |


jpnurmi:
There's no ticket to close yet. I just started looking at the possibility of submitting reviews and there are multiple problems to solve. This stood out right away so I decided to start by fixing the easy and obvious...

Feichtmeier:
Yes sorry I have realized this! Thanks for the fix!

Issue #1186, opened 2023-03-23
'write a review' popup: changes to the form


anasereijo:
A few points on this (I can then open separate issues, if better). Open to discussion, as usual :) :)

- title of pop to change to 'Write a review' (I proposed [changing the button](https://github.com/ubuntu-flutter-community/software/issues/1178) as well)
- we don't have the titles for each input - was that deliberate?
- do we need to have a field for the name? If so, it should be mandatory. I checked the old store, and the name isn't requested, but is added to all reviews, so seems to be automatically included - can we keep like that?
- can we keep summary as mandatory and change the label to 'Summary' instead of 'Review Title'?
- size of the stars to be the same size as on app page (I [requested to be bigger](https://github.com/ubuntu-flutter-community/software/issues/1183))
- add the hover state to the stars as r[equested for the app-page](https://github.com/ubuntu-flutter-community/software/issues/1181)
- privacy policy text and link missing
- change 'Send' button to 'Submit'

wireframe:
<img width="842" alt="Screenshot 2023-03-23 at 17 14 00" src="https://user-images.githubusercontent.com/36884067/227287972-a4176ed3-ec92-43b4-97d1-0962183628e1.png">

flutter store:
<img width="413" alt="Screenshot 2023-03-23 at 17 27 45" src="https://user-images.githubusercontent.com/36884067/227288085-c1266326-3844-48be-9ef7-b04219a59023.png">

old store:
<img width="788" alt="Screenshot 2023-03-23 at 17 16 54" src="https://user-images.githubusercontent.com/36884067/227287897-183a5f3e-39e8-487b-a59c-e07fbb53b476.png">

madsrh:
> we don't have the titles for each input - was that deliberate?

This is the [Material design](https://m3.material.io/components/text-fields/overview).

In [your screenshot](https://user-images.githubusercontent.com/36884067/227288085-c1266326-3844-48be-9ef7-b04219a59023.png) the placeholder text IS the title and moves up when you click the field. If we add a hint text, it is only shown when you click the field and until you start typing.

![Ikke-navngivet-1](https://user-images.githubusercontent.com/3986894/227568621-0f7a735b-ad4e-46a5-8dc6-2987f2244e24.gif)

Would it be better to do "regular" text field label?

![image](https://user-images.githubusercontent.com/3986894/227572359-ba49453c-3441-4e32-89fc-f3ac73ddf65b.png)

anasereijo:
[...]

PR #1257, opened 2023-06-08
refactor(review): remove user name field

jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2023-06-08 17-07-30 | Screenshot from 2023-06-08 17-11-58 |

The glib library offers a way to get the user's [real name](https://docs.gtk.org/glib/func.get_real_name.html) and we have that available in [glib.dart](https://pub.dev/documentation/glib/latest/glib/GUtilsMixin/getRealName.html). It will be used in the service when submitting reviews, similarly to what gnome-software does.

See also:
- #1186


Feichtmeier (approved):
Ty

PR #1268, opened 2023-06-16
reviews: filter dropdown missing from the 'show all reviews' dropdown

### LINKED ISSUE #1177: reviews: filter dropdown missing from the 'show all reviews' dropdown
the filter is missing from the 'show all reviews' dropdown

Screenshot 2023-03-22 at 15 52 21

wireframe:

image

[[original issue](https://github.com/ubuntu-flutter-community/software/issues/751) with all designs for reviews and ratings]

Feichtmeier:
@jpnurmi to keep the app page and its children independent of any models I've moved the state just to the dialog itself. In this case this should be okay since the sorting is not something to persist (yet?).

If the other designers want the sorting to be also inside the main page, I could create a separate state for this there

[Bildschirmaufzeichnung vom 2023-06-16, 13-58-42.webm](https://github.com/ubuntu-flutter-community/software/assets/15329494/e5a75b9f-f220-4755-9521-86974e593992)

Fixes #1177

### DIFF
--- a/lib/app/common/app_page/app_page.dart
+++ b/lib/app/common/app_page/app_page.dart
@@ -33,10 +33,11 @@ import 'package:software/app/common/link.dart';
 import 'package:software/app/common/safe_network_image.dart';
 import 'package:software/app/explore/explore_model.dart';
 import 'package:software/l10n/l10n.dart';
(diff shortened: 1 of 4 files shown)


jpnurmi (approved):
LGTM :+1:

PR #1277, opened 2023-07-14
feat: about page

jpnurmi:
image


d-loose:
Could we generalize and re-use `SnapIcon` for better image quality and caching?
Also, how about sorting contributors alphabetically?

PR #1286, opened 2023-07-20
feat(navi-rail): push manage & about to the bottom

jpnurmi:
| Before | After |
|---|---|
| image | image |

### DIFF
--- a/lib/src/store/store_pages.dart
+++ b/lib/src/store/store_pages.dart
@@ -54,7 +54,7 @@ final pages = <StorePage>[
     // TODO: this is a counter-spacer for YaruNavigationRail's spacer (with
     // default flex = 1) after the last item. we should add proper support for
     // spacers and dividers in YaruNavigationRail.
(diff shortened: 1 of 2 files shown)


lib/src/store/store_pages.dart
```diff
@@ -1,45 +1,81 @@
...
+    pageBuilder: (_) => const GamesPage(),
   ),
   (
-    icon: DevelopmentPage.icon,
-    labelBuilder: DevelopmentPage.label,
-    builder: (_) => const DevelopmentPage(),
+    // TODO: this is a counter-spacer for YaruNavigationRail's spacer (with
+    // default flex = 1) after the last item. we should add proper support for
+    // spacers and dividers in YaruNavigationRail.
+    itemBuilder: (context, selected) => const Spacer(flex: 2 << 53),
```
d-loose, line 57: should be `1 << 53` if you want it to be 2^53

jpnurmi: thanks :D

PR #1288, opened 2023-07-20
feat: tweak paddings & geometries

jpnurmi:
`kYaruPagePadding` (20) is not divisible by 8. This PR introduces a local `kPagePadding` (16) constant to make it easier to match things with a layout grid.

NOTE: `YaruBanner` has a hard-coded 4px margin around for the hover effect. This has been taken into account so that the visual spacing between cards becomes 16px.

image


d-loose (approved):
LGTM :+1:

PR #1289, opened 2023-07-20
fix(search): reduce icon size

jpnurmi:
_Originally posted by @Feichtmeier in https://github.com/ubuntu/app-store/issues/1287#issuecomment-1644088825_

              ((( the search field content padding now makes the icon and the text to be not in one line anymore
grafik
I guess either change the content padding or decrease icon and text size
)))

| | |
|---|---|
| image | image |


d-loose (approved):
LGTM :+1:

PR #1290, opened 2023-07-21
refactor: introduce `SnapModel`

d-loose:
UDENG-791, UDENG-796

Replaces the snap providers with an old-fashioned `ChangeNotifier`.
I consider moving more of the presentation logic from the detail page into the view model later on.

@anasereijo we'd need your input  on what snap metadata should be shown on the app page depending on:
 - whether a snap is already installed
 - whether a snap is available in the store
 - which channel is currently selected
 - any combination of the above


jpnurmi (approved):
LVGTM :+1:

PR #1291, opened 2023-07-21
feat: implement responsive grid with fixed card sizes

d-loose:
UDENG-1048

Intermediate widths between 800 and 1280 look a bit empty - I'll see if the medium sized cards work better there next week (and maybe a single column for narrow widths).

A weird issue that puzzled me for a bit: even though the minimum window size is set to 800x600 in the GTK app, the window can be resized down to 748x548 - is there an explanation for that difference?

|resolution|screenshot|
|-|-|
|800x600|Screenshot from 2023-07-21 17-32-22|
|1280x800|Screenshot from 2023-07-21 17-33-07|
|1680x1024|Screenshot from 2023-07-21 17-34-01|

### DIFF
--- a/lib/constants.dart
+++ b/lib/constants.dart
@@ -4,16 +4,14 @@ const kAppName = 'App Store';
 const kGitHubRepo = 'ubuntu/app-store';
 
 const kCardMargin = 4.0;
-const kNaviRailWidth = 205.0;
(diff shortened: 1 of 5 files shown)


jpnurmi (approved):
LGTM :+1:

PR #1298, opened 2023-07-25
feat(detail): add localized 'last updated' date

d-loose:
UDENG-1048

Logic to be refined later together with the rest of the snap metadata.

### DIFF
--- a/lib/src/detail/detail_page.dart
+++ b/lib/src/detail/detail_page.dart
@@ -3,6 +3,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_markdown/flutter_markdown.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:intl/intl.dart';
 import 'package:snapd/snapd.dart';
 import 'package:ubuntu_widgets/ubuntu_widgets.dart';
 import 'package:yaru_icons/yaru_icons.dart';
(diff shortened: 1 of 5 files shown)


lib/src/detail/detail_page.dart
```diff
@@ -67,6 +70,12 @@ class _SnapView extends ConsumerWidget {
           value: context.formatByteSize(
               channelInfo?.size ?? model.storeSnap!.downloadSize!)
         ),
+      (
+        label: l10n.detailPageLastUpdatedLabel,
+        value: DateFormat.yMd(
+                Platform.environment['LC_TIME'] ?? Platform.localeName)
```
jpnurmi, line 76: is there any intl function that could be called on startup to initialize things in one place?

jpnurmi (approved):
LGTM :+1:

lib/src/detail/detail_page.dart
d-loose: Hm, I don't quite understand the [docs](https://pub.dev/packages/intl#date-formatting-and-parsing):
 > Note that before doing any DateTime formatting for a particular locale, you must load the appropriate data by calling.
 ```dart
import 'package:intl/date_symbol_data_local.dart';
...
initializeDateFormatting('de_DE', null).then(formatDates);
```

`DateFormat.yMd()` still needs to be called with the correct locale as an argument, though

Comment (jpnurmi) [reply lib/src/detail/detail_page.dart:76]: It probably won't respect `LC_TIME` specifically but there's a helper function like this in `ubuntu_localizations` (originates from UDI):

```dart
/// Initializes [Intl.defaultLocale] for formatting dates, times, and numbers.
///
/// If [locale] is not specified, the system locale is used.
///
/// See also:
/// * [Intl.defaultLocale]
/// * [Intl.systemLocale]
Future<void> initDefaultLocale([String? locale]) async {
  Intl.defaultLocale = locale ?? await findSystemLocale();
}
```

PR #1299, opened 2023-07-26
feat: responsive detail page

d-loose:
UDENG-1048

First iteration on a responsive detail page whose width matches that of the grid of cards.

[Screencast from 2023-07-26 17-27-56.webm](https://github.com/ubuntu/app-store/assets/113362648/d6965c92-3338-4781-b151-243848273577)

### DIFF
--- a/lib/constants.dart
+++ b/lib/constants.dart
@@ -3,15 +3,6 @@ import 'package:flutter/material.dart';
 const kAppName = 'App Store';
 const kGitHubRepo = 'ubuntu/app-store';
 
-const kCardMargin = 4.0;
-const kPaneWidth = 204.0;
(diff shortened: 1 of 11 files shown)


jpnurmi (approved):
LGTM :+1:

Feichtmeier:
Borders aside, but why don't you use the AppPage? 
We specifically designed the page to be responsive to 3 sizes of the screen. Utilizing the ride side more when the window is really big:

[Bildschirmaufzeichnung vom 2023-07-27, 00-33-47.webm](https://github.com/ubuntu/app-store/assets/15329494/87bcb307-8d58-4f4b-af37-3c9f2b2e751c)

@anasereijo ?

PR #1300, opened 2023-07-27
feat: add progressProvider

d-loose:
Adds a provider that returns a stream of the total progress values for a list of change IDs


jpnurmi (approved):
LGTM :+1:

PR #1303, opened 2023-07-28
fix(detail): ensure either local or store snap is available

d-loose:
Waits for `getStoreSnap` to emit at least one event before building the detail page


jpnurmi (approved):
LGTM :+1:

PR #1304, opened 2023-07-28
feat: add method to cancel ongoing change

d-loose:
Assumes that there is only a single active change, triggered by the user.


jpnurmi (approved):
LGTM :+1:

PR #1307, opened 2023-08-01
feat: add secondary actions button

d-loose:
UDENG-1048

[Screencast from 2023-08-01 12-26-28.webm](https://github.com/ubuntu/app-store/assets/113362648/352f794d-5877-4356-99c9-c5b287bea6f7)


lib/layout.dart
```diff
@@ -5,6 +5,7 @@ const kPaneWidth = 204.0;
 const kPagePadding = 16.0;
 const kSearchBarWidth = 424.0 - 2 * kCardMargin;
 const kIconSize = 56.0;
+const kPrimaryActionButtonWidth = 136.0;
```
jpnurmi, line 8: i'm confused :) isn't this the same as the existing push button width?

jpnurmi (approved):
LGTM :+1:

PR #1310, opened 2023-08-02
feat: manage page plumbing and updates

d-loose:
UDENG-794

Adds a view model for the manage page that exposes an 'update all' method and provides separate lists for locally installed snaps with and without available updates.

There's some overlap with the detail page's view model - maybe we can extract the waiting logic into a separate class/mixin?


jpnurmi (approved):
Nice, LGTM :+1:

PR #1314, opened 2023-08-04
refactor: add SnapCategoryEnum

d-loose:
Adds more structure for the snap categories in the spirit of the former `SnapSection`.


jpnurmi (approved):
LGTM :+1:

PR #1315, opened 2023-08-04
feat: add explore page banner

d-loose:
UDENG-1048

[Screencast from 2023-08-04 17-41-02.webm](https://github.com/ubuntu/app-store/assets/113362648/4d647ea4-7541-41d5-8a2d-de8453d2a369)

Will import the rest of @Zoospora's gradient colors later.


jpnurmi (approved):
LGTM :+1: 

If you have spare time in the future, it'd be strongly recommended to use `CustomScrollView` + `SliverGrid`. It would allow mixing the grid with other stuff in the same scroll view without having to turn on shrink wrapping.

PR #1320, opened 2023-08-08
feat: localize snap confinement and add shield icon

d-loose:
UDENG-1048

### DIFF
--- a/lib/src/detail/detail_page.dart
+++ b/lib/src/detail/detail_page.dart
@@ -64,9 +64,20 @@ class _SnapView extends ConsumerWidget {
       ), // Placeholder
       (
         label: l10n.detailPageConfinementLabel,
-        value: Text(
-          snapModel.channelInfo?.confinement.name ??
-              snapModel.snap.confinement.name,
+        value: Row(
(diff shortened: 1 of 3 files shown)


lib/src/detail/detail_page.dart
```diff
@@ -506,9 +517,18 @@ class _ChannelDropdownEntry extends StatelessWidget {
 }
 
 extension SnapdChangeL10n on SnapdChange {
-  String? localize(AppLocalizations l10n) => switch (kind) {
+  String localize(AppLocalizations l10n) => switch (kind) {
         'install-snap' => l10n.snapActionInstallingLabel,
         'remove-snap' => l10n.snapActionRemovingLabel,
-        _ => null,
+        _ => kind,
```
jpnurmi, line 523: is this intentional? would this show unlocalized strings (link-snap, mount-snap, validate-snap, ...) in the GUI?

d-loose: Reverting this bit - will thing about that part later™

jpnurmi (approved):
LGTM :+1:

PR #1324, opened 2023-08-10
feat: generalize search provider to support 'category'

d-loose:
Preparation step to generalize the search page (links to a specific category on the explore page will open a general search page with an empty query and a pre-selected category filter).
Also: autodispose the search providers, as recommended [here](https://riverpod.dev/docs/concepts/modifiers/family#prefer-using-autodispose-when-the-parameter-is-not-constant).


jpnurmi (approved):
LGTM :+1:

PR #1325, opened 2023-08-10
fix: don't autodispose search providers

d-loose:
This seems to cause problems with the autocompletion - let's revert this for now and come back to it in case caching too many searches causes any issues in the future.


jpnurmi (approved):
LGTM :+1:

PR #1327, opened 2023-08-11
feat: add 'snap://' argument handler

d-loose:
Handles 'snap://foo' as an optional positional cli argument

### DIFF
--- a/lib/src/store/store_providers.dart
+++ b/lib/src/store/store_providers.dart
@@ -30,8 +30,8 @@ String? _parseRoute(List<String>? args) {
   parser.addOption('search', valueHelp: 'query', help: 'Search for snaps');
 
   try {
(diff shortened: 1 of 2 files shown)


lib/src/store/store_providers.dart
```diff
@@ -28,6 +30,12 @@ String? _parseRoute(List<String>? args) {
   parser.addOption('search', valueHelp: 'query', help: 'Search for snaps');
 
   try {
+    if ((args?.isNotEmpty ?? false) && args![0].startsWith(_kUrlPrefix)) {
```
jpnurmi, line 33: it's a matter of taste but perhaps `firstOrNull` would be easier to parse? :)

jpnurmi (approved):
LGTM :+1:

PR #1333, opened 2023-08-15
feat(manage): show progress in 'update all' button

d-loose:
* Disable refresh button when an update is in progress
* Disable 'update all' button while checking for updates
* Show progress indicator in the 'update all' button during an update (similar to the primary action button on the detail page)

### DIFF
--- a/lib/src/manage/manage_page.dart
+++ b/lib/src/manage/manage_page.dart
@@ -127,6 +127,7 @@ class _ManageView extends ConsumerWidget {
   }
 }
 
+// TODO: refactor/generalize - similar to `_SnapActionButtons`
 class _ActionButtons extends ConsumerWidget {
   const _ActionButtons();
 
@@ -152,7 +153,9 @@ class _ActionButtons extends ConsumerWidget {
       mainAxisSize: MainAxisSize.min,
       children: [
         PushButton.outlined(
-          onPressed: ref.read(updatesModelProvider).refresh,
+          onPressed: updatesModel.activeChangeId != null
+              ? null
+              : ref.read(updatesModelProvider).refresh,
           child: Row(
             mainAxisSize: MainAxisSize.min,
             children: [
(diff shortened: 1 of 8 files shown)


lib/src/manage/manage_page.dart
```diff
@@ -152,7 +153,9 @@ class _ActionButtons extends ConsumerWidget {
       mainAxisSize: MainAxisSize.min,
       children: [
         PushButton.outlined(
-          onPressed: ref.read(updatesModelProvider).refresh,
+          onPressed: updatesModel.activeChangeId != null
+              ? null
+              : ref.read(updatesModelProvider).refresh,
```
CarlosNihelton, line 158: Honest question: you cannot reuse the `updatesModel` local variable because it was created with `ref.watch`, can you?

CarlosNihelton (approved):
LGTM  👍🏽 

My questions are more for my own learning.

lib/src/manage/manage_page.dart
d-loose: Oh, you totally can, thanks for spotting!
It's just a force of habit to always use `read` when calling methods on a provider

lib/src/manage/manage_page.dart
```diff
@@ -170,23 +173,53 @@ class _ActionButtons extends ConsumerWidget {
...
-              const SizedBox(width: 8),
-              Flexible(
-                child: Text(
-                  l10n.managePageUpdateAllLabel,
-                  maxLines: 1,
-                  overflow: TextOverflow.ellipsis,
+          child: updatesModel.activeChangeId != null
+              ? Consumer(
+                  builder: (context, ref, child) {
+                    final change = ref
```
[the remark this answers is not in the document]
d-loose: It's `null` while the `changeProvider`, which gives an `AsyncValue<SnapdChange>`, is fetching the data from snapd

PR #1334, opened 2023-08-15
feat: add dropdown for category filter

d-loose:
* Exposes the category filter, introduced in #1324, as a dropdown on the search result page
* Generalizes the search provider mocks accordingly and adds new test cases


lib/src/snapd/snap_category_enum.dart
```diff
@@ -40,6 +40,11 @@ enum SnapCategoryEnum {
   utilities,
   unknown;
 
+  bool get hidden => switch (this) {
```
CarlosNihelton, line 43: Is not this `switch` an overkill over a trivial Boolean expression? Even considering that the list of values resulting in `true` can grow, it will still be a one-liner conjunction expression.

CarlosNihelton (approved):
A minor question on readability, otherwise LGTM 👍🏽

lib/src/snapd/snap_category_enum.dart
d-loose: Totally :D
I'll add at least one more 'hidden' category in the near future, but this can still be simplified a little - thanks

PR #1335, opened 2023-08-15
Fix launchpad builds

kenvandine:
Workaround pub get stack overflow


snap/snapcraft.yaml
```diff
@@ -34,8 +35,9 @@ parts:
     source: .
     plugin: nil
     override-build: |
-      set -eux
-      flutter pub get
+      # work around pub get stack overflow # https://github.com/dart-lang/sdk/issues/51068#issuecomment-1396588253
+      set +e
+      dart pub get
```
d-loose, line 40: Should we still add `set -eux` after the pub get command?

d-loose (approved):
Thanks!
LGTM :+1:

PR #1338, opened 2023-08-16
fix: dropdown button only works once

d-loose:
Thanks to @gerwitz for spotting this! Clicking on the '...' dropdown buttons containing the secondary actions on the manage and app pages only works once.

Turns out [this](https://github.com/ubuntu/app-store/pull/1322#discussion_r1289682273) doesn't actually work, since the `YaruOptionButton` doesn't get rebuild when the controller state changes, so we have to pass a closure that explicitly checks the controller state.


CarlosNihelton (approved):
LGTM 👍🏽 Good catch! That's a really subtle issue to spot in code.

PR #1339, opened 2023-08-16
feat: add custom 'ubuntu desktop' category

d-loose:
We'd like to present a curated list of snaps on the explore page (to make sure users easily find all the applications that used to be part of the [default installation](https://discourse.ubuntu.com/t/rethinking-ubuntu-desktop-a-more-thoughtful-default-installation/36736) but aren't anymore). In the long run, it'd be great if an 'ubuntu desktop' category for that purpose could be added in the backend by the store team.
Until then this PR adds support for hand-picked 'featured' snaps and custom categories that aren't provided by the backend.

In order to receive the snaps from a custom category I added `getStoreSnaps` to the `SnapCache` extension. This method is similar to `getStoreSnap`, but accepts a list of snap names.

@CarlosNihelton the PR might look much larger than it actually is, due to some of the JSON data I added to the tests! I hope the 'multiple snaps' test group makes clear how the caching logic works in this case - let me know if you have any questions :)


lib/src/snapd/snap_category_enum.dart
```diff
@@ -38,13 +38,33 @@ enum SnapCategoryEnum {
...
+  unknown,
+  ubuntuDesktop;
 
-  bool get hidden => [unknown].contains(this);
+  bool get hidden => [unknown, ubuntuDesktop].contains(this);
 
   String get categoryName => name.replaceAllMapped(
       RegExp(r'[A-Z]'), (match) => '-${match[0]!.toLowerCase()}');
 
+  List<String>? get featuredSnapNames => switch (this) {
```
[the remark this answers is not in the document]
d-loose: I don't know if there are any concrete plans yet, but it definitely shouldn't stay in the desktop client indefinitely.
I'll ask @tim-hm after the summer break :)

lib/src/search/search_provider.dart
```diff
@@ -22,12 +22,19 @@ class SnapSearchParameters {
...
   final snapd = getService<SnapdService>();
-  return snapd.find(
-    query: searchParameters.query,
-    category: searchParameters.category?.categoryName,
-  );
+  if (searchParameters.category?.hidden ?? false) {
+    yield* snapd.getStoreSnaps(searchParameters.category!.featuredSnapNames
+            ?.where((name) => name.contains(searchParameters.query ?? ''))
+            .toList() ??
+        []);
```
CarlosNihelton, line 31: Why is this block desired to run if `category.hidden` is `true`?

CarlosNihelton: AH, I see: `ubuntuDesktop` is `hidden`. Sorry.

lib/src/snapd/snapd_cache.dart
```diff
@@ -77,6 +78,57 @@ mixin SnapdCache on SnapdClient {
       }
     }
   }
+
+  Stream<List<Snap>> getStoreSnaps(
```
CarlosNihelton, line 82: This is a quite complex chunk of code. It'll be good to have some documentation on it.

CarlosNihelton (approved):
Under the current circumstances this looks good enough for me.

lib/src/search/search_provider.dart
d-loose: Since we can't search for our custom category in the store (yet?), this just 'searches' the hand-picked list of featured snaps. Maybe re-using the `hidden` property isn't very clear - is it worth it to add another `custom` property?

lib/src/snapd/snapd_cache.dart
d-loose: Will add some!

lib/src/search/search_provider.dart
CarlosNihelton: If this is only due the nature of the `ubuntuDesktop` category we could just compare the category against that value, instead.

d-loose: Right, that's much clearer - thanks

lib/src/snapd/snap_category_enum.dart
```diff
@@ -38,13 +38,33 @@ enum SnapCategoryEnum {
   serverAndCloud,
   social,
   utilities,
-  unknown;
+  unknown,
+  ubuntuDesktop;
 
-  bool get hidden => [unknown].contains(this);
+  bool get hidden => [unknown, ubuntuDesktop].contains(this);
```
CarlosNihelton, line 44: Just checking: is having the `ubuntuDesktop` as part of the `hidden` property still desirable after your last commit?

anasereijo: Good point, however I am not 100% what we would call it, or even if it's too generic. But definitely something to think about in the future! 

We will have more of these banners in the (near) future, with 'hand-picked' snaps. The objective is to have banners with collections of snaps that are not necessarily existing categories. For now, we only have that banner for ubuntuDesktop, but in a few weeks/months we should be changing that so something else. We need to think of a way to easily create these 'categories'/collections as it's something we aim to be updating quite often

lib/src/snapd/snap_category_enum.dart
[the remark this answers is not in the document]
tim-hm: This isn't something we want to do long term ... so yes it should be coming directly from the store but I'm not sure when they'll have capacity to start curating these lists.

PR #1344, opened 2023-08-17
feat: show message if search results are empty

d-loose:
[Screencast from 2023-08-16 16-19-08.webm](https://github.com/ubuntu/app-store/assets/113362648/7241ab22-5a4c-4d86-a1ba-f2a517f4d90f)

### DIFF
--- a/lib/src/l10n/app_en.arb
+++ b/lib/src/l10n/app_en.arb
@@ -68,8 +68,9 @@
             }
         }
     },
+    "searchPageNoResultsHint": "Try a using different or more general keywords",
     "searchPageNoResultsCategory": "Sorry, we couldn’t find any packages in this category",
-    "searchPageNoResultsHint": "Try a different category or use more general keywords",
+    "searchPageNoResultsCategoryHint": "Try a different category or use more general keywords",
     "searchPageSortByLabel": "Sort by",
     "searchPageRelevanceLabel": "Relevance",
     "searchPageTitle": "Results for \"{query}\"",
--- a/lib/src/l10n/app_en.arb
+++ b/lib/src/l10n/app_en.arb
@@ -60,6 +60,16 @@
     "searchPageAlphabeticalLabel": "Alphabetical",
     "searchPageDownloadSizeLabel": "Download Size",
     "searchPageFilterByLabel": "Filter by",
+    "searchPageNoResults": "No results found for \"{query}\"",
+    "@searchPageNoResults": {
+        "placeholders": {
+            "query": {
+                "type": "String"
+            }
+        }
+    },
+    "searchPageNoResultsCategory": "Sorry, we couldn’t find any packages in this category",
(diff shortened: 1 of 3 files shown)


CarlosNihelton (approved):
LGTM 👍🏽

CarlosNihelton:
My understanding is that the empty category will be handled in a different PR. Is that right, @d-loose ?

madsrh:
@anasereijo 

English isn't my first language, but shouldn't it say _"...packages_ **in** _this category..."_ ?

d-loose:
> For the category empty state, I would like to have a different message:


Ah, I misunderstood the logic then. That message is currently only shown when a category doesn't contain any snaps (which I now realize is probably quite unlikely :D)

So if I understand it correctly now, the 'No results found for ... ' should be shown only when searching in 'All categories' and the other message when we search in a specific category?

anasereijo:
good catch @madsrh !!! 🙏  thank you

anasereijo:
> Looking again at Figma, the logic is a bit twisted:
> 
> This is with a category field selected: image
> 
> And this should be with no category selected (only mentions the keywords):
> 
> image

This is correct!

First example/screenshot: imagine you search for Spotify. We get results with that query, with the 'all categories' filter by default. If the user selects, let’s say, 'Finance' it won't give any results back.

Second screenshot: a query that doesn't return any apps at all.

d-loose:
Thanks everyone for helping me understand this very elaborate empty search logic :D I really need a vacation

Works like this now:

[Screencast from 2023-08-17 15-39-11.webm](https://github.com/ubuntu/app-store/assets/113362648/3508073b-b13f-4f3c-bc54-aefaf56b107b)

[Screencast from 2023-08-17 15-38-55.webm](https://github.com/ubuntu/app-store/assets/113362648/93b9ffd0-68b8-4dfe-accb-789121616067)

d-loose:
CI failures seem to be due to some network troubles .. :/

lib/src/l10n/app_en.arb
```diff
@@ -60,6 +60,17 @@
...
     "searchPageFilterByLabel": "Filter by",
+    "searchPageNoResults": "No results found for \"{query}\"",
+    "@searchPageNoResults": {
+        "placeholders": {
+            "query": {
+                "type": "String"
+            }
+        }
+    },
+    "searchPageNoResultsHint": "Try a using different or more general keywords",
```
CarlosNihelton, line 71: Try ~~a~~ using different or more...

CarlosNihelton (changes requested):
A last small detail and we'd be good to go

lib/src/l10n/app_en.arb
d-loose: Thanks a lot

CarlosNihelton (approved):
Now we are all set!

PR #1345, opened 2023-08-18
feat: update explore page

d-loose:
Final explore page updates with temporary workarounds, since we don't have any data for "What's new", "Top rated", and "Recently updated" yet. Instead we just show six snaps from a given category below the respective banners.

[Screencast from 2023-08-18 14-31-43.webm](https://github.com/ubuntu/app-store/assets/113362648/14e80a8a-0d2d-423e-9b8c-796e14968d15)

### DIFF
--- a/lib/src/explore/explore_page.dart
+++ b/lib/src/explore/explore_page.dart
@@ -108,18 +108,23 @@ class _CategorySnapList extends ConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final categorySnaps = ref
+        // get snaps from `category`
         .watch(searchProvider(SnapSearchParameters(category: category)))
         .whenOrNull(data: (data) => data)
+        // .. without the banner snaps, if we don't want them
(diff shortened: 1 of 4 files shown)


lib/src/explore/explore_page.dart
```diff
@@ -94,6 +90,54 @@ class _Title extends StatelessWidget {
...
+              : true,
+        );
+    final featuredSnaps = category.featuredSnapNames
+        ?.map((name) =>
+            categorySnaps?.singleWhereOrNull(((snap) => snap.name == name)))
+        .whereNotNull();
+    final snaps = (onlyFeatured ? featuredSnaps : categorySnaps)
+            ?.take(numberOfSnaps)
+            .toList() ??
+        [];
```
CarlosNihelton, line 126: This is not trivial stuff. It deserves some comments (and maybe even be extracted as a helper method).

CarlosNihelton (approved):
LGTM 👍🏽

PR #1346, opened 2023-08-18
feat: improve sorting and filtering

d-loose:
UDENG-797

* Generalizes the sorting functionality from the search page so that it can be reused elsewhere
* Adds sorting and filtering to the manage page

[Screencast from 2023-08-18 17-29-35.webm](https://github.com/ubuntu/app-store/assets/113362648/9d30c8eb-65ab-4b14-b70c-ef9e89984a21)

### DIFF
--- a/lib/src/manage/local_snap_providers.dart
+++ b/lib/src/manage/local_snap_providers.dart
@@ -1,0 +1,19 @@
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+import '/snapd.dart';
+import 'manage_model.dart';
+
+final localSnapFilterProvider = StateProvider.autoDispose<String>((_) => '');
+final showLocalSystemAppsProvider = StateProvider<bool>((_) => false);
+final localSnapSortOrderProvider =
+    StateProvider<SnapSortOrder>((_) => SnapSortOrder.alphabeticalAsc);
+final localSnapsProvider = Provider.autoDispose(
+  (ref) => ref
+      .watch(manageModelProvider.select((m) => m.nonRefreshableSnaps))
+      .where((snap) => snap.titleOrName
+          .toLowerCase()
+          .contains(ref.watch(localSnapFilterProvider).toLowerCase()))
+      .where((snap) =>
+          ref.watch(showLocalSystemAppsProvider) || snap.apps.isNotEmpty)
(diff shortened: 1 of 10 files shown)


lib/src/manage/local_snap_providers.dart
```diff
@@ -0,0 +1,19 @@
...
+final localSnapsProvider = Provider.autoDispose(
+  (ref) => ref
+      .watch(manageModelProvider.select((m) => m.nonRefreshableSnaps))
+      .where((snap) => snap.titleOrName
+          .toLowerCase()
+          .contains(ref.watch(localSnapFilterProvider).toLowerCase()))
+      .where((snap) =>
+          ref.watch(showLocalSystemAppsProvider) || snap.apps.isNotEmpty)
+      .sortedSnaps(ref.watch(localSnapSortOrderProvider)),
+);
```
CarlosNihelton, line 19: This seems hard to read at first, but the proxying of providers it offers is awesome!

lib/src/manage/manage_page.dart
```diff
@@ -106,13 +108,66 @@ class _ManageView extends ConsumerWidget {
...
+                    final sortOrder = ref.watch(localSnapSortOrderProvider);
+                    return MenuButtonBuilder<SnapSortOrder>(
+                      values: const [
+                        SnapSortOrder.alphabeticalAsc,
+                        SnapSortOrder.alphabeticalDesc,
+                        SnapSortOrder.installedDateAsc,
+                        SnapSortOrder.installedDateDesc,
+                        SnapSortOrder.installedSizeAsc,
+                        SnapSortOrder.installedSizeDesc,
+                      ],
```
CarlosNihelton, line 143: Did we consider the download size also as an ordering criteria?

CarlosNihelton (approved):
LGTM 👍🏽

lib/src/manage/manage_page.dart
d-loose: No, local snaps don't have a download size.

lib/src/manage/local_snap_providers.dart
d-loose: Yes, it works really well for this purpose :)

PR #567, opened 2022-11-24
Appstream service to expose metadata and search for packages.

oSoMoN:
Fixes #128, #464, #509.

PR #1353, opened 2023-09-06
feat: re-add appstream service

d-loose:
To kick-start support for debian packages, this PR re-adds the appstream service created by Olivier in #567.


tim-hm (approved):
@d-loose I've made a few tweaks as discussed on our call just now; over to you for review and merging if you're happy with them.

There is more to do here but lets keep things moving and we can do a wider review in due course.

lib/src/appstream/appstream_utils.dart
```diff
@@ -0,0 +1,118 @@
+import 'dart:io';
+import 'dart:ui';
+
+import 'package:appstream/appstream.dart';
+import 'package:collection/collection.dart';
+import 'package:ubuntu_logger/ubuntu_logger.dart';
+
+final log = Logger.setup(name: "appstream_utils");
```
d-loose, line 8: Can we stick to single quotes here and in all the other strings for consistency?
We could also add more [linter rules](https://dart.dev/tools/linter-rules/prefer_single_quotes) to avoid those nitpicks in the future

lib/src/appstream/appstream_utils.dart
```diff
@@ -0,0 +1,118 @@
...
+
+    for (final icon in local) {
+      if (File(icon.filename).existsSync()) {
+        return icon.filename;
+      }
+    }
+
+    final stock = icons.whereType<AppstreamStockIcon>().firstOrNull;
+    if (stock != null) {
+      throw UnimplementedError("Need to check if icon is on disk");
```
d-loose, line 112: I'd just leave a TODO note here rather than throw an exception, otherwise we'd need to handle the exception when displaying the icon in the UI

lib/src/appstream/appstream_utils.dart
```diff
@@ -0,0 +1,118 @@
...
+
+extension Localization on AppstreamComponent {
+  String getId() => id.toLowerCase();
+  String getPackage() => package?.toLowerCase() ?? '';
+
+  String getLocalizedName() {
+    final key = bestLanguageKey(name);
+
+    if (key == null) {
+      log.warning("Component does not have a name for locale");
```
d-loose, line 19: Now I'm also wondering if those log messages are useful at all. When I load the local appstream pool there are ~1500 of those messages, since the metadata for most packages is somewhat incomplete.

lib/src/appstream/appstream_utils.dart
```diff
@@ -0,0 +1,118 @@
...
+
+  List<String> getLocalizedSummary() {
+    final key = bestLanguageKey(summary);
+
+    if (key == null) {
+      log.warning("Component does not have summary for locale");
+      return [];
+    } else {
+      final results = summary[key] ?? '';
+      return results.toLowerCase().split(nonWordCharacters).toList();
```
d-loose, line 46: Tokenizing the fields here has the drawback that this extension can't be used to easily display their localized versions in the UI. It'd probably be better to keep the tokenization in `appstream_service.dart`.

lib/src/appstream/appstream_utils.dart
tim-hm: Opps double quotes are muscle memory! Will fix and enable the lint.

lib/src/appstream/appstream_utils.dart
tim-hm: Looking at this a bit closer, I think it should be removed until its needed ... its just not great to have lots of dead code around. For example this block is never used:

```
final cached = icons.whereType<AppstreamCachedIcon>().toList();
cached.sort((a, b) => sizeComparison(b.width ?? 0, a.width ?? 0));
```

lib/src/appstream/appstream_utils.dart
tim-hm: crikey! Okay will remove.

lib/src/appstream/appstream_utils.dart
tim-hm: From an outside perspective, is this not the same behaviour that existed in the previous implementation ... with my limited understanding at this point, aren't these tokens are already localised?

d-loose: Ah, before the refactoring we had this extension:
```dart
extension LocalizedComponent on AppstreamComponent {
  String localizedName({Locale? locale}) =>
      name[bestLanguageKey(name.keys, locale: locale)] ?? '';
  String localizedSummary({Locale? locale}) =>
      summary[bestLanguageKey(summary.keys, locale: locale)] ?? '';
  String localizedDescription({Locale? locale}) =>
      description[bestLanguageKey(description.keys, locale: locale)] ?? '';
}
```
which was really useful, but wasn't used in the service (only in the UI).
So ideally we could do the localization in the extension in `appstream_utils.dart` (which could then be used in UI code as well) and keep the `toLowerCase()` and tokenization in the factory constructor of the `_CachedComponent` in the service.

tim-hm: As the extension isn't private, is that same behaviour available with `component.getLocalizedName()`?

d-loose: Sorry, maybe my previous comments weren't clear.
The public `Localization` extension currently does what the now removed `LocalizedComponent` extension did, but additionally processes all the fields further (converting them to lower case and splitting them into tokens). While this is needed for the matching algorithm, it makes the extension useless for the UI, where we need the original localized fields without any processing.

So the extension would be more useful if it looked something like this and the processing was moved back into the service.

```dart
String getLocalizedName() {
    final key = bestLanguageKey(name);

    if (key == null) {
      return '';
    } else {
      return name[key] ?? '';
    }
  }

  List<String> getLocalizedKeywords() {
    final key = bestLanguageKey(keywords);

    if (key == null) {
      return [];
    } else {
      return keywords[key] ?? [];
    }
  }

  String getLocalizedSummary() {
    final key = bestLanguageKey(summary);

    if (key == null) {
      return '';
    } else {
      return summary[key] ?? '';
    }
  }

  String getLocalizedDescription() {
    final key = bestLanguageKey(description);

    if (key == null) {
      return '';
    } else {
      return description[key] ?? '';
    }
  }

  List<String> getLocalizedMediaTypes() {
    final List<String> mediaTypes = [];

    for (final provider in provides) {
      if (provider is AppstreamProvidesMediatype) {
        mediaTypes.add(provider.mediaType);
      }
    }

    return mediaTypes;
  }
```

tim-hm: Thanks for the explanation :pray: 

I've refactored the Localisation to follow the original extension name `LocalisedComponent` and restored the original behaviour (ie if you call `component.getLocalisedDescription()` it returns the string it its original form).

I've moved the lowercasing behaviour into the `_CachedComponent` constructor.

d-loose: Thanks :+1: 
Did you remove the entire icon extension on purpose?

tim-hm: Yes because there were issues, but I didn't have an appreciation for how/where it was going to be used. If we need to add it back in, we should tidy it up a little bit. For example:

```
final cached = icons.whereType<AppstreamCachedIcon>().toList();
cached.sort((a, b) => _sizeComparison(b.width, a.width));
// and now cached is never used as the subsequent lines are commented out
```

Another bit was:

```
if (stock != null) {
  // TODO: check whether the stock icon exists on disk, and return it
}
```

So if you're going to use it and rely on local + remote that makes sense, but lets understand when/how cached and stock are being used rather than just allowing the state-and-purpose-unknown code in.

d-loose: Alright thanks - will add it when I get to the UI bits

Issue #1363, opened 2023-09-11
(Re-)add and refactor PackageKit service


d-loose:
As before, we'll use Appstream (#1353) to search for packages and retrieve their metadata. PackageKit is needed to install/update/remove packages.
I'd like to refactor the old service a little so that it behaves more similar to the snapd service, i.e.
* Maintain a list of active transactions (similar to snapd changes)
* Return transaction IDs from long-lasting actions (like installing a package)
* Expose active transactions, so that we can easily track their progress in the UI

d-loose:
#1369

Issue #1215, opened 2023-04-11
unable to see debs tabs of installed apps


alculete:
![image](https://user-images.githubusercontent.com/2129497/231224881-f19b7951-f5ac-45e0-a73f-b395189a4b29.png)

soumyaDghosh:
Same issue here, seems like previously packagekit was slow to load, but now it's not loading anymore

Feichtmeier:
I can not confirm this on my machine. @madsrh does this also happen to you?

alculete:
Now I can see debs installations. It was like this for over a week. I will close the ticket

soumyaDghosh:
# This hasn't been fixed yet.
 **Still only snaps are shown.**
![Screenshot from 2023-04-24 00-28-48](https://user-images.githubusercontent.com/72045785/233859549-ed04f2d6-73be-4b15-9fa0-4914aa1fe14d.png)

alculete:
I think we are reporting different issues, mine was on installed apps that I couldn't see debs. Try to search GIMP and open it see if you have deb option

![image](https://user-images.githubusercontent.com/2129497/233859898-2af05793-0553-4e5d-aa5f-f57dd1b10a64.png)

madsrh:
> madsrh does this also happen to you?

Nope, everything looks normal here 🤷‍♂️ 

@soumyaDghosh the debs are not as fast as snaps, so you'll have to give the Explore page a few seconds to load before the deb icons appear.

> I think we are reporting different issues, mine was on installed apps that I couldn't see debs. Try to search GIMP and open it see if you have deb option.

+1 @alculete alculete

alculete:
having the same issue again after upgrading some snaps, closing snap store, re-opening, and going to manage
![image](https://user-images.githubusercontent.com/2129497/234170429-2214fe69-8af5-44b4-bfc1-7a77a1fffb60.png)

Feichtmeier:
> having the same issue again after upgrading some snaps, closing snap store, re-opening, and going to manage
> ![image](https://user-images.githubusercontent.com/2129497/234170429-2214fe69-8af5-44b4-bfc1-7a77a1fffb60.png)
> 

Thanks!

Could you open it via terminal and copy paste the output if there is any?

Feichtmeier:
@alculete does this still happen to you?

d-loose:
Seems like the `packagekit-control` plug is not connected.
[...]

PR #1369, opened 2023-09-11
feat: add packagekit service

d-loose:
UDENG-1354
Ref #1363

Adds `PackageKitService` with some basic functionality:
* Explicitly activates the DBus service (see previous discussions [here](#1215) and [here](https://forum.snapcraft.io/t/apparmor-denial-in-new-snap-store-despite-connected-packagekit-control-interface/35290))
* Keeps track of active transactions internally (it's probably be a good idea to implement [`GetTransactionList`](https://www.freedesktop.org/software/PackageKit/gtk-doc/PackageKit.html#PackageKit.GetTransactionList) in packagekit.dart in the future instead)
* Allows transactions to be observed from the outside (for progress information etc)
* Exposes essential methods (install/remove/resolve)


lib/src/packagekit/packagekit_service.dart
```diff
@@ -0,0 +1,129 @@
...
+  /// the daemon is inactive.
+  /// See https://github.com/ubuntu/app-center/issues/1215
+  /// and https://forum.snapcraft.io/t/apparmor-denial-in-new-snap-store-despite-connected-packagekit-control-interface/35290
+  Future<void> activateService() async {
+    if (_isAvailable) return;
+
+    final object = DBusRemoteObject(
+      _dbus,
+      name: 'org.freedesktop.DBus',
+      path: DBusObjectPath('/org/freedesktop/DBus'),
```
tim-hm, line 42: Is it worth moving some of these into variables as they're also repeated quite a bit in tests as well?

lib/src/packagekit/packagekit_service.dart
```diff
@@ -0,0 +1,129 @@
...
+      if (event is PackageKitFinishedEvent || event is PackageKitDestroyEvent) {
+        _transactions.remove(id);
+        subscription.cancel();
+      }
+    });
+    await action?.call(transaction);
+    return id;
+  }
+
+  /// Waits until the transaction specified by the internal `id` has finished.
```
tim-hm, line 79: One aspect to consider, is how much does a comment like this add? I can get the same detail simply by reading the method signature, so is it worth adding such a comment?

What if in time, the code diverges, then you've now also go to update the comment to keep in line with the code.

Comments are helpful, especially in circumstances where context/the why might be missing, or perhaps code is doing something funky which needs an additional explanation, but in situations like these I'm not sure its adding any value!

lib/src/packagekit/packagekit_service.dart
```diff
@@ -0,0 +1,129 @@
...
+  // that run in the background for longer (e.g. installing a package).
+  // The respective methods will return a transaction ID, similar to how methods
+  // in the snap client return change IDs.
+  PackageKitTransaction? getTransaction(int id) => _transactions[id];
+
+  /// Explicitly activates the PackageKit service in case it is not running.
+  /// Prevents AppArmor denials when trying to call a well-known method while
+  /// the daemon is inactive.
+  /// See https://github.com/ubuntu/app-center/issues/1215
+  /// and https://forum.snapcraft.io/t/apparmor-denial-in-new-snap-store-despite-connected-packagekit-control-interface/35290
```
tim-hm, line 35: Re my comment further down, these comments are very useful and a great example of when they add value :+1:

lib/src/packagekit/packagekit_service.dart
```diff
@@ -0,0 +1,129 @@
...
+    );
+    await object.callMethod(
+      'org.freedesktop.DBus',
+      'StartServiceByName',
+      const [DBusString('org.freedesktop.PackageKit'), DBusUint32(0)],
+    );
+    try {
+      await _client.connect();
+      _isAvailable = true;
+    } on DBusServiceUnknownException catch (_) {}
```
tim-hm, line 52: If we encounter an exception, what does that mean? Is it worth logging that something wrong wrong?

tim-hm (approved):
LGTM.

I've added a few comments for you to consider; go ahead and merge when you're happy.

lib/src/packagekit/packagekit_service.dart
d-loose: I don't like sharing constants between the code and the tests, but I'll clean up the tests a bit.

lib/src/packagekit/packagekit_service.dart
d-loose: No, nothing went wrong, it just means packagekit isn't available. @kenvandine asked us to handle this case gracefully so that we can easily use this in core desktop. I've added a comment.

PR #1371, opened 2023-09-13
Ratings intergration

matthew-hagemann:
Creating as a draft until I finish updating the tests, but probably best to start getting eyes on this now

- Created new provider `ratingsModelProvider` for getting rating of snap in the details_page.
- Created new service `ratingsService` that wraps the ratings client that can be found [here](https://github.com/matthew-hagemann/app_center_ratings_client.dart)


lib/src/detail/detail_page.dart
```diff
@@ -20,7 +21,12 @@ import '/widgets.dart';
 const _kPrimaryButtonMaxWidth = 136.0;
 const _kChannelDropdownWidth = 220.0;
 
-typedef SnapInfo = ({String label, Widget value});
+class SnapInfo {
```
d-loose, line 24: Does this need to be upgraded to a class or can we leave it as a record type?

lib/src/detail/detail_page.dart
```diff
@@ -31,8 +37,9 @@ class DetailPage extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     final snapModel = ref.watch(snapModelProvider(snapName));
     final updatesModel = ref.watch(updatesModelProvider);
+
     return snapModel.state.when(
-      data: (_) => ResponsiveLayoutBuilder(
+      data: (snapData) => ResponsiveLayoutBuilder(
```
d-loose, line 42: `state` is `AsyncValue<void>`, so the argument of the closure doesn't need to be named
```suggestion
      data: (_) => ResponsiveLayoutBuilder(
```

lib/src/detail/detail_page.dart
```diff
@@ -45,7 +52,10 @@ class DetailPage extends ConsumerWidget {
 }
 
 class _SnapView extends ConsumerWidget {
-  const _SnapView({required this.snapModel, required this.updatesModel});
+  const _SnapView({
+    required this.snapModel,
+    required this.updatesModel,
+  });
```
d-loose, line 58: It'd be great to keep pure formatting changes out of PRs like this, since they can clutter the diff a bit :)

lib/src/explore/explore_page.dart
```diff
@@ -108,18 +108,15 @@ class _CategorySnapList extends ConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final categorySnaps = ref
-        // get snaps from `category`
         .watch(searchProvider(SnapSearchParameters(category: category)))
         .whenOrNull(data: (data) => data)
-        // .. without the banner snaps, if we don't want them
         ?.where(
```
d-loose, line 113: Any particular reason you removed those comments?

lib/src/widgets/snap_card.dart
```diff
@@ -1,14 +1,17 @@
...
-import 'package:yaru/yaru.dart';
 import 'package:yaru_widgets/yaru_widgets.dart';
 
 import '/layout.dart';
 import '/snapd.dart';
 import '/widgets.dart';
 
 class SnapCard extends StatelessWidget {
-  const SnapCard({
+  SnapCard({
```
d-loose, line 14: Should stay `const` (prefer_const_constructors_in_immutables)

lib/src/widgets/snap_card.dart
```diff
@@ -38,7 +41,7 @@ class SnapCard extends StatelessWidget {
 }
 
 class SnapImageCard extends StatelessWidget {
-  const SnapImageCard({super.key, required this.snap, this.onTap});
+  SnapImageCard({super.key, required this.snap, this.onTap});
```
d-loose, line 44: Should stay `const` (prefer_const_constructors_in_immutables)

lib/src/widgets/snap_card.dart
```diff
@@ -67,7 +70,10 @@ class SnapImageCard extends StatelessWidget {
             flex: 306 - 160, // based on mockups
             child: Padding(
               padding: const EdgeInsets.all(kCardSpacing),
-              child: _SnapCardBody(snap: snap, maxlines: 1),
+              child: _SnapCardBody(
+                snap: snap,
+                maxlines: 1,
+              ),
```
d-loose, line 76: Unrelated style change

lib/src/widgets/snap_card.dart
```diff
@@ -77,7 +83,7 @@ class SnapImageCard extends StatelessWidget {
 }
 
 class _SnapCardBody extends StatelessWidget {
-  const _SnapCardBody({required this.snap, this.maxlines = 2});
+  _SnapCardBody({required this.snap, this.maxlines = 2});
```
d-loose, line 86: Should stay `const` (prefer_const_constructors_in_immutables)

lib/src/widgets/snap_card.dart
```diff
@@ -103,25 +109,47 @@ class _SnapCardBody extends StatelessWidget {
...
-            Text('|', style: Theme.of(context).textTheme.bodySmall),
-            const SizedBox(width: 2),
-            Text(
-              '200 Ratings',
-              style: Theme.of(context).textTheme.bodySmall,
-            ),
+            if (rating?.totalVotes != null) ...[
+              const SizedBox(width: 2),
+              Text(
+                ' | ${rating?.totalVotes} ${l10n.snapRatingsVotes}',
```
d-loose, line 144: It'd be better to use an l10n string that accepts a placeholder argument here, since the placement of the translated word(s) and the number of votes might not be the same in every language. Also makes it easier to test.

test/detail_page_test.dart
```diff
@@ -13,20 +16,27 @@ import 'package:yaru_widgets/widgets.dart';
...
+    publisher: const SnapPublisher(displayName: 'testPublisher'),
+    version: '2.0.0',
+    website: 'https://example.com',
+    confinement: SnapConfinement.classic,
+    license: 'MIT',
+    description: 'this is the **description**',
+    trackingChannel: 'latest/edge',
+    channel: 'latest/edge',
+    installDate: DateTime(1970),
+    id: "r4LxMVp7zWramXsJQAKdamxy6TAWlaDD");
```
d-loose, line 38: Could we get a trailing comma? :)

test/snap_card_test.dart
```diff
@@ -0,0 +1,44 @@
...
+        ],
+        child: SnapCard(snap: snap),
+      ),
+    );
+
+    await tester.pumpAndSettle();
+
+    expect(find.text('testsnap'), findsOneWidget);
+    expect(find.text('Good'), findsOneWidget);
+    expect(find.text(' | 123 votes'), findsOneWidget);
```
d-loose, line 42: Those tests should use localizations strings. You can access them via the l10n extension on the `WidgetTester` from the test utils

lib/src/detail/detail_page.dart
matthew-hagemann: I was getting the error "'SnapInfo' isn't a function.
Try correcting the name to match an existing function, or define a method or function named 'SnapInfo'." when trying to return SnapInfo from the _SnapInfos build class method. Not sure if there is a better way to solve this than turning it into a class?

d-loose (changes requested):
Could we also get that lib/ratings.dart we talked about and `import '/ratings.dart'` instead?

Hope the comments aren't too blunt - great work overall, thanks a lot!

matthew-hagemann:
Not quite sure what to make of CI failures.

Analyze wants me to swap to relative imports in lib/src, but I believe we want to avoid doing that? Is there somthing we want to update in `analysis_options.yaml`?
Mocks are freshly updated, not sure why that failed?

d-loose (approved):
Cleaned up imports and regenerated mocks. Also removed the loading spinner from the card, we can check back later with Ana to see if it's necessary, but I think it doesn't look great :)

LGTM :+1:

Issue #1367, opened 2023-09-11
Implement basic deb support


d-loose:
Depends on the design from #1365

PR #1372, opened 2023-09-13
feat: add appstream results to autocompletion

d-loose:
UDENG-1364
Ref #1367

Screenshot from 2023-09-13 15-06-44

@tim-hm when running this in debug mode loading the appstream metadata slows down the app drastically to the point where the UI becomes unresponsive for a short time. This should be hardly noticeable in release mode, though. However, there's some delay before the appstream autocompletion results become available - we should cache the metadata pool in a format that can be loaded more quickly later on.

I had to refactor the search providers a bit:
* src/search now only contains the common search logic (i.e. the search query and the autocompletion provider
* src/appstream and src/snapd now have their own separate search providers - there's no common filtering and sorting logic yet.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -1,11 +1,15 @@
+import 'dart:async';
+
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:github/github.dart';
 import 'package:gtk/gtk.dart';
 import 'package:snapcraft_launcher/snapcraft_launcher.dart';
+import 'package:ubuntu_logger/ubuntu_logger.dart';
 import 'package:ubuntu_service/ubuntu_service.dart';
 import 'package:yaru_widgets/yaru_widgets.dart';
 
+import 'appstream.dart';
 import 'l10n.dart';
 import 'snapd.dart';
 import 'store.dart';
@@ -24,7 +28,15 @@ Future<void> main(List<String> args) async {
   registerService(() => GitHub());
   registerService(() => GtkApplicationNotifier(args));
 
+  final appstream = AppstreamService();
+  // Explicitly ignore the future to continue while appstream is reading the
+  // metadata from the disk.
+  unawaited(appstream.init());
+  registerServiceInstance(appstream);
+
   await initDefaultLocale();
+
+  Logger.setup();
 
   runApp(const ProviderScope(child: StoreApp()));
 }
--- a/lib/src/search/search_field.dart
+++ b/lib/src/search/search_field.dart
@@ -1,3 +1,4 @@
+import 'package:appstream/appstream.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/scheduler.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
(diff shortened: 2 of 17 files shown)


lib/src/search/search_provider.dart
```diff
@@ -1,66 +1,29 @@
...
-});
-
-final sortOrderProvider =
-    StateProvider.autoDispose<SnapSortOrder?>((_) => null);
-
-final sortedSearchProvider = FutureProvider.family
-    .autoDispose((ref, SnapSearchParameters searchParameters) {
-  return ref.watch(searchProvider(searchParameters).future).then(
-      (snaps) => snaps.sortedSnaps(ref.watch(sortOrderProvider)).toList());
+  return (snaps: <Snap>[], debs: <AppstreamComponent>[]);
```
tim-hm, line 28: I don't understand the inner workings of dart async or the behaviours of many of the libraries at play here ... so take this with a grain of salt!

Normally if you want to race two Futures you'd do something like this:

```dart
final autoCompleteProvider = FutureProvider<AutoCompleteOptions>((ref) async {
  final query = ref.watch(queryProvider);
  final completer = Completer();
  ref.onDispose(completer.complete);
  await Future.delayed(const Duration(milliseconds: 100));

  if ((query?.isNotEmpty ?? true) && !completer.isCompleted) {
    final futures = [
      ref.watch(snapSearchProvider(SnapSearchParameters(query: query)).future),
      ref.watch(appstreamSearchProvider(query ?? '').future),
    ];

    final results = await Future.wait(futures);
    final snaps = results[0] as Iterable<Snap>;
    final debs = results[1] as Iterable<AppstreamComponent>;

    return (snaps: snaps, debs: debs);
  }
  return (snaps: <Snap>[], debs: <AppstreamComponent>[]);
});
```

tim-hm, line 28: Is this delay necessary?

lib/src/search/search_field.dart
```diff
@@ -44,18 +82,34 @@ class _SearchFieldState extends ConsumerState<SearchField> {
...
+        return <AutoCompleteOption>[
+          if (snapOptions.isNotEmpty) ...[
+            AutoCompleteSectionOption(l10n.searchFieldSnapSection),
+            ...snapOptions,
+            AutoCompleteDividerOption()
+          ],
+          if (debOptions.isNotEmpty) ...[
+            AutoCompleteSectionOption(l10n.searchFieldDebSection),
+            ...debOptions,
+            AutoCompleteDividerOption(),
```
tim-hm, line 107: There's something not quiet right here ... it could just be that you haven't implemented the feature yet so do ignroe if that's the case.

If i use my keyboard to navigate to a snap search result and hit enter it takes me to that snaps details page. When I do the same with a deb result it simply replaces my search text with the selected deb result text.

lib/src/search/search_provider.dart
d-loose: Good point, didn't even consider racing them, but there's no reason not to do it, even if the performance gain might be minimal. I'll change it

d-loose: The delay was introduced on purpose to avoid sending queries every time the user presses a key

lib/src/search/search_field.dart
d-loose: That's exactly what's supposed to happen at this stage, since the detail page has not been generalized to support debs yet.
See todo note [here](https://github.com/ubuntu/app-center/pull/1372/files/9c70c310b5d4b5e1e1c23fdb9bf2b755b612b029#diff-c0f5a00f4386bf642ae40054f0a78e9f63b9edb91b2b953610d4dec4434c57a1R49)

lib/src/store/store_app.dart
```diff
@@ -45,7 +45,8 @@ class _StoreAppState extends ConsumerState<StoreApp> {
               child: SearchField(
                 onSearch: (query) =>
                     _navigator.pushAndRemoveSearch(query: query),
-                onSelected: (name) => _navigator.pushDetail(name: name),
+                onSnapSelected: (name) => _navigator.pushDetail(name: name),
+                onDebSelected: (_) {}, // TODO: push detail page
```
tim-hm, line 49: I'd prefer to see this logged so its easier for someone else to notice this ... TODOs are fine for personal use, but difficult in a setting where there are 2+ engineers on a project

lib/main.dart
```diff
@@ -24,6 +27,10 @@ Future<void> main(List<String> args) async {
   registerService(() => GitHub());
   registerService(() => GtkApplicationNotifier(args));
 
+  final appstream = AppstreamService();
+  unawaited(appstream.init());
```
tim-hm, line 31: I would add a simple comment explaining why this

tim-hm:
A few bits for your consideration! Go ahead and merge when you're happy.

tim-hm (approved):

PR #1376, opened 2023-09-15
feat: adjust snapcraft.yaml for appstream+packagekit

d-loose:
UDENG-1364
Ref #1367

Reintroduces the interfaces needed to communicate with packagekit and access the appstream metadata.


tim-hm:
Merged at 44844cb

PR #1384, opened 2023-09-18
feat: add rate snap functionality

matthew-hagemann:


lib/src/ratings/ratings_model.dart
```diff
@@ -1,40 +1,90 @@
...
+      );
+      await ratings.vote(vote);
+      _vote = newVoteStatus;
+    }
+
+    notifyListeners();
+  }
+}
+
+enum VoteStatus {
```
d-loose, line 86: Would it be clearer to remove the 'none' variant and make `_vote` nullable? This way we could also use the enum as an argument in `castVote()` to avoid having a [positional boolean](https://dart.dev/tools/linter-rules/avoid_positional_boolean_parameters).

lib/src/snapd/snap_page.dart
```diff
@@ -314,11 +314,67 @@ class _SnapActionButtons extends ConsumerWidget {
           cancelButton
         else if (snapModel.isInstalled)
           secondaryActionsButton,
+        const SizedBox(width: 8),
```
d-loose, line 317: Nitpick: could we make the `SizedBox` conditional as well? e.g.
```dart
if (snapModel.isInstalled) ...[
    const SizedBox(width: 8),
    _RatingsActionButtons(
        snap: snapModel.snap,
    ),
]

File: lib/src/ratings/ratings_model.dart [hunk 3]
```diff
@@ -1,40 +1,90 @@
...
+        ratings.getRating(snapId),
+        ratings.getSnapVotes(snapId),
+      ]);
+
+      final rating = results[0] as Rating;
+      final votes = results[1] as List<Vote>;
+
       _setSnapRating(rating);
+      _setUserVote(votes);
       notifyListeners();
```
d-loose, line 50: Something I didn't catch earlier but am noticing now: the `notifyListeners()` call should be moved outside the `AsyncValue.guard()`, otherwise the UI might not update when the model updates its state.

d-loose: I'm realizing I'm not doing this correctly in other models either - just noticed this issue when testing your service with hardcoded responses. It works fine if I add an `await Future.delayed()` with an arbitrarily small duration.

matthew-hagemann:
Still stumped on the mocks issue. 
I did a `flutter clean` then ran the commands in the order they appear in the pipeline, on melos 3.1.1
`flutter pub global activate melos`
`melos pub get `
`melos generate `
Any ideas :sweat_smile: ?

d-loose (approved):
LGTM :+1:

Issue #1394, opened 2023-09-20
Log errors and other useful information


d-loose:
We need to log errors and other relevant information so that users can provide us with more detailed bug reports in the future.

anasereijo:
@d-loose is this still relevant?

d-loose:
Logging errors is of course always relevant! But I used this issue specifically for https://github.com/ubuntu/app-center/pull/1396 and https://github.com/ubuntu/app-center/pull/1418, so I'll close it now.

Issue #1362, opened 2023-09-11
Define and design error handling UX


d-loose:
We need to define how/if errors are presented to the user.
The most common case we need to handle is the failure of a user-initiated action (i.e. error during installation/refresh/removal - can happen e.g. because there's already another action in progress, or for various other reasons like network failure).

Maybe a simple modal dialog (like here: https://github.com/canonical/firmware-updater/pull/101) is enough for now?

anasereijo:
closing as I believe we covered all this in the recent work. @d-loose pls let me know if that's not the case.

PR #1396, opened 2023-09-20
feat: snapd error handling

### LINKED ISSUE #1366: Implement basic error handling for snap actions
Depends on the design from #1362

d-loose:
UDENG-1363
UDENG-1422
Close #1366
Ref #1394 

image
(This state isn't reachable in practice, since we don't show the update button unless there is an update available)

* Adds a rudimentary error dialog widget
* Adds error streams to `SnapModel` and `UpdateModel`
* Logs snapd exceptions caught in the `SnapdModel` and `UpdateModel`
* Shows an error dialog on the `SnapPage` and `ManagePage` when errors are caught in the underlying models

### DIFF
--- a/lib/snapd.dart
+++ b/lib/snapd.dart
@@ -7,4 +7,4 @@ export 'src/snapd/snap_search.dart';
 export 'src/snapd/snap_sort.dart';
 export 'src/snapd/snapd_service.dart';
 export 'src/snapd/snapx.dart';
-export 'src/snapd/updates_model.dart' hide log;
+export 'src/snapd/updates_model.dart';
--- a/lib/snapd.dart
+++ b/lib/snapd.dart
@@ -7,4 +7,4 @@ export 'src/snapd/snap_search.dart';
 export 'src/snapd/snap_sort.dart';
 export 'src/snapd/snapd_service.dart';
 export 'src/snapd/snapx.dart';
-export 'src/snapd/updates_model.dart';
+export 'src/snapd/updates_model.dart' hide log;
--- a/lib/src/manage/manage_page.dart
+++ b/lib/src/manage/manage_page.dart
@@ -46,7 +46,7 @@ class _ManagePageState extends ConsumerState<ManagePage> {
 
   Future<void> showError(SnapdException e) => showErrorDialog(
         context: context,
-        title: e.kind ?? '',
+        title: e.kind ?? 'Unknown Snapd Exception',
         message: e.message,
       );
 
--- a/lib/src/manage/manage_page.dart
+++ b/lib/src/manage/manage_page.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:snapd/snapd.dart';
@@ -13,7 +15,7 @@ import '/widgets.dart';
 import 'local_snap_providers.dart';
 import 'manage_model.dart';
 
-class ManagePage extends ConsumerWidget {
+class ManagePage extends ConsumerStatefulWidget {
   const ManagePage({super.key});
 
   static IconData icon(bool selected) => YaruIcons.app_grid;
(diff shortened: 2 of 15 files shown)


lib/src/snapd/updates_model.dart
```diff
@@ -1,10 +1,17 @@
...
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:meta/meta.dart';
 import 'package:snapd/snapd.dart';
+import 'package:ubuntu_logger/ubuntu_logger.dart';
 import 'package:ubuntu_service/ubuntu_service.dart';
 
 import '/snapd.dart';
 
+@internal
+final log = Logger('snap');
```
tim-hm, line 13: There is also a logger in snap_model.dart. I don't know what the convention is so ignore me if this doesn't apply, but would you normally have a single logger at the package / folder level?

lib/snapd.dart
```diff
@@ -7,4 +7,4 @@ export 'src/snapd/snap_search.dart';
 export 'src/snapd/snap_sort.dart';
 export 'src/snapd/snapd_service.dart';
 export 'src/snapd/snapx.dart';
-export 'src/snapd/updates_model.dart';
+export 'src/snapd/updates_model.dart' hide log;
```
tim-hm, line 10: Ref comment elsewhere about the log instance in snap_model.dart, you might want to hide it here as well (or consolidate to 1 logger instance)

tim-hm (changes requested):
I'm getting an error when trying to install a snap. Are you experiencing the same?

```
flutter: ERROR snap: Caught exception for snap "gnome-mahjongg"
	SnapdException: SnapdException(kind: login-required, message: access denied)
```

Otherwise a few comments scattered throughout.

d-loose:
> I'm getting an error when trying to install a snap. Are you experiencing the same?
> 
> ```
> flutter: ERROR snap: Caught exception for snap "gnome-mahjongg"
> 	SnapdException: SnapdException(kind: login-required, message: access denied)
> ```
> 
> Otherwise a few comments scattered throughout.

No, I don't. Are you logged into snapd? If not, clicking 'install' should invoke polkit and ask for your password.

lib/src/widgets/dialogs.dart
```diff
@@ -0,0 +1,80 @@
+import 'package:flutter/material.dart';
```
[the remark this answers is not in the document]
d-loose: I'm anticipating adding additional dialog types (e.g. confirmation before removing an app). Should have left a comment, though :)

lib/src/snapd/updates_model.dart
d-loose: I was following the pattern used in ubuntu-desktop-provision, but I'm open to trying out a different approach, as this can potentially be a little error-prone if we forget to explicitly hide `log` from some exports

tim-hm: You have more experience with this codebase so you tell me how you want to do it!

tim-hm:
Ah the error is caused because I'm using vscodes remote development to run the app so I never get the polkit prompt. Works as expected when I run directly from my desktop.

lib/src/manage/manage_page.dart
```diff
@@ -13,15 +15,43 @@ import '/widgets.dart';
...
+  @override
+  void dispose() {
+    _errorSubscription?.cancel();
+    _errorSubscription = null;
+    super.dispose();
+  }
+
+  Future<void> showError(SnapdException e) => showErrorDialog(
+        context: context,
+        title: e.kind ?? '',
```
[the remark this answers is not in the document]
d-loose: I'm hardcoding those for now, since we'll address showing proper error messages in #1362

lib/src/snapd/updates_model.dart
d-loose: I've added a 'logger.dart' in lib/src/snapd, which can be imported within the directory, but is not listed in the exports in lib/snapd.dart. Will do the same for the other directories as we go.

tim-hm (approved):
Apologies I meant to say go ahead and merge when you're happy!

PR #1409, opened 2023-09-26
feat(appstream): keep original search tokens

### LINKED ISSUE #1305: Geary isn't showing up in the app-store
When searched for `geary` it's not showing up in the searches of the app-store. 

image

But, it's indeed in the Ubuntu repo.

https://packages.ubuntu.com/lunar/geary

The app-store version is 0.3.0-alpha and I'm using Ubuntu 23.04

d-loose:
UDENG-1440
Fix #1305

The stemming algorithm used in the appstream search function tends to turn 'y' into 'i' at the end of the word for some English words (can be tested [here](https://snowballstem.org/demo.html)).
I'm not sure what the best solution is, but simply keeping the original search tokens along with the ones generated by the stemmer seems to work well.
I couldn't reproduce the issue with `appstreamcli` so I had a brief look at appstream's implementation of the search logic, but didn't spot any obvious differences.

### DIFF
--- a/lib/src/appstream/appstream_service.dart
+++ b/lib/src/appstream/appstream_service.dart
@@ -226,7 +226,13 @@ class AppstreamService {
         stemmersMap[PlatformDispatcher.instance.locale.languageCode];
     if (algorithm != null) {
       final stemmer = SnowballStemmer(algorithm);
-      return words.map((element) => stemmer.stem(element)).toSet().toList();
+      return words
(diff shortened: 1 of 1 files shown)


tim-hm (approved):
LGTM

I didn't conduct an exhaustive test by any stretch of the imagination, but I did find with this PR my search results were a bit more as I would expect. We'll probably want to revisit this in due course especially as I don't know what the implications for non-english searches might be.

PR #1412, opened 2023-09-26
feat: filter snap search query

### LINKED ISSUE #863: Search with special characters gives bad query

d-loose:
UDENG-1484
Fix #863

### DIFF
--- a/lib/src/snapd/snap_search.dart
+++ b/lib/src/snapd/snap_search.dart
@@ -10,6 +10,7 @@ class SnapSearchParameters {
 
   // Removes characters from the search query that would cause snapd to reply
   // with 'bad query'.
+  // See https://github.com/snapcore/snapd/blob/29c9752d66bf95ffa85d85309531e2f5c7971553/store/store.go#L1048-L1056
   String? get cleanedQuery => query?.replaceAll(
         RegExp(r'[+=&|><!(){}\[\]^"~*?:\\/]'),
         ' ',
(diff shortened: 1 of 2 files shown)


lib/src/snapd/snap_search.dart
```diff
@@ -8,6 +8,13 @@ class SnapSearchParameters {
   final String? query;
   final SnapCategoryEnum? category;
 
+  // Removes characters from the search query that would cause snapd to reply
+  // with 'bad query'.
```
tim-hm, line 12: Any chance we have a reference to this in the snapd docs?

test/snap_search_test.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:app_center/snapd.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  test('cleaned query', () {
+    const searchParamters = SnapSearchParameters(
+      query: 'a+b=c&d|e>f<g!h(i)j{k}l[m]n^o"p~q*r?s:t\\u/v',
+    );
+    expect(searchParamters.cleanedQuery,
+        equals('a b c d e f g h i j k l m n o p q r s t u v'));
```
tim-hm, line 10: Neat!

tim-hm (approved):
LGTM. Just one comment inline for your consideration.

lib/src/snapd/snap_search.dart
d-loose: Haven't found anything in the docs on this, but added the source code reference

Issue #1414, opened 2023-09-27
Update button not working properly


soumyaDghosh:
[Screencast from 2023-09-27 12-32-34.webm](https://github.com/ubuntu/app-center/assets/72045785/1d77af21-8064-4739-92f9-7eb8dfd80de2)

This is linked to #1410. Here as firefox is running, I cannot update others too. 

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: SnapdException(kind: null, message: cannot refresh "firefox", "bavarder", "icon-library": snap "firefox" has running apps (firefox), pids: 6014,6753,6799,6942,7131,14756,14759,32823,32956,35550,97814,99845,110217,116155,117610,117776,117817)
#0      _SnapdErrorResponse.result (package:snapd/src/snapd_client.dart:1164)
#1      SnapdClient._postAsync (package:snapd/src/snapd_client.dart:1564)
<asynchronous suspension>
#2      SnapdClient.refreshMany (package:snapd/src/snapd_client.dart:1315)
<asynchronous suspension>
#3      UpdatesModel.updateAll (package:app_center/src/snapd/updates_model.dart:39)
<asynchronous suspension>
```

d-loose:
Thanks for the bug report!

#1396 should take care of that as well. I'm currently working on #1406 which might be related. I'll leave this open until the changes for this are merged.

d-loose:
Just had a chance to test this with real-world updates:

![image](https://github.com/ubuntu/app-center/assets/113362648/497f009f-57c4-4a38-8e84-f1e84fc83a5e)

While the error message isn't very beautiful yet (#1362) it does the job for now.

soumyaDghosh:
looks good.

PR #1417, opened 2023-09-27
fix(manage): update all button

### LINKED ISSUE #1406: 'Update all' button remains enabled during update (manage page)
It also doesn't correctly display the "Updating" state.

d-loose:
UDENG-1487
Fixes #1406, #1414
Also prevents division by zero in the progress calculation

### DIFF
--- a/test/snap_model_test.dart
+++ b/test/snap_model_test.dart
@@ -232,4 +232,30 @@ void main() {
     );
     await model.install();
   });
+  group('change progress', () {
+    final testCases = [
+      (
+        name: 'no tasks',
+        change: SnapdChange(spawnTime: DateTime(1970)),
+        expectedProgress: 0.0,
+      ),
+      (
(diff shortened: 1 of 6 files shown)


test/snap_model_test.dart
```diff
@@ -232,4 +232,30 @@ void main() {
     );
     await model.install();
   });
+  group('change progress', () {
```
d-loose, line 235: I'm experimenting with golang-style table driven unit tests here. Unlike Flutter's `testWidgets()`, Dart's `test()` doesn't offer [variants](https://api.flutter.dev/flutter/flutter_test/TestVariant-class.html). Calling `test()` in a loop like this works fine, but might confuse the editor/IDE.

tim-hm (approved):
I've struggled to test the update button locally, but looking through the code it looks good to me!

PR #1421, opened 2023-09-29
feat: UI tweaks

d-loose:
@anasereijo, as discussed:

* Adjusted margins/paddings on the snap page
* Removed bottom margins
* Added the summary to the description section
* Changed labels to sentence case

We'll need to come back to color and text theme tweaks later on - sorry!

Screenshot from 2023-09-29 16-30-36
Screenshot from 2023-09-29 16-30-40

### DIFF
--- a/lib/src/explore/explore_page.dart
+++ b/lib/src/explore/explore_page.dart
@@ -23,64 +23,65 @@ class ExplorePage extends ConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final l10n = AppLocalizations.of(context);
-    return Padding(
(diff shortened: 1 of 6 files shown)


anasereijo:
Tiny detail: These margins/paddings seem to be quite big, but could be an optical illusion. I have 48px on my designs. Everything else is looking great

Screenshot 2023-09-29 at 15 38 59
Screenshot 2023-09-29 at 15 39 19

d-loose:
> Tiny detail: These margins/paddings seem to be quite big, but could be an optical illusion. I have 48px on my designs. Everything else is looking great

Ah, good catch - the titles of the expandable have an 8px padding ..

anasereijo (approved):

PR #1436, opened 2023-10-04
docs: update README and add CONTRIBUTING.md

d-loose:
I've used the [templates](https://github.com/canonical/desktop-engineering/tree/main/project-repo/templates/common) as a starting point, but left out some things regarding bug reporting as there are still some things to figure out:

* Do we plan to track any issues on launchpad and use `ubuntu-bug`?
* Based on the previous question, we should also create different templates for bug reports and feature requests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,120 @@
+# Contributing to App Center
+
+A big welcome and thank you for considering contributing to App Center and Ubuntu! It’s people like you that make it a reality for users in our community.
+
+Reading and following these guidelines will help us make the contribution process easy and effective for everyone involved. It also communicates that you agree to respect the time of the developers managing and developing this project. In return, we will reciprocate that respect by addressing your issue, assessing changes, and helping you finalize your pull requests.
+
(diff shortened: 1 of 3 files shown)


tim-hm:
@matthew-hagemann can i ask that you review this? Sorry gonna be swamped until after release so don't want to hold these things up.

CONTRIBUTING.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Translations
+
+Translations are managed using [Weblate](https://hosted.weblate.org/projects/ubuntu-software/app-center/)
+
+## Contributing to the code
+
+### Required dependencies
+
+[Install Flutter](https://flutter.dev/docs/get-started/install/linux) (Note: Currently version 3.10.6 is required, which can be installed most easily by following the manual installation instructions).
```
matthew-hagemann, line 73: I know a few of us find it easier to install Flutter via [fvm](https://fvm.app/) or [asdf](https://asdf-vm.com/) to move between versions and pin to the one App Center is on. Is that perhaps a suggestion we want to make to the community as well?

matthew-hagemann (approved):

PR #1443, opened 2023-10-06
fix(updates_model): handle error on refresh

### LINKED ISSUE #1429: Check for Updates button suddenly gone while the operation was done and I was scrolling the apps
image

d-loose:
Turns out the `refresh()` method wasn't handling snapd errors.
There will be an error dialog instead of an empty button now, in case snapd can't get the list of pending updates.

Fix #1429


matthew-hagemann (approved):
LGTM :rocket:

PR #1474, opened 2023-10-26
fix(snap-page): fix channel dropdown item height

### LINKED ISSUE #1472: bug: text overflows in channel dropdown
Screenshot of Ubuntu App Centre's channel selection dropdown

I think the title and the screenshot are pretty self-explanatory. Looks like there isn't enough height for the text to fit properly.

The app inside the screenshot is Darktable.

d-loose:
Fix #1472


matthew-hagemann (approved):
Looks good! :rocket:

PR #1486, opened 2023-11-08
fix(packagekit): check arch compatibility in resolve

### LINKED ISSUE #1485: If multiarch is enabled, deb packages for foreign architectures are installed without a warning
When resolving a package name with PackageKit and the same package is available for multiple architectures, `PackageService` always picks the last one and doesn't check whether the packages architecture matches the native one.

e.g.:
```
$ dpkg --print-architecture 
amd64
$ dpkg --print-foreign-architectures 
i386
$ pkcon resolve inkscape
Resolving                               [=========================]         
Loading cache                           [=========================]         
Available   	inkscape-1.2.2-2ubuntu4.amd64 (ubuntu-mantic-universe)      	vector-based drawing program
Available   	inkscape-1.2.2-2ubuntu4.i386 (ubuntu-mantic-universe)       	vector-based drawing program
```

Here the package for i386 is selected instead of the native amd64 package. 

It subsequently installs the package without any user prompts using PackageKit's DBus API (which doesn't seem to provide a warning). This can lead to a lot of important packages being removed from the system, as described in [this](https://old.reddit.com/r/Ubuntu/comments/17k1rz6/installing_inkscape_deb_using_app_center_just/) reddit thread.

d-loose:
Fix #1485
- reads the local architecture from the snap environment (or from the output of `dpkg --print-architecture`)
- makes sure `resolve()` only returns packages with matching architecture

Adding proper support for foreign architectures requires a bit more care and some UI/UX input - I'd leave that as a separate issue.

### DIFF
--- a/lib/src/packagekit/packagekit_service.dart
+++ b/lib/src/packagekit/packagekit_service.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:io';
 
 import 'package:dbus/dbus.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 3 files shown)


matthew-hagemann (approved):
LGTM :rocket:

PR #1500, opened 2023-11-20
refactor: Move to monorepo file structure

spydon:
When using melos it preferably needs its own `pubspec.yaml` file in the root, so even though there is only one package/app here it's good to move to a monorepo structure.

We could also put it in `apps/` instead of in `packages/`.

Also moves the CI to be consistent with the rest of the repositories.


d-loose (approved):
Thanks a lot, LGTM! :+1:

PR #1505, opened 2023-11-24
chore: Move ratings client to monorepo

spydon:


matthew-hagemann:
@spydon Looks good! I have made some [changes](https://github.com/canonical/app_center_ratings_client.dart/pull/9) to the client that need to be reviewed and then included for the chart endpoints :slightly_smiling_face:

packages/app_center_ratings_client/pubspec.yaml
```diff
@@ -0,0 +1,22 @@
+name: app_center_ratings_client
```
matthew-hagemann, line 1: I see you renamed the main dart file to `ratings_client.dart` . Should we not then update the name in the `pubspec.yaml` to reflect that?

matthew-hagemann (approved):
LGTM :guitar: just one small question on naming.

packages/app_center_ratings_client/pubspec.yaml
spydon: It probably should be a barrel file though, I'll update the PR to that.

spydon: Haha, we're back to `app_center_ratings_client.dart`, because otherwise it will have the same name as the now new `src/ratings_client.dart`. Good talk. :joy:

matthew-hagemann: Haha, I feel left out now :joy:

matthew-hagemann: I have it in the back of my head I renamed it to match the name in the pubspec because of a rule re: uploading to pub.dev? I could be mistaken

PR #1508, opened 2023-11-28
fix: ErrorWidget breaking the about page UI.

### LINKED ISSUE #1506: Having GitHub blocked by your current network breaks the About page
The image should say all:
image

...also the resulting grey block is absurdly tall - probably about 20+ times the height of that window.

BLKKKBVSIK:
In this PR:

- Refactor an `ErrorWidget` to a `Text` widget when the app-center can't communicate with Github; the `ErrorWidget` was breaking the UIs with no possibility to see the rest of the content after the contributors.

fix #1506 

Before:
Screenshot from 2023-11-28 11-41-52

After:
Screenshot from 2023-11-28 11-41-40

### DIFF
--- a/packages/app_center/lib/src/about/about_page.dart
+++ b/packages/app_center/lib/src/about/about_page.dart
@@ -121,7 +121,7 @@ class _ContributorView extends ConsumerWidget {
         const SizedBox(height: 8),
         state.when(
           data: (contributors) => _ContributorWrap(contributors),
-          error: (error, stackTrace) => ErrorWidget(error),
+          error: (error, stackTrace) => Text(error.toString()),
           loading: () => Shimmer.fromColors(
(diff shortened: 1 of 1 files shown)


spydon (approved):

BLKKKBVSIK:
> Maybe we could make this a bit clearer to the user that it is an error?
> 
> And the user shouldn't need to see the content of the error, just get an error string like "Can't fetch content right now, try again later." or something like that.

Definitely, but again we have the problem of l10n.

spydon:
> Definitely, but again we have the problem of l10n.

Aah, there is not even a generic error string in the i10n...

spydon (approved):
Approved, since any other solution is blocked by the localization blocker.

PR #1511, opened 2023-11-29
fix: plural spelling issue for l10n

### LINKED ISSUE #1451: 1 day, 2 days in Manage list
Hi,

Not a big deal !!

image

BLKKKBVSIK:
In this PR:

- Adding plural for l10n.

fix #1451 

Ex:
Now:
Screenshot from 2023-11-29 10-43-57
Before:
Screenshot from 2023-11-29 10-44-19

### DIFF
--- a/packages/app_center/lib/src/l10n/app_en.arb
+++ b/packages/app_center/lib/src/l10n/app_en.arb
@@ -32,7 +32,7 @@
     "managePageShowDetailsLabel": "Show details",
     "managePageShowSystemSnapsLabel": "Show system snaps",
     "managePageUpdateAllLabel": "Update all",
-    "managePageUpdatedDaysAgo": "Updated {n} days ago",
+    "managePageUpdatedDaysAgo": "Updated {n, plural, =0{{n} day} =1{{n} day} other{{n} days}} ago",
     "@managePageUpdatedDaysAgo": {
         "placeholders": {
             "n": {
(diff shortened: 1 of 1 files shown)


spydon (changes requested):
This is actually wrong, nouns following zero should (confusingly) be treated as plural.
https://english.stackexchange.com/questions/38293/why-is-zero-followed-by-a-plural-noun

spydon (approved):

fthx:
That's not the case in French.
We write:
0 jour
1 jour
2 jours
...

spydon:
> That's not the case in French. We write: 0 jour 1 jour 2 jours ...

Google translate tells me that the current translation is correct for 0 in French already, but I don't speak French:
```
"managePageUpdatedDaysAgo": "Mis à jour il y a {n} jours",
```

https://github.com/ubuntu/app-center/blob/main/packages/app_center/lib/src/l10n/app_fr.arb#L164C32-L164C59

fthx:
No this translation is not correct.
That's the point of this BR.

spydon:
Aah now I get it #1451 was supposed to be for French? Didn't notice since that sentence was still in English on your screenshot.

fthx:
The FR one is wrong for 0 too.
https://dictionnaire.lerobert.com/guide/accord-du-nom-apres-zero

In FR we write:
« I made no errors in my code. »
(because usually there are many)
*but*
« I made zero error in my code.»

PR #1519, opened 2023-12-07
feat(snap): use melos in snapcraft.yaml

d-loose:
This ensures all local package dependencies are correctly set up before building the binary.
Should resolve the recent build failures on launchpad :crossed_fingers:


d-loose:
Works! https://launchpad.net/~desktop-snappers/+snap/app-center/+build/2325803

PR #1523, opened 2023-12-12
fix: We shouldn't assume that the package info will be not null or valid

### LINKED ISSUE #1521: Void debian package window
Hi,

- search for "pychess" in search bar
image

- click on PyChess
image

- problem ;-)
image

BLKKKBVSIK:
In this PR:

- Fixing the UI when accessing a deb package with no package info, it was rendering an `ErrorWidget`.

fix #1521 
Before:
image

After:
image


packages/app_center/lib/src/deb/deb_model.dart
```diff
@@ -76,11 +76,13 @@ class DebModel extends ChangeNotifier {
 
   Future<void> install() {
     assert(packageInfo != null);
+    //TODO: Display a snackbar if packageInfo is null
     return _packageKitAction(() => packageKit.install(packageInfo!.packageId));
   }
```
BLKKKBVSIK, line 81: Going to update this PR to add `Snackbar` when clicking on the install button and remove button because they will not work.

d-loose (dismissed):
This is masking the underlying issue, caused by [this](https://github.com/ubuntu/app-center/pull/1486). We should instead fix `PackageKitService.resolve()` such that the architecture 'all' is always accepted - currently it's checking for an exact match in architectures, but for pychess the package information looks like this:


```
$ pkcon resolve pychess
Resolving                               [=========================]         
Loading cache                           [=========================]         
Available   	pychess-1.0.4-0.1.all (ubuntu-mantic-universe)              	Fully featured, nice looking, easy to use Chess Client
```

BLKKKBVSIK:
I will investigate `PackageKitService.resolve()`, but I think the previous fix should stay anyway, the UI shouldn't break in any case.

d-loose:
> I will investigate `PackageKitService.resolve()`, but I think the previous fix should stay anyway, the UI shouldn't break in any case.

In principle PackageKit should always be able to resolve the package for a given appstream component. But I agree that we should handle this case more gracefully if something does go wrong. However, in this case we'd also need to disable all action buttons, since we won't be able to install/remove/update the package in that case. Maybe we could show an error message instead of the button on that case? (Probably something for @anasereijo to decide). We should also write some error messages to the log here.

packages/app_center/lib/src/packagekit/packagekit_service.dart
```diff
@@ -146,12 +146,16 @@ class PackageKitService {
     String name, [
     @visibleForTesting String? architecture,
   ]) async {
-    final arch = architecture ?? await _getNativeArchitecture();
+    final possibleArchs = [
+      architecture ?? await _getNativeArchitecture(),
+      'all'
```
d-loose, line 151: could we get a test case for 'all'? :)

d-loose:
What about disabling the action buttons and showing an error, could we do that in this PR as well? Otherwise this will create another 'broken' UI state in case an error occurs.

anasereijo:
Let's do that: disable the buttons and have a tooltip telling the user what is happening.

anasereijo:
what sort of error is this so I can help with the copy?

d-loose:
> what sort of error is this so I can help with the copy?

It's a somewhat technical issue that happens when PackageKit (the service responsible for installing/removing/updating packages) can't find a corresponding package for a given AppStream component (AppStream is the service that provides the metadata for the packages and allows us to perform searches).
In this case it was our fault, because we excluded a valid package by being overly cautious with respect to the architecture.
Since this problem likely indicates a problem on our end rather than the user's, it's ok (and probably better) if it's fairly technical.

@BLKKKBVSIK we should definitely log an error if `resolve()` doesn't find a package.

spydon (approved):
Lgtm

PR #1534, opened 2023-12-13
Show "App Center" in Gnome Shell activities view

### LINKED ISSUE #1425: The app name is App Center but shows snap-store in overview
image

Currently I am on Ubuntu 23.04. Is this name intentional?

sergio-costas:
Currently, the window in the Activities view in Gnome Shell has "snap-store" as its name, which is incorrect because this is the brand-new "App Center". The reason is that the Activities view shows the window title, but the app center doesn't set it, so it is set to the binary name.

The patch uses the translations from the .desktop file, thus it doesn't require extra effort to ensure that the name shown in the Activities view is correctly translated.

fix #1425

### DIFF
--- a/packages/app_center/lib/src/l10n/app_en.arb
+++ b/packages/app_center/lib/src/l10n/app_en.arb
@@ -1,4 +1,5 @@
 {
+    "appCenterLabel": "App Center",
     "appstreamSearchGreylist": "app;application;package;program;programme;suite;tool",
     "snapPageChannelLabel": "Channel",
     "snapPageConfinementLabel": "Confinement",
--- a/packages/app_center/lib/src/store/store_app.dart
+++ b/packages/app_center/lib/src/store/store_app.dart
@@ -43,55 +43,72 @@ class _StoreAppState extends ConsumerState<StoreApp> {
         localizationsDelegates: localizationsDelegates,
         navigatorKey: ref.watch(materialAppNavigatorKeyProvider),
         supportedLocales: supportedLocales,
-        home: Scaffold(
-          appBar: YaruWindowTitleBar(
-            title: ConstrainedBox(
-              constraints: const BoxConstraints(maxWidth: kSearchBarWidth),
-              child: SearchField(
-                onSearch: (query) =>
-                    _navigator.pushAndRemoveSearch(query: query),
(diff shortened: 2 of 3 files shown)


packages/app_center/linux/my_application.cc
```diff
@@ -33,6 +33,7 @@ static void my_application_activate(GApplication* application) {
 
   GtkWindow* window = GTK_WINDOW(hdy_application_window_new());
   gtk_window_set_application(window, GTK_APPLICATION(application));
+  gtk_window_set_title(window, "App Center");
```
AsciiWolf, line 36: This should ideally be marked as translatable. "App Center" is translated in many languages.

sergio-costas: Yes, but the problem is that it is already translated in the .desktop file, so I think that we should try to get the translation from there. What do you think?

AsciiWolf: That would be ideal, if it is possible.

sergio-costas: I'll try to use Gio.DesktopAppInfo.

sergio-costas: Done!

AsciiWolf:
Looks good to me, however I am not an app-center developer. :-)

sergio-costas:
Agh! :-D

d-loose (approved):
Thanks @sergio-costas LGTM :+1:

I've pinned the mockito version, like we do in the provision monorepo.

sergio-costas:
@d-loose Thanks!

AsciiWolf (commented):
LGTM

d-loose:
In hindsight, it might be cleaner to do this in Flutter instead of further customizing the GTK application. With [this](https://github.com/ubuntu/yaru_window.dart/pull/42) change, we could change the title from within Flutter using `YaruWindow.of(context).setTitle(..)` and simply add 'App Center' to the l10n strings. What do you think?

d-loose:
Just published yaru_window 0.2.0 with @Feichtmeier's help :)
@BLKKKBVSIK, if you find some time to implement a Flutter-based version of this fixed, as explained above, that'd be awesome!
I'm off for the remainder of the year now :)
Happy holidays everyone!

spydon (approved):

PR #1544, opened 2023-12-19
fix: updating the background color of the content pages

### LINKED ISSUE #1529: content section white
Would be nice to make the 'content sections' of the app white instead of grey. 
See some screenshots (or refer to figma):
Screenshot 2023-12-13 at 16 17 21
Screenshot 2023-12-13 at 16 16 54

I think this will be a hack to the default yaru, but we can propose having that option too?

BLKKKBVSIK:
In this PR:

- Updating the background color of the content pages in the light theme.

fix #1529

### DIFF
--- a/packages/app_center/lib/src/store/store_app.dart
+++ b/packages/app_center/lib/src/store/store_app.dart
@@ -37,7 +37,7 @@ class _StoreAppState extends ConsumerState<StoreApp> {
 
     return YaruTheme(
       builder: (context, yaru, child) => MaterialApp(
-        theme: yaru.theme,
+        theme: yaru.theme?.copyWith(scaffoldBackgroundColor: Colors.white),
         darkTheme: yaru.darkTheme,
         debugShowCheckedModeBanner: false,
         localizationsDelegates: localizationsDelegates,


packages/app_center/lib/src/store/store_app.dart
```diff
@@ -37,7 +37,7 @@ class _StoreAppState extends ConsumerState<StoreApp> {
 
     return YaruTheme(
       builder: (context, yaru, child) => MaterialApp(
-        theme: yaru.theme,
+        theme: yaru.theme?.copyWith(scaffoldBackgroundColor: Colors.white),
```
spydon, line 40: Should this not be changed in the yaru theme?

spydon:
Updated to `fix` since this isn't a new feature

packages/app_center/lib/src/store/store_app.dart
BLKKKBVSIK: I thought it was only for this app. If we want to update the color on white theme for the whole theme, I can update the PR.

spydon: Isn't it strange to not have it consistent?

BLKKKBVSIK: You're probably correct, gonna update my PR.

spydon: We could leave it open until Ana is back so that she can reply

Feichtmeier: Hello 👋  😄 
Sure, Colors.white sounds good for yaruLight Scaffold background :) ! 
Surface could stay a bit darker to have something to make contrast against if this is needed
best regards

BLKKKBVSIK:
Closing this PR since it will be fixed by ubuntu/yaru.dart#405

PR #1536, opened 2023-12-16
fix: Allow cancelling updates in manage page

codinesh:
Fixes #1533 by adding a Cancel button next to 'Update All' in manage page.

Cancel button is disabled by default, when no updates are not initialized.
image

Once update all is triggered and in progress, Cancel button will be enabled, and will stop the updates on click.
image

Hello @anasereijo, I couldn't find the link to Figma, to make the button width match with designs. Will be happy to take that change, please help me with Figma link and access.

PR #1557, opened 2024-01-03
fix: Manage page: Hide cancel button when updates are not in progress.

codinesh:
> One thing to amend is that the cancel button should be hidden until the 'Update all' is trigged, as it can be misleading for the users to have a cancel button without context.

As per the comments form #1536, this PR will hide cancel button when updates are not in progress.

### DIFF
--- a/packages/app_center/lib/src/manage/manage_page.dart
+++ b/packages/app_center/lib/src/manage/manage_page.dart
@@ -238,8 +238,9 @@ class _ActionButtons extends ConsumerWidget {
     final updatesInprogress = updatesModel.refreshableSnapNames.isNotEmpty &&
         !updatesModel.state.isLoading &&
         updatesModel.activeChangeId != null;
-    return Row(
-      mainAxisSize: MainAxisSize.min,
+    return Wrap(
+      spacing: 10,
+      runSpacing: 10,
       children: [
         PushButton.outlined(
           onPressed:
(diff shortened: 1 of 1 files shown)


soumyaDghosh:
@anasereijo Is it possible to add that cancel button with singular updates also? like the `update` button will change to `cancel` button? @codinesh will it be easy/viable?

codinesh:
Yes @soumyaDghosh, it would be easy to change text, but we loos the ability to indicate updates in progress as we show loading indicator on the Update all button

Feichtmeier:
The cancel button overflows
grafik

you could use a wrap instead of a row to prevent this

grafik

return Wrap(
      spacing: 10,
      runSpacing: 10,

but then remove the sizedboxes between

codinesh:
Hello @Feichtmeier, thanks for suggestion to use Wrap. I have made the changes in latest commit, can you please check when feasible

BLKKKBVSIK:
LGTM !

@Feichtmeier your reviews are always welcome tho! 💙

BLKKKBVSIK (approved):

soumyaDghosh:
The update button looks wiered here

Screenshot from 2024-01-19 21-12-57


@BLKKKBVSIK @anasereijo

codinesh:
> The update button looks wiered here
> 
> Screenshot from 2024-01-19 21-12-57
> 
> @BLKKKBVSIK @anasereijo

Hi @soumyaDghosh, thanks for reporting, I will work on the fix for this.

soumyaDghosh:
@codinesh thanks.. But I have already fixed it 😁.. But thanks again for all of your work on the store.

PR #1561, opened 2024-01-07
snap_page: use mouse hand instead of pointer for voting icons

soumyaDghosh:
### before: 

Screenshot from 2024-01-07 15-00-22

### after:

Screenshot from 2024-01-07 15-18-25


BLKKKBVSIK (approved):
LGTM

PR #1575, opened 2024-01-18
tests: regenerate snapd mocks

jssotomdz:
`package:snapd/snapd.dart` added more options to the find method
Regenerating mocks added these lines
Is this the right way to regenerate mocks on the main branch? Via PR?


d-loose (approved):
In this case it's the best way :+1: (but it doesn't usually happen that mocks need to be regenerated due to a new minor release of an external dependency).

PR #1568, opened 2024-01-10
fix(manage page): button content alignment is not consistent

shreyasSarve:
fix: Manage page: button content aligment is not consistent
Screenshot from 2024-01-10 17-12-27

fix #1558

PR #1576, opened 2024-01-18
Fix #1558

### LINKED ISSUE #1558: Manage page: button content aligment is not consistent
grafik

In the top control panel with the "update all button" the button height is correct and thus the content padding matches and thus the content is positioned in the middle.
However in those containers with the border, it looks like the button height is a little bit squeezed and thus the content padding is too much for the content and thus the content is positioned a little bit too much down
May be an optical illusion but I am almost certain :crossed_fingers:

shreyasSarve:
I have reponed new pr previous one was failling some tests

### DIFF
--- a/packages/app_center/test/snapd_cache_test.mocks.dart
+++ b/packages/app_center/test/snapd_cache_test.mocks.dart
@@ -317,6 +317,7 @@ class MockSnapdClient extends _i1.Mock implements _i2.SnapdClient {
     String? category,
     String? section,
     _i2.SnapFindFilter? filter,
+    _i2.SnapFindScope? scope,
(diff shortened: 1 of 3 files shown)


Feichtmeier:
Hey @shreyasSarve 
Are you sure this fixes the issue?
In your previous, canceled, pull request the diff looks very much like it might fix the issue
https://github.com/ubuntu/app-center/pull/1568/files
However this pull request only changes mocks
https://github.com/ubuntu/app-center/pull/1576/files

I think you need to combine both changes, even though the mocks might be outdated again
wouldntit be eitherif you re-open the oldpull request ?
Thanks and best regards

Issue #1563, opened 2024-01-08
Enhance the Games Tab


ashuntu:
Right now the Games tab in the App Center is just a procedural page based on app category, but it could serve as a useful hub for gaming related apps, information, and tools.

@jssotomdz, @ZoopOTheGoop, and I would like to make a dedicated Games tab that provides more information than just lists of apps. Our initial ideas include a featured carousel of games and apps (think [Steam's carousel of games](https://store.steampowered.com/)), "packs" that include apps and games that often are used together (like Steam + Discord + OBS + etc.), and more robust headings similar to the Explore page like Top 10 apps/games, "Hot" apps/games, etc.

Additionally, we could query what games/apps someone already has installed, and recommend others. Like if you have Steam installed but not Discord, etc.

Gaming on Linux can be daunting for newcomers, so, approval permitting, we'd also like to include links to external resources that may not be controlled by Canonical or Ubuntu. This could include helpful resources like [ProtonDB](https://www.protondb.com/), [Lutris](https://lutris.net/), [Heroic](https://heroicgameslauncher.com/), [Wine](https://www.winehq.org/), etc.

These are all goals that could take suggestions or criticism, so feel free to comment thoughts.

soumyaDghosh:
The games tab can be made like a completely separate page, and so does the other tabs, IMO. Like eg: play-store in android, it shows big screenshots of games, shows ongoing events, like an event going on epic-store right now, can be shown there, which can they be played by using heroic, which is currently not yet available properly as a snap, but will be soon.

michaldev:
> The games tab can be made like a completely separate page, and so does the other tabs, IMO. Like eg: play-store in android, it shows big screenshots of games, shows ongoing events, like an event going on epic-store right now, can be shown there, which can they be played by using heroic, which is currently not yet available properly as a snap, but will be soon.

In my opinion, it is too complicated and requires too much from developers. Many package developers have no control over the game code at all.

The thread author's proposal is more reasonable. A list of useful tools and a description of what they can be used for will be a hit for people who are not familiar with the Linux application ecosystem.

PR #1598, opened 2024-02-12
feat: Games tab overhaul

ashuntu:
This PR overhauls the games tab of the app center. More work is needed, so this is still a draft.

Closes https://github.com/ubuntu/app-center/issues/1563

Here's what the tab looks like right now as you scroll down it:

Carousel and featured games
All games
External resources and app bundles

Example app bundle page:

image

External resources page:

image


michaldev:
"develop your own game" should be in the "development" tab in my opinion. The "games" tab should be intended for players.

packages/app_center/lib/src/widgets/app_card.dart
```diff
@@ -46,6 +47,23 @@ class AppCard extends StatelessWidget {
         onTap: onTap,
       );
 
+  factory AppCard.fromTool({
+    required Tool tool,
+    required AppLocalizations l10n,
```
d-loose, line 52: Remove `l10n` argument and get it from the build context instead

packages/app_center/lib/src/widgets/app_card.dart
```diff
@@ -46,6 +47,23 @@ class AppCard extends StatelessWidget {
...
+  }) =>
+      AppCard(
+        title: AppTitle.fromTool(tool.name, tool.publisher),
+        summary: tool.summary,
+        iconUrl: tool.iconUrl,
+        footer: OutlinedButton(
+          onPressed: () async {
+            await _launchUrl(Uri.parse(tool.url));
+          },
+          child: Text(l10n.openInBrowser),
```
d-loose, line 63: You can wrap the `Text` widget in a `Builder` to get a build context here

packages/app_center/lib/src/widgets/app_title.dart
```diff
@@ -37,6 +37,17 @@ class AppTitle extends StatelessWidget {
         large: large,
       );
 
+  factory AppTitle.fromTool(
```
d-loose, line 40: The name suggests a signature like this:
```dart
factory AppTitle.fromTool(Tool tool, {...})
```

packages/app_center/lib/src/widgets/banner.dart
```diff
@@ -0,0 +1,338 @@
...
+    this.kIconSize = 48.0,
+    this.kNumberOfBannerSnaps = 3,
+    super.key,
+  });
+
+  final SnapCategoryEnum category;
+  final double padding;
+  final double height;
+  final double kMaxSize;
+  final double kIconSize;
```
d-loose, line 27: Please use appropriate types, i.e. `EdgeInsets` for paddigns, `Size` for sizes etc

packages/app_center/lib/src/widgets/banner.dart
```diff
@@ -0,0 +1,338 @@
...
+  });
+
+  final Iterable<Snap> snaps;
+  final String slogan;
+  final String? buttonLabel;
+  final VoidCallback? onPressed;
+  final double padding;
+  final double height;
+  final double kMaxSize;
+  final double kIconSize;
```
d-loose, line 83: Same as above, please use proper types :)

packages/app_center/lib/src/widgets/banner.dart
```diff
@@ -0,0 +1,338 @@
...
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return Container(
+      padding: const EdgeInsets.all(48),
+      decoration: BoxDecoration(
+        borderRadius: BorderRadius.circular(8),
+        gradient: const LinearGradient(
+          colors: [
+            Color.fromARGB(255, 29, 27, 112),
+            Color.fromARGB(255, 49, 1, 82),
```
d-loose, line 238: Could we move this to the list of banner colors instead of hardcoding it here?

packages/app_center/lib/src/widgets/app_card.dart
```diff
@@ -201,3 +219,17 @@ class _RatingsInfo extends ConsumerWidget {
...
+
+class Tool {
+  Tool(this.name, this.summary, this.publisher, this.iconUrl, this.url);
+
+  final String name;
+  final String summary;
+  final String publisher;
+  final String iconUrl;
+  final String url;
+}
```
d-loose, line 231: This data class is unrelated to the `AppCard` widget and should be moved elsewhere. Also consider using [freezed](https://pub.dev/packages/freezed). The class should at least be immutable and have a `const` constructor.

packages/app_center/lib/src/widgets/app_card.dart
```diff
@@ -46,6 +47,23 @@ class AppCard extends StatelessWidget {
         onTap: onTap,
       );
 
+  factory AppCard.fromTool({
+    required Tool tool,
+    required AppLocalizations l10n,
+    VoidCallback? onTap,
```
d-loose, line 53: `onTap` isn't used

packages/app_center/lib/src/widgets/banner.dart
```diff
@@ -0,0 +1,338 @@
...
+class _BannerIcon extends StatefulWidget {
+  const _BannerIcon({
+    required this.snap,
+    required this.kMaxSize,
+    required this.kIconSize,
+  });
+
+  final Snap snap;
+  final double kMaxSize;
+  final double kIconSize;
```
d-loose, line 158: Types :)

d-loose: Also the `k` naming convention indicates constants, which those are not

packages/app_center/lib/src/widgets/banner.dart
d-loose: And don't use the `k`-prefix for non constant symbols

packages/app_center/lib/src/widgets/snap_grid.dart
```diff
@@ -36,6 +37,19 @@ class AppCardGrid extends StatelessWidget {
         ),
       );
 
+  factory AppCardGrid.fromTools({
+    required List<Tool> tools,
+    required AppLocalizations l10n,
```
d-loose, line 42: See somewhere above - you shouldn't need to pass the `l10n` argument all the way down, since the localizations can be retrieved from the build context

packages/app_center/lib/src/search/search_page.dart
```diff
@@ -149,6 +158,48 @@ class SearchPage extends StatelessWidget {
   }
 }
 
+class InstallAll extends ConsumerWidget {
```
d-loose, line 161: Could this just be an `ElevatedButton` in a `Center` with appropriate padding?

packages/app_center/lib/src/search/search_page.dart
```diff
@@ -267,3 +318,26 @@ extension on PackageFormat {
     };
   }
 }
+
+Future<void> installAll(List<Snap> snaps) async {
```
d-loose, line 322: As discussed before, this business logic shouldn't be here in the UI, but in a dedicated model. I'd suggest adding a `MultiSnapModel` that takes a snap category as an argument and handles all the communication with the `SnapdService` (similar to what `SnapModel` does for a single snap). You could then create a `multiSnapModelProvider` family and use it in the UI like this:

```dart
...
final model = ref.watch(multiSnapModelProvider(category));
...
ElevatedButton(
  onPressed: model.installAll,
  child: Text(...),
)
...
```
Of course this needs some additional logic for the loading state, similar to the snap model.

The model itself could roughly look like this:
```dart
class MultiSnapModel extends ChangeNotifier {
  MultiSnapModel({
    required this.snapd,
    required this.category,
  }) ...

  Future<void> init() async {
// fetch snaps from snapd (instead of doing this in the search provider)
...
  }

  Future<void> installAll() {
// handle confinement logic and install all the snaps
  }

File: packages/app_center/lib/src/snapd/snap_search.dart [hunk 13]
```diff
@@ -43,6 +43,16 @@ final snapSearchProvider =
...
+  } else if (searchParameters.category == SnapCategoryEnum.gameDev ||
+      searchParameters.category == SnapCategoryEnum.gameEmulators ||
+      searchParameters.category == SnapCategoryEnum.gnomeGames ||
+      searchParameters.category == SnapCategoryEnum.kdeGames) {
+    yield* snapd.getStoreSnaps(
+      searchParameters.category!.featuredSnapNames
+              ?.where((name) => name.contains(searchParameters.query ?? ''))
+              .toList() ??
+          [],
+    );
```
d-loose, line 55: As mentioned above, I'd move this into its own model, since this isn't really search related anymore.
Maybe it'd be good to completely separate the search page from the category page again as it's quite messy (in a future PR, though).

packages/app_center/lib/src/store/store_navigator.dart
```diff
@@ -69,4 +73,8 @@ extension StoreNavigatorState on NavigatorState {
       category: category,
     ));
   }
+
+  Future<void> pushExternalTools({String? query, String? category}) {
+    return pushNamed('/externalTools');
```
d-loose, line 78: See above

packages/app_center/test/games_page_test.dart
d-loose, line 1: Awesome :tada:

d-loose (changes requested):
Great work overall! I've left a few minor comments on general issues regarding conventions and structure. In a lot of places [trailing commas](https://docs.flutter.dev/tools/formatting#using-trailing-commas) would help increase legibility - I didn't leave explicit comments for that.
One thing that still needs a bit of work is the many-snap installation logic. It should really be separated from the UI and moved into a model that interacts with snapd and is made accessible to the UI via a provider family.

packages/app_center/lib/src/games/games_page_external_tools.dart
```diff
@@ -0,0 +1,164 @@
...
+    final l10n = AppLocalizations.of(context);
+    return ResponsiveLayoutScrollView(
+      slivers: [
+        AppCardGrid.fromTools(tools: tools, l10n: l10n),
+      ],
+    );
+  }
+}
+
+//TODO: should we localize this stuff?
```
d-loose, line 108: Forgot to comment here - yes, absolutely! Data like this also shouldn't be stored in the UI code, but moved to a file or database and then be loaded by a view model that provides it to the UI

packages/app_center/lib/src/games/games_page_external_tools.dart
```diff
@@ -0,0 +1,164 @@
...
+            child: Column(
+              crossAxisAlignment: CrossAxisAlignment.start,
+              children: [
+                Container(
+                  padding: const EdgeInsets.all(48),
+                  decoration: BoxDecoration(
+                    borderRadius: BorderRadius.circular(8),
+                    gradient: const LinearGradient(
+                      colors: [
+                        Color.fromARGB(255, 29, 27, 112),
```
BLKKKBVSIK, line 39: We shouldn't have hard-coded colors directly here.

packages/app_center/lib/src/widgets/app_card.dart
jssotomdz: I remember doing this to avoid a circular dependency. Is it better to create a tool.dart that lives in `lib/src/games` and just import it where necessary?

packages/app_center/lib/src/search/search_page.dart
jssotomdz: I need to get the snaps from the category, which are fetched with the `snapSearchProvider` which in turn requires a `WidgetRef ref`. The way I solved this was wrapping it in a `ConsumerWidget`. Does it make sense? Am I missing something?

jssotomdz: Got the point, implemented the multiple snaps model and now this isn't needed. Will refactor in next commits

packages/app_center/lib/src/snapd/snap_search.dart
d-loose: Ah, right I see what you mean now. You'll need to keep it then, as long you're using the search page to display a gaming category with the 'install all' button. In this case it might make sense to simplify the `MultiSnapModel` and directly pass the list of snaps to it instead of fetching them in the  `init()` again. That'd avoid unnecessary requests and also ensure that the snaps that would be installed are actually exactly the same as those shown on the page.
I'll add some more detailed comments on the `MultiSnapModel` later today!

d-loose: Noticed you're OOO until after the weekend, so I'll comment on the new changes sometime tomorrow - looks very good at first glance!

packages/app_center/lib/src/games/games_page.dart
```diff
@@ -0,0 +1,177 @@
...
+  Widget build(BuildContext context) {
+    return Text(text, style: Theme.of(context).textTheme.headlineSmall);
+  }
+}
+
+class _CategoryBannerProperties {
+  static const EdgeInsets padding = EdgeInsets.all(24);
+  static const double height = 150;
+  static const Size maxSize = Size(60, 60);
+  static const Size iconSize = Size(32, 32);
```
d-loose, line 163: `Size.square()` would be clearer here

packages/app_center/lib/src/games/games_page.dart
```diff
@@ -0,0 +1,177 @@
...
+}
+
+List<Tool> takeRandom(List<Tool> tools, int num) {
```
d-loose, line 166: it might be simpler to use `List`'s `shuffle()` and `take()`

packages/app_center/lib/src/snapd/multisnap_model.dart
```diff
@@ -0,0 +1,161 @@
...
+    snapd: getService<SnapdService>(),
+    category: category,
+  )..init(),
+);
+
+final changeProvider =
+    StreamProvider.family.autoDispose<SnapdChange, String?>((ref, id) {
+  if (id == null) return const Stream.empty();
+  return getService<SnapdService>().watchChange(id);
+});
```
d-loose, line 23: This isn't used anywhere (yet). Do you plan to add a progress indicator?

packages/app_center/lib/src/snapd/multisnap_model.dart
```diff
@@ -0,0 +1,161 @@
...
+  final SnapdService snapd;
+  final SnapCategoryEnum category;
+
+  String? get activeChangeId => _activeChangeId;
+  String? _activeChangeId;
+
+  AsyncValue<void> get state => _state;
+  AsyncValue<void> _state;
+
+  late List<Snap> categorySnaps;
```
d-loose, line 39:
```suggestion
  final List<Snap> categorySnaps = [];
```

packages/app_center/lib/src/snapd/multisnap_model.dart
```diff
@@ -0,0 +1,161 @@
...
+
+  StreamSubscription<Snap?>? _storeSnapSubscription;
+  StreamSubscription<SnapdChange>? _activeChangeSubscription;
+
+  Stream<SnapdException> get errorStream => _errorStreamController.stream;
+  final StreamController<SnapdException> _errorStreamController =
+      StreamController.broadcast();
+
+  Future<void> init() async {
+    categorySnaps = List.empty(growable: true);
```
d-loose, line 49:
```suggestion
```

packages/app_center/lib/src/snapd/multisnap_model.dart
```diff
@@ -0,0 +1,161 @@
...
+    final storeSnapCompleter = Completer();
+    for (final name in category.featuredSnapNames!) {
```
d-loose, line 51: It'd be nice to handle the case where a category doesn't have any featured snaps more gracefully here. A simple `assert()` to warn other contributors would be enough for now.

d-loose: Also, I'm just realizing that we have `getStoreSnaps` for multiple snaps

packages/app_center/lib/src/snapd/multisnap_model.dart
```diff
@@ -0,0 +1,161 @@
...
+      _handleError(e);
+    }
+  }
+
+  Future<void> cancel() async {
+    if (activeChangeId == null) return;
+    await snapd.abortChange(activeChangeId!);
+  }
+
+  Future<void> install() {
```
d-loose, line 115: This should probably be called `installAll`

packages/app_center/lib/src/snapd/multisnap_model.dart
```diff
@@ -0,0 +1,161 @@
...
+    for (final task in tasks) {
+      done += task.progress.done;
+      total += task.progress.total;
+    }
+
+    return total != 0 ? done / total : 0;
+  }
+}
+
+Future<String> installAll(List<Snap> snaps) async {
```
d-loose, line 135: This helper method shouldn't be public

packages/app_center/lib/src/snapd/multisnap_model.dart
```diff
@@ -0,0 +1,161 @@
...
+    var done = 0.0;
+    var total = 0.0;
+    for (final task in tasks) {
+      done += task.progress.done;
+      total += task.progress.total;
+    }
+
+    return total != 0 ? done / total : 0;
+  }
+}
```
d-loose, line 133: Rather than duplicating this, I'd move it to a common file like `snapx.dart`

packages/app_center/lib/src/snapd/multisnap_model.dart
```diff
@@ -0,0 +1,161 @@
...
+
+Future<String> installAll(List<Snap> snaps) async {
+  final snapd = getService<SnapdService>();
```
d-loose, line 136: Once this is part of the `MultiSnapModel` you can use the model's `snapd` reference

packages/app_center/lib/src/widgets/app_card.dart
d-loose: If it's not intended to be used outside of the 'games' context, I'd move `tool.dart` to `lib/src/games` and export it from `lib/games.dart`.

packages/app_center/lib/src/games/games_page.dart
d-loose: Right, shuffling the list isn't a great idea then. I'd suggest using [`sample()`](https://api.flutter.dev/flutter/package-collection_collection/IterableExtension/sample.html) from [`collection`](https://pub.dev/packages/collection) since we depend on it anyways.

jssotomdz: it has the same problem but now that I think again I could add a `list.where('has logo').sample(3)`. I'll implement it in the next commits

packages/app_center/test/snap_model_test.dart
```diff
@@ -275,4 +276,23 @@ void main() {
       );
     }
   });
+  test('install-many', () async {
+      final service = createMockSnapdService(
+        storeSnap: storeSnap,
+      );
+      final model = MultiSnapModel(snapd: service, category: SnapCategoryEnum.gameDev);
```
d-loose: Since we're testing the `MultiSnapModel` here, this belongs in a separate `multi_snap_model_test.dart`.

packages/app_center/test/test_utils.dart
```diff
@@ -215,6 +215,35 @@ MockSnapdService createMockSnapdService({
...
+                  ),
+                  'latest/edge': SnapChannel(
+                    confinement: SnapConfinement.classic,
+                    size: 31337,
+                    releasedAt: DateTime(1970, 1, 2),
+                    version: '2.0.0',
+                  ),
+                },
+              ),
+          growable: false)));
```
d-loose, line 247: Instead of hardcoding this here in the test utils, could we add a corresponding argument to `createMockSnapdService` so that each test can setup the mock according to its needs?

d-loose:
Do you intend to use the `MultiSnapModel`'s `errorStream` and `activeChangeId` to present errors and installation progress in the UI? If so, please add tests for those :). Otherwise, let's remove those parts to avoid having dead code lying around.

packages/app_center/test/search_page_test.dart
```diff
@@ -115,6 +121,31 @@ void main() {
...
+          category: 'gameDev',
+        ),
+      ),
+    );
+
+    await tester.pumpAndSettle();
+
+    expect(find.text(tester.l10n.snapCategoryGameDev), findsOneWidget);
+    expect(find.byType(MenuButtonBuilder<SnapCategoryEnum?>), findsNothing);
+    expect(find.text(tester.l10n.installAll), findsOneWidget);
```
d-loose, line 143: We could also test whether pressing the 'install all' button calls the corresponding method here

packages/app_center/lib/src/widgets/app_card.dart
jssotomdz: I am reusing the banners, appcards, app title and snap grid widgets to create the tools page and they need access to tool.dart. Games page needs access to these widgets hence creating a circular dependency. Not sure if that's considered outside of games context but these widgets do require the class. Should I proceed with the move anyways?

packages/app_center/test/search_page_test.dart
```diff
@@ -137,13 +139,15 @@ void main() {
     );
 
     await tester.pumpAndSettle();
+    await tester.tap(find.byType(ElevatedButton));
```
d-loose, line 142: Nitpick: I'd move this further down below your UI assertions, as those test the state of the UI after `pumpAndSettle()`. The only effect of tapping the button is calling `installAll()` - to make that clear, it'd be best to group `tester.tap()` and `verify(..)` together.

packages/app_center/test/multisnap_model_test.dart
```diff
@@ -0,0 +1,50 @@
...
+      confinement: SnapConfinement.classic,
+      size: 31337,
+      releasedAt: DateTime(1970, 1, 2),
+      version: '2.0.0',
+    ),
+  },
+);
+
+void main() {
+  test('install-many', () async {
```
d-loose, line 36: I think it'd be good to also add a test with mixed classic and strictly confined snap, as this was the case you had issues with.

d-loose (approved):
Great work! :+1:
Left two small comments. I think in general there's some room for improvement concerning the tests and the UI (e.g. loading states). But I think we could address those things in a future PR :)

packages/app_center/test/multisnap_model_test.dart
jssotomdz: this will be addressed in a future pr

jssotomdz (approved):

PR #1599, opened 2024-02-13
ci: Produce snap artifact every build, not just on manual runs

ashuntu:
The workflow that builds the Snap only actually uploads the artifact on manual runs. It would be useful to produce an artifact every build to aid in testing ability, especially for those who don't want to checkout the code themselves.

This PR removes the `workflow_dispatch` condition of the artifact upload so that it should produce an artifact every time it runs.

You can even see the artifact produced for this PR with this change!: https://github.com/ubuntu/app-center/actions/runs/7890140289?pr=1599

### DIFF
--- a/.github/workflows/snap.yaml
+++ b/.github/workflows/snap.yaml
@@ -17,7 +17,6 @@ jobs:
       - uses: snapcore/action-build@v1
         id: snapcraft
       - uses: actions/upload-artifact@v3
-        if: github.event_name == 'workflow_dispatch'
         with:
           name: 'snap'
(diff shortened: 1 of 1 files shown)


tim-hm:
@ashuntu I think there's a type-o in your commit message: `Product` -> `Produce`.

tim-hm:
Good pt @d-loose. I suggest dev always builds on new commits ... and feature branches only on manual invocation?

spydon:
Closing this since it's not a behaviour we want for now.

PR #1601, opened 2024-02-19
fix(snap): apply melos setting before running melos

d-loose:
Might help to apply the correct setting *before* running melos .. :facepalm:


spydon (approved):
😅

PR #1606, opened 2024-02-24
[manage_page]: fix the buttons alignment

soumyaDghosh:
### After change
image

### Before Change
image

### DIFF
--- a/packages/app_center/lib/src/manage/manage_page.dart
+++ b/packages/app_center/lib/src/manage/manage_page.dart
@@ -485,6 +485,8 @@ class _ManageSnapTile extends ConsumerWidget {
             final updatesModel = ref.watch(updatesModelProvider);
 
             return PushButton.outlined(
+              style: const ButtonStyle(
+                  padding: MaterialStatePropertyAll(EdgeInsets.zero)),
               onPressed: updatesModel.activeChangeId != null
                   ? null
                   : () {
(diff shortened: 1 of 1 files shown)


Feichtmeier:
The clipping is now gone 👍 

However I see no reason to have different heights of buttons in the app.

Can't we use the normal height of outline/elevated/filled buttons everywhere in the app and everywhere on this page?
@anasereijo

Feichtmeier:
> > However I see no reason to have different heights of buttons in the app.
> > Can't we use the normal height of outline/elevated/filled buttons everywhere in the app and everywhere on this page?
> 
> That looks odd indeed. If I remember correctly, in Figma, we have the buttons all with the same height. Could you please confirm what is the default height?

The content padding for material buttons styled with yaru.dart, as everywhere in community "yaru" apps so far, have a content padding of 16, what makes them in the end having this size:

https://ubuntu.github.io/yaru.dart/
grafik

To fix this problem everywhere in this app with one change - I would suggest to change the buttons in app center to plain material flutter buttons with the yaru.dart theme 

Then you have the choice between TextButton, OutlinedButton, ElevatedButton (the green ones), and FilledButton (the grey ones)

Feichtmeier (approved):
This specific change is def. better than main. 
[My suggestion here](https://github.com/ubuntu/app-center/pull/1606#issuecomment-1966579702) can be moved into a different issue
LGTM!

PR #1620, opened 2024-03-19
fix: make back buttons round

Feichtmeier:
I think the placement could involve canonical designers.
But the shape should be a circle I think :)

Before:
grafik

After:
grafik


Feichtmeier:
lets get the other branch in first :P

PR #1621, opened 2024-03-19
feat: use a paned look

### LINKED ISSUE #1562: Enhancement: A more gnome like titlebar
Screenshot from 2024-01-07 20-51-25
Screenshot from 2024-01-07 20-51-20

I was not able to implement the titlebar in the snap_page, but, I don't think it's tough.

@anasereijo @BLKKKBVSIK thoughts?

### LINKED ISSUE #1612: Incorrect Coloring of Search Field with "High Contrast" accessibility enabled
The "High Contrast" accessibility setting works properly but there appears to be an error in the way the search field is colored - I tried typing in it but I can't read the letters. Tested in the latest **noble numbat**  

Edit: Confirmed the issue on Mantic Minotaur too

I happily remain at your disposal for any further information - I wasn't sure if this was a launchpad-appropriate bug so I filed it here directly

Screenshot from 2024-03-12 10-37-33

Feichtmeier:
- to reduce the number of visual color background areas, use the paned look that can be found in gtk4 apps
- since the right titlebar then has more contrast against the search field, remove the searchfield fill color override, this also...

Fixes #1612
Fixes #1562

| | Before | After |
|-|-|-|
|light|grafik|grafik|
|dark|grafik|grafik|


Feichtmeier:
CC @anasereijo

Feichtmeier:
@anasereijo 
If you might be concerned about the whitespace, we could also move certain elements out of the bottom left bar into the appbars at the top
grafik
(example)
I think a fully fletched dedicated menu entry only for "about" is very uncommon anyways
what modern gnome apps often do, is place a hamburger menu in the top left bar if it is a paned application
which then leads to a setting dialog
grafik
grafik
(examples and mockups)

anasereijo:
@Feichtmeier sorry for the delayed reply. We will be revisiting the App Center next cycle (starting mid May). These are great proposals, and we will be looking at this for sure when we start looking at the store again.

Thanks for your input 🙏

Feichtmeier:
@anasereijo 

Just had another look on this, and I think the settings button is fine in the left panel.
The visual center really improves here because now the right pane content and appbar are on the same axis

grafik

anasereijo (approved):

spydon (approved):
Lgtm, can you open an issue for moving the other things around @Feichtmeier ?

anasereijo:
@Feichtmeier changes approved, it looks much better! Thanks for doing this, and apologies for taking a long time to come back to you.

PR #1622, opened 2024-03-19
feat: make back buttons round

Feichtmeier:
I think the placement could involve canonical designers.
But the shape should be a circle I think :)

Before:
grafik

After:
grafik


Feichtmeier:
CC @anasereijo

anasereijo:
good call! I think looks much better, and more in-line with the theme

d-loose (approved):

PR #1648, opened 2024-04-22
fix(snap): Let snap-store snap refresh while the app is running Fixes #1643

### LINKED ISSUE #1643: Snap store update via the app center fails
(X11 session) The same error was showing up for Firefox as well. I went to the Firefox page on the App Center separately, and hit "Update" and that was resolved.

But, that doesn't work for snap store. I can't seem to update the snap-store via terminal as well (it shows the same error) even if the app center is closed.

A user should not go all the way to kill the specific process ID just to apply an update available here. Defeats the purpose of having an app center and getting the "update" button.

[Screencast from 2024-04-14 18-08-19.webm](https://github.com/ubuntu/app-center/assets/14110456/28477efd-204d-498b-b1bb-c65dc2b23d4c)

kenvandine:
This will require a snapcraft built after
https://github.com/canonical/snapcraft/pull/4749 has been merged to build.


kenvandine:
Snapcraft 8.2.7 can build this, which is now released in the stable channel.  This should be safe to release.

d-loose (approved):

d-loose:
Awesome, thanks a lot Ken!

PR #1660, opened 2024-04-29
refactor: Migrate away snap model from old provider style

spydon:
This migrates away from the custom `ChangeNotifierProvider` approach.

It also makes sure that futures aren't completed until the polling of the changeId has returned which makes the the structure slightly cleaner.

Things that are left to consider:
 * Whether the snap operations should be happening within a model instead of in separate providers like they do now.
 * How to handle actions from buttons.

With a refactor this big I'm sure that some bugs have been introduced, so make sure to test a lot if you continue working on this!


packages/app_center/lib/src/snapd/snap_data.dart
```diff
@@ -0,0 +1,65 @@
...
+import 'package:collection/collection.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:freezed_annotation/freezed_annotation.dart';
+import 'package:snapd/snapd.dart';
+
+part 'snap_data.freezed.dart';
+
+// TODO: Better naming, easily confused with the Snap class.
+@freezed
+class SnapData with _$SnapData {
```
spydon, line 11: Can we come up with a better name for this class?

packages/app_center/lib/src/snapd/snap_model.dart
```diff
@@ -1,223 +1,237 @@
...
+    return switch (action) {
+      SnapAction.cancel => abort,
+      SnapAction.install =>
+        state.valueOrNull?.storeSnap != null ? install : null,
+      SnapAction.open =>
+        launcher?.isLaunchable ?? false ? launcher!.open : null,
+      SnapAction.remove => remove,
+      // TODO: This must surely be wrong?
+      SnapAction.switchChannel =>
+        state.valueOrNull?.storeSnap != null ? refresh : null,
```
spydon, line 164: @d-loose do you know how this worked before?

d-loose: It was the same before and looks correct to me. The snap vocabulary is just slightly confusing - to switch a channel you essentially run `snap refresh --channel=foo ..`. We decided to use the more commonly understood labels 'Update' and 'Switch channel' instead of just using 'Refresh' in the UI.

packages/app_center/lib/src/snapd/snap_data.dart
d-loose: What if we stick with `SnapModel` for the view model (it's called `SnapPackage` in this PR) and use `SnapModelState` for the state it exposes?

d-loose: Hm, that of course conflicts a bit with `AsyncValue<Snap?> storeSnapState`...

packages/app_center/lib/src/snapd/snap_model.dart
```diff
@@ -1,223 +1,218 @@
...
-  )..init(),
-);
+part 'snap_model.g.dart';
 
-final progressProvider =
-    StreamProvider.family.autoDispose<double, List<String>>((ref, ids) {
-  final snapd = getService<SnapdService>();
+@Riverpod(keepAlive: true)
+class SnapPackage extends _$SnapPackage {
+  late final _snapd = getService<SnapdService>();
```
d-loose, line 17: I guess this doesn't need to be `late`?

packages/app_center/lib/src/snapd/snap_model.dart
```diff
@@ -1,223 +1,218 @@
...
-      })
-  };
-  ref.onDispose(() {
-    for (final subscription in subscriptions.values) {
-      subscription.cancel();
+  @override
+  Future<SnapData> build(String snapName) async {
+    Snap? localSnap;
+    try {
+      localSnap = await getService<SnapdService>().getSnap(snapName);
```
d-loose, line 23:
```suggestion
      localSnap = await _snapd.getSnap(snapName);
```

packages/app_center/lib/src/snapd/snap_model.dart
```diff
@@ -1,223 +1,218 @@
...
+    final AsyncValue<Snap?> storeSnapState;
+    if (localSnap == null) {
+      // If we don't have a local snap we keep the provider in loading until the
+      // store snap is loaded.
+      storeSnapState =
+          AsyncData(await ref.watch(storeSnapProvider(snapName).future));
+    } else {
+      // Since we have a local snap we don't need to wait for the store snap to
+      // be loaded, but the store snap will be updated in the data once it is.
+      storeSnapState = ref.watch(storeSnapProvider(snapName));
```
d-loose, line 37: I really like this solution :)

packages/app_center/lib/src/snapd/snap_model.dart
```diff
@@ -1,223 +1,218 @@
...
-      _selectedChannel = 'latest/stable';
-    } else {
-      _selectedChannel =
-          channels?.firstWhereOrNull((c) => c.contains('stable')) ??
-              channels?.firstOrNull;
-    }
+  /// Uninstalls the snap.
+  Future<void> remove() async {
+    assert(state.hasValue, 'The snap must be loaded before removing it');
+    final changeId = await getService<SnapdService>().remove(snapName);
```
d-loose, line 112:
```suggestion
    final changeId = await _snapd.remove(snapName);
```

packages/app_center/lib/src/snapd/snap_model.dart
```diff
@@ -1,223 +1,218 @@
...
-    } else {
-      _activeChangeSubscription =
-          snapd.watchChange(id).listen(_activeChangeListener);
-    }
+  @visibleForTesting
+  void setMockState(
+    AsyncValue<SnapData> mockState,
+  ) {
+    state = mockState;
   }
```
d-loose, line 129: Is this needed? The `state` setter is public and can be used directly

packages/app_center/test/snap_page_test.dart
d-loose, line 1: Some of the tests fail when run individually due to missing services. I'd suggest we add `tearDown(resetAllServices)` (and make sure to do this consistently in all tests that rely on the service locator).

packages/app_center/test/test_utils.dart
```diff
@@ -220,6 +222,7 @@ MockSnapdService createMockSnapdService({
   when(service.getStoreSnaps(any)).thenAnswer((_) => Stream.value(
       List<Snap>.generate(storeSnapsCount!, (index) => storeSnap!,
           growable: false)));
+  registerMockService<SnapdService>(service);
```
d-loose, line 225: With this statement included here, we should rename the method to `registerMockSnapdService` to distinguish it from the methods that don't register their mocks (yet).

packages/app_center/lib/src/snapd/snap_model.dart
d-loose: Hm, or maybe not? :sweat_smile:
In the case where both a local and a store snap are available there's an ugly rebuild:

[Screencast from 2024-06-07 16-20-24.webm](https://github.com/ubuntu/app-center/assets/113362648/43a764f3-e930-4e6f-99cd-67df01ad47cf)

packages/app_center/lib/src/snapd/snap_page.dart
```diff
@@ -25,66 +26,52 @@ const _kChannelDropdownWidth = 220.0;
...
-      data: (_) => ResponsiveLayoutBuilder(
-        builder: (_) => _SnapView(
-          snapModel: snapModel,
-          updatesModel: updatesModel,
-        ),
+    ref.listen(snapPackageProvider(snapName), (_, value) {
+      if (value.hasError && value.error is SnapdException) {
+        showError(context, value.error as SnapdException);
+      }
+    });
```
d-loose, line 50: Without catching the exceptions thrown by the asynchronous service calls in the view model and setting a corresponding state, this doesn't work.

Example where I'm trying to refresh firefox while its running:

[Screencast from 2024-06-07 16-28-06.webm](https://github.com/ubuntu/app-center/assets/113362648/cafbce15-8541-4283-9a32-5241f045c90a)

In general, I think errors that result from user initiated actions shouldn't put the entire view model in an error state, since the state itself is still valid.
I'm sure there's a cleaner solution than the one introduced in #1396 though.

packages/app_center/lib/src/snapd/snap_model.dart
spydon: I guess this provider will always be used so it doesn't matter here, but it could be good to keep it late so that other providers follow the same pattern to lazily get the services?

packages/app_center/lib/src/snapd/snap_model.dart
spydon: Yeah, I think that is because it is using the snap field instead of the `storeSnapState` field, so that it can show loaders on the places where it is waiting for store snap stuff. I tried to minimize the changes to the actual UI code in this PR, so it still uses the old fields in many places, I thought that would be better as a follow-up, but maybe it is better to do it in here?

packages/app_center/lib/src/snapd/snap_model.dart
```diff
@@ -1,222 +1,242 @@
...
-    for (final subscription in subscriptions.values) {
-      subscription.cancel();
+  @override
+  Future<SnapData> build(String snapName) async {
+    Snap? localSnap;
+    try {
+      localSnap = await _snapd.getSnap(snapName);
+    } on SnapdException catch (e) {
+      // TODO: Do we really need a try-catch here, don't we want to go into
+      // the AsyncError in all cases?
```
d-loose, line 28: Here we're handling the case where the snap isn't locally installed, but it's available in the store (`SnapdClient` throws an exception when it receives a 4xx status from snapd).

spydon: Not a very smooth way of dealing with a 4xx, maybe we should fix that in the snapdclient at some point.

packages/app_center/lib/src/snapd/snap_model.dart
d-loose: Maybe until we sort out the `yaru.dart` issues we can simply force `AppCard` to use the theme's `cardColor`

```diff
diff --git a/packages/app_center/lib/src/widgets/app_card.dart b/packages/app_center/lib/src/widgets/app_card.dart
index 365bf709..5948faec 100644
--- a/packages/app_center/lib/src/widgets/app_card.dart
+++ b/packages/app_center/lib/src/widgets/app_card.dart
@@ -75,6 +75,7 @@ class AppCard extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return YaruBanner(
+      color: Theme.of(context).cardColor,
       padding: const EdgeInsets.all(kCardSpacing),
       onTap: onTap,
       child: Flex(
```

d-loose:
> Looks like there are some yaru related style changes (background colors of the containers) we need to address:


Maybe until we sort out the `yaru.dart` issues we can simply force `AppCard` to use the theme's `cardColor`

```diff
diff --git a/packages/app_center/lib/src/widgets/app_card.dart b/packages/app_center/lib/src/widgets/app_card.dart
index 365bf709..5948faec 100644
--- a/packages/app_center/lib/src/widgets/app_card.dart
+++ b/packages/app_center/lib/src/widgets/app_card.dart
@@ -75,6 +75,7 @@ class AppCard extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return YaruBanner(
+      color: Theme.of(context).cardColor,
       padding: const EdgeInsets.all(kCardSpacing),
       onTap: onTap,
       child: Flex(
```

spydon:
> > Looks like there are some yaru related style changes (background colors of the containers) we need to address:
> 
> Maybe until we sort out the `yaru.dart` issues we can simply force `AppCard` to use the theme's `cardColor`
> 
> ```diff
> diff --git a/packages/app_center/lib/src/widgets/app_card.dart b/packages/app_center/lib/src/widgets/app_card.dart
> index 365bf709..5948faec 100644
> --- a/packages/app_center/lib/src/widgets/app_card.dart
> +++ b/packages/app_center/lib/src/widgets/app_card.dart
> @@ -75,6 +75,7 @@ class AppCard extends StatelessWidget {
>    @override
>    Widget build(BuildContext context) {
>      return YaruBanner(
> +      color: Theme.of(context).cardColor,
>        padding: const EdgeInsets.all(kCardSpacing),
>        onTap: onTap,
>        child: Flex(
> ```
> 
> P.S. We'd need a corresponding change for the `SnapImageCard`, which is a separate widget

Fixed in https://github.com/ubuntu/app-center/pull/1660/commits/06fe6cf30da120605671ca1925920267e12496b5 and https://github.com/ubuntu/app-center/pull/1660/commits/79f7dc4982ea189dcd723333ef6154073b795b5c

packages/app_center/lib/src/snapd/snap_model.dart
d-loose: I think all our dart client libraries for HTTP APIs do it that way at the moment.

In the subiquity client we've got this:
```dart
    if (response.statusCode != 200) {
      throw SubiquityException(method, response.statusCode, str);
    }
```

But if you think it's worth improving this, I'm up for it :)

packages/app_center/lib/src/snapd/snap_page.dart
```diff
@@ -513,40 +487,6 @@ class _RatingsActionButtons extends ConsumerWidget {
...
-      switch (this) {
-        cancel => model.cancel,
-        install => model.storeSnap != null ? model.install : null,
-        open => launcher?.isLaunchable ?? false ? launcher!.open : null,
-        remove => model.remove,
-        switchChannel => model.storeSnap != null ? model.refresh : null,
-        update => model.storeSnap != null ? model.refresh : null,
-      };
-}
-
```
d-loose, line 549: Just a final nitpick/question, since it came up in the meeting with Innes as well: to me this feels like UI specific code (maybe icons/l10n more than the actions and corresponding callbacks). Should we keep (some of) this in `snap_page.dart`?

packages/app_center/lib/src/snapd/snap_page.dart
```diff
@@ -559,7 +499,7 @@ class _Section extends YaruExpandable {
 class _Header extends ConsumerWidget {
   const _Header({required this.snapModel});
 
-  final SnapModel snapModel;
+  final SnapData snapModel;
```
d-loose, line 502: I guess we should rename the variable as well to avoid confusion

packages/app_center/lib/src/snapd/snap_page.dart
```diff
@@ -611,13 +551,13 @@ class _Header extends ConsumerWidget {
   }
 }
 
-class _ChannelDropdown extends StatelessWidget {
+class _ChannelDropdown extends ConsumerWidget {
   const _ChannelDropdown({required this.model});
 
-  final SnapModel model;
+  final SnapData model;
```
d-loose, line 557: Same here :)

packages/app_center/lib/src/snapd/snap_page.dart
spydon: Yeah, I don't remember why I moved it now, it should maybe even get its own file

spydon: Done!

PR #1664, opened 2024-05-05
Move the back button of search_page to the left

oyajun:
Before
image

After
image


spydon:
Feels like the most natural would be to have it like this:
app-center-back

What do you think @anasereijo?

spydon:
Thanks a lot for your effort to fix this @oyajun!
We got another design of it implemented in #1726, so I'll close this.

PR #1669, opened 2024-05-11
Used @freezed to create Rating class

j-j-gajjar:


spydon (approved):
Lgtm!

packages/app_center_ratings_client/lib/src/chart.freezed.dart
```diff
@@ -30,6 +30,8 @@ abstract class $ChartDataCopyWith<$Res> {
       _$ChartDataCopyWithImpl<$Res, ChartData>;
   @useResult
   $Res call({double rawRating, common.Rating rating});
+
+  $RatingCopyWith<$Res> get rating;
```
d-loose, line 34: There seems to be an issue with nested `freezed` classes here - this is only part of `ratings.dart` but not `chart.dart`. `Chart` is another `freezed` class, that has a `Rating` field.

I don't know from the top of my head how to fix this, though.

d-loose (changes requested):

d-loose (approved):
Code generation really doesn't work well with namespaces and typedefs sometimes.. Thanks for fixing!

Issue #1407, opened 2023-09-26
Support installing local deb packages


d-loose:
There's no support for dealing with local packages yet. However 'snap-store' is supposed to handle those by default:

```
$ grep debian /usr/share/applications/defaults.list 
application/vnd.debian.binary-package=snap-store_ubuntu-software-local-file.desktop
application/x-debian-package=snap-store_ubuntu-software-local-file.desktop
```

Do we have an alternative way of handling those for the time being?

soumyaDghosh:
> Do we have an alternative way of handling those for the time being?

We can use `gdebi` or `eddy` for those, until the feature doesn't comes up in snap-store.

Baltix:
@soumyaDghosh , what is `eddy`, there are no package `eddy` in official Ubuntu repos :(

> We can use `gdebi` or `eddy` for those, until the feature doesn't comes up in snap-store.

soumyaDghosh:
> @soumyaDghosh , what is `eddy`, there are no package `eddy` in official Ubuntu repos :(
> 
> > We can use `gdebi` or `eddy` for those, until the feature doesn't comes up in snap-store.
> 
> 

Eddy is probably not in the repo.

Feichtmeier:
As an inspiration the previous incarnation of this app here had local deb side loading (it is still in the preview/edge channel to try)
Alternatively we could eventually create a second app for only the deb and/or app image sideloading? Similar to how ot looks on Mac maybe? I mean, the code is still here:

https://github.com/ubuntu/app-center/blob/archive/main/lib/app/package_installer/package_installer_page.dart
https://github.com/ubuntu/app-center/tree/archive/main/lib/app/common/packagekit
https://github.com/ubuntu/app-center/tree/archive/main/lib/services/packagekit

🤷 

https://github.com/ubuntu/app-center/assets/15329494/fdbc0d51-316b-4b38-aecf-bff7418f893d

mrworldwide1:
> In my opinion, it is absolutely essential that this problem is fixed for the next version of Ubuntu, in fact, I really think that the current version of Ubuntu should have the new store without this problem.
> 
> The most common thing is that a new user, to install Google Chrome for example, goes to the official website, downloads the binary and double-clicks to install. When he notices that it doesn't work he will be completely lost... As Ubuntu is considered one of the most important distros for new users, it is very important that this problem is resolved.

[...]

PR #1680, opened 2024-05-28
feat(packagekit): add local deb methods

d-loose:
Adds packagekit methods to retrieve local package information and install a local package.

Ref #1407
UDENG-1490


spydon (approved):
Lgtm!

PR #1681, opened 2024-05-29
feat: implement UI for 3rd party debs

d-loose:
Adds a first skeleton for the local deb package UI. As we're in the process of refactoring the entire app to follow a more standardized riverpod-based architecture, things in this PR might change a lot.

Ref #1407
UDENG-1490
UDENG-3041


packages/app_center/lib/src/deb/local_deb_model.dart
```diff
@@ -0,0 +1,39 @@
...
+
+part 'local_deb_model.freezed.dart';
+part 'local_deb_model.g.dart';
+
+@freezed
+class LocalDebData with _$LocalDebData {
+  factory LocalDebData({
+    required String path,
+    // TODO: use PackageKitPackageDetails type alias here (freezed seems to
+    //       expand typedefs, which causes some issues)
```
spydon, line 15: Since this class probably isn't going to change much, maybe it is easier if we just make an `@immutable` class instead of a freezed one, if the code generation is causing problems?

packages/app_center/lib/src/deb/local_deb_model.dart
```diff
@@ -0,0 +1,39 @@
...
+class LocalDebModel extends _$LocalDebModel {
+  @override
+  Future<LocalDebData> build({required String path}) async {
+    final packageKit = getService<PackageKitService>();
+    final details = await packageKit.getDetailsLocal(path);
+    return LocalDebData(path: path, details: details);
+  }
+
+  Future<void> install() async {
+    if (!state.hasValue) return;
```
spydon, line 31: Do we want to throw an exception here, so that it comes back with an error?
I've been experimenting with having a `PrematureSnapOperationException` for these cases, but there are of course other ways to do it too.

packages/app_center/test/packagekit_service_test.dart
```diff
@@ -19,8 +20,11 @@ void main() {
   group('activate service', () {
     test('service available', () async {
       final dbus = createMockDbusClient();
-      final packageKit =
-          PackageKitService(dbus: dbus, client: createMockPackageKitClient());
+      final packageKit = PackageKitService(
+        dbus: dbus,
+        client: createMockPackageKitClient(),
+        fs: MemoryFileSystem.test(),
```
spydon, line 26: Would it, similarly to having a provider for the filesystem, make sense to have a service for it that could be overridden, so that it could just be done in the test setup step instead of having to pass it in explicitly?

spydon (commented):
Overall it looks good!

packages/app_center/test/packagekit_service_test.dart
d-loose: Hm, not sure. To do that consistently we'd have to register services for the DBus and PackageKit clients (and potentially others) as well. In the current architecture all of those "low level" services are hidden from the service layer in the application by passing them explicitly to the services we actually want to expose.
Would giving up that separation have any other benefits besides simplifying the tests?

spydon: >  Would giving up that separation have any other benefits besides simplifying the tests?

Nope, don't think so. We can keep it like this, it's probably not that many extra lines anyways.

packages/app_center/lib/src/deb/local_deb_model.dart
d-loose: I was hoping we could use the `LocalDebModel` as an example for the other classes we want to refactor, so I'd prefer using `freezed` here if we plan to do it for the other states as well.
I'll have another look at https://github.com/rrousselGit/freezed/issues/533 and check if there's something I'm missing here.

d-loose: It looks like the analyzer expands typedefs of the form `typedef Foo = Bar`, while keeping the names for function types like `typedef Foo = String Function()` :shrug:
I'll use `@immutable` here, as you suggested then

d-loose: Apparently `freezed` handles that case correctly. So I guess the best solution is to use `freezed` and just live with the inconvenient `PackageKitDetailsEvent` type name

d-loose:
I've started implementing @anasereijo's Figma designs here.
@spydon please complain if this PR is getting too large, I can break it down a bit if necessary.

@local-optimum screenshots:

Screenshot from 2024-06-12 17-47-59

Screenshot from 2024-06-12 17-48-03

d-loose:
UI changes after receiving some feedback from @anasereijo:

* Removed placeholder icon entirely (debian package metadata doesn't provide an icon)
* Moved the warning above the install button
* Made the 'Description' section non-collapsible since it's the only section on this page
* De-emphasized the 'Install' action in the dialog

Screenshot from 2024-06-13 14-22-47

Screenshot from 2024-06-13 14-22-53

packages/app_center/integration_test/app_center_test.dart
```diff
@@ -23,6 +25,35 @@ void main() {
     await testInstallSnap(tester, installedFile: file);
     await testRemoveSnap(tester, installedFile: file);
   });
+
+  testWidgets('install a local debian package', (tester) async {
+    await app.main([
+      '--local-deb',
+      'integration_test/assets/appcenter-testdeb_1.0_amd64.deb'
```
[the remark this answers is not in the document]
d-loose: Yes, good point. Currently the integration test runs directly in the github runner (which is amd64 by default). We could probably run it in a docker container with a specific architecture, though.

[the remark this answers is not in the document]
Feichtmeier: Prbly Not a blocker and I think Linux arm runners are available for open source repos end of the year

packages/app_center/lib/src/deb/deb_providers.dart
```diff
@@ -0,0 +1,13 @@
...
+import 'package:ubuntu_service/ubuntu_service.dart';
+
+part 'deb_providers.g.dart';
+
+@riverpod
+Stream<PackageKitTransaction> transaction(TransactionRef ref, int id) {
+  final transaction = getService<PackageKitService>().getTransaction(id);
+  if (transaction == null) return const Stream.empty();
+
+  return transaction.propertiesChanged.asyncMap((_) => transaction);
```
spydon, line 12: This one returns the same transaction as initially fetched every time a property of that transaction changes? :thinking:

spydon: Oooh it's a mutable map holding those ones, that map should maybe be turned into a freezed class at some point so that we get some better type safety.

packages/app_center/lib/src/deb/local_deb_model.dart
```diff
@@ -0,0 +1,59 @@
...
+
+@riverpod
+class LocalDebModel extends _$LocalDebModel {
+  @override
+  Future<LocalDebData> build({required String path}) async {
+    final packageKit = getService<PackageKitService>();
+    await packageKit.activateService();
+    final details = await packageKit.getDetailsLocal(path);
+    if (details == null) {
+      throw StateError('Failed to get package details');
```
spydon, line 32: Hmm, I don't think this is a `StateError`

packages/app_center/lib/src/deb/local_deb_model.dart
```diff
@@ -0,0 +1,59 @@
...
+    await packageKit.waitTransaction(activeTransactionId);
+    ref.invalidateSelf();
+  }
+
+  Future<void> cancel() async {
+    assert(state.asData?.value.activeTransactionId != null,
+        'cancel() called without active transaction');
+    final packageKit = getService<PackageKitService>();
+    await packageKit
+        .cancelTransaction(state.asData!.value.activeTransactionId!);
```
spydon, line 53:
```suggestion
        .cancelTransaction(state.value!.activeTransactionId!);
```
I don't think you need `asData` here?

packages/app_center/lib/src/deb/local_deb_page.dart
```diff
@@ -0,0 +1,222 @@
...
+  Widget build(BuildContext context) {
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.start,
+      children: [
+        Text(
+          AppLocalizations.of(context).snapPageDescriptionLabel,
+          style: Theme.of(context)
+              .textTheme
+              .titleMedium
+              ?.copyWith(fontWeight: FontWeight.w500),
```
spydon, line 81: Should this be in the theme maybe? Or should all `titleMedium` have this font weight?

packages/app_center/lib/src/widgets/app_title.dart
```diff
@@ -50,6 +51,7 @@ class AppTitle extends StatelessWidget {
 
   final String title;
   final String? publisher;
+  final bool showPublisher;
   final bool verifiedPublisher;
   final bool starredPublisher;
```
spydon, line 56: This must be a typo right? Not from this PR though

packages/app_center/lib/src/widgets/dialogs.dart
```diff
@@ -1,81 +1,98 @@
...
-  return showDialog(
-    context: context,
-    builder: (context) => ErrorDialog(
-      title: title,
-      message: message,
-      icon: icon,
-    ),
-  );
-}
+}) =>
```
spydon, line 22: I would use a body here and break out `titleMedium` to a variable

packages/app_center/test/initial_route_test.dart
```diff
@@ -57,6 +57,24 @@ void main() {
...
+
+    final listener = MockInitialRouteListener();
+    container.listen<String?>(
+      initialRouteProvider,
+      listener.call,
+      fireImmediately: true,
+    );
+
+    verify(listener(
+            null, StoreRoutes.namedLocalDeb(path: '/path/to/local.deb')))
```
spydon, line 74: I would add a trailing comma here to make it easier to read

spydon (commented):
Good stuff!

packages/app_center/lib/src/deb/deb_providers.dart
d-loose: Yes, didn't want to change things in packagekit.dart for this PR, but it's a bit ugly, I agree. Here I've just moved the old `transactionProvider` into it's own file and used a riverpod annotation.

packages/app_center/lib/src/deb/local_deb_model.dart
d-loose: Right, blame it on copilot :)
Can this be a simple `Exception` for now? I'd like to improve error handling in general next pulse (there are a lot of bare `ErrorWidget` in some places). With the new app level error stream in place, we should take inventory of all the exceptions we propagate up and wrap them in meaningful error messages.

packages/app_center/lib/src/deb/local_deb_page.dart
d-loose: I've already added this to my ever-growing list of discrepancies between Yaru and the design team's Figma library. It should be fixed in either of those.
We have a bunch of those customizations in other places as well. I'd keep it for now, since we're planning to do a UI cleanup with Ana soon anyways

packages/app_center/lib/src/widgets/app_title.dart
d-loose: What do you mean? I don't see a typo here

spydon (approved):

samvde:
I honestly think this is a big mistake. Regardless the warnings, people such as my family members can now easily install software outside of the known Ubuntu ecosystem. Why would you allow that?

Experts can  just use apt install or gdebi for it. If one can't figure that out, perhaps installing software from other sources is not the best idea for them.

I hope there will at least be a setting to block this feature. Can you comment on that? If not, I could of course block sudo for these users, but then they can't install anything from the software center either.

Feichtmeier:
> I honestly think this is a big mistake. Regardless the warnings, people such as my family members can now easily install software outside of the known Ubuntu ecosystem. Why would you allow that?
> 
> Experts can  just use apt install or gdebi for it. If one can't figure that out, perhaps installing software from other sources is not the best idea for them.
> 
> I hope there will at least be a setting to block this feature. Can you comment on that? If not, I could of course block sudo for these users, but then they can't install anything from the software center either.

There's already an issue which contains a lot of discussion. Please keep discussions unrelated to code review out of pull requests

PR #1698, opened 2024-06-14
refactor: Cleaner file structure

spydon:
* This moves away from the `src` directory style, since that is to be used in packages, not in apps (since there are no dependents of an app).
* Since the root was a mix of barrel files and classes the barrel files moved in to their respective directories.
* The vote class was named `user.dart` in the ratings package, renamed it to `vote.dart`.


packages/app_center/l10n.yaml
```diff
@@ -1,4 +1,4 @@
-arb-dir: lib/src/l10n
+arb-dir: lib/l10n
```
spydon, line 1: @d-loose will this need a setup change in weblate?

d-loose: Yes :melting_face:

spydon: Moved it back in [ac7fcb5](https://github.com/ubuntu/app-center/pull/1698/commits/ac7fcb5120b3e118e1d54648fc27f8ea2434720c)

d-loose (approved):
Thanks, LGTM :+1:

PR #1710, opened 2024-06-20
feat: add ErrorObserver

d-loose:
Adds a `ProviderObserver` that logs provider build errors. This mostly concerns exceptions thrown due to connection errors (ratings backend or snapd unavailable), but potentially other errors we're not aware of that might be useful for debugging.


spydon (approved):
Nice!

PR #1715, opened 2024-06-25
chore: Update to latest snapd client

spydon:
This is still using a git dependency for snapd.dart, but it will be published as soon as it is verified that all the types work.


packages/app_center/test/app_card_test.dart
```diff
@@ -14,10 +13,17 @@ const snapRating = Rating(
...
   name: 'testsnap',
   id: 'r4LxMVp7zWramXsJQAKdamxy6TAWlaDD',
   summary: 'Its a summary!',
+  revision: '',
+  version: '',
+  channel: '',
+  base: '',
+  contact: '',
+  description: '',
+  type: '',
```
d-loose, line 26: Those aren't needed here as they correspond to the defaults in `createSnap`

d-loose:
Works fine :+1:
Just wondering if it's a good idea to mark some additional fields in `Snap` as required (e.g. `contact` and `description`). Maybe it's better to keep them optional and provide empty defaults? I'll have a closer look at the fields in the corresponding snapd.dart PR

packages/app_center/test/games_page_test.dart
```diff
@@ -25,9 +25,9 @@ void main() {
     const SnapSearchParameters(query: 'testsn'): [
       createSnap(name: 'testsnap', title: 'Test Snap', downloadSize: 3),
       createSnap(
-          name: 'testsnap2', title: 'Another Test Snap', downloadSize: 1),
+          name: 'testsnap2', title: 'Another Test Snap', downloadSize: 1,),
```
d-loose, line 28: Could you run `melos format`?
Seems like the CI doesn't complain about this. I think we used to run `dart format --set-exit-if-changed` before switching to the melos action.

spydon: Whaaat, I'll fix that, not much sense in running formatting in the pipeline if it doesn't complain :smile:

d-loose (approved):

packages/app_center/test/games_page_test.dart
spydon: Added to check in https://github.com/ubuntu/app-center/pull/1716

Issue #1684, opened 2024-06-03
Missing error handling for 'SnapdException: too many requests on ubuntu 24.04'


aimanabdul:
![Screenshot from 2024-06-03 14-38-06](https://github.com/ubuntu/app-center/assets/45210967/38b0203a-d1bf-41af-94ef-d36fef2630b3)

d-loose:
Thanks for your report!
It seems like there was a temporary outage in the snap ecosystem, which should be resolved by now.
We clearly need better error handling here, though!

d-loose:
Related (error handling): #1668

rgborn:
This is still happening, I just got the error, SnapdException(kind: too many requests), with the same screen as above only I was doing a search.

FemLolStudio:
Same error here too.

d-loose:
Will be fixed by #1723

Issue #1668, opened 2024-05-08
App Center (Ubuntu 24.04) not working behind corporate proxy


Langohr23:
Hello together,

I have setup Ubuntu 24.04. at my company PC. The network needs a corporate proxy (non intercepting, just a forwarding proxy). Everything works fine (APT in the terminal, other applications in the terminal, Browsers etc.) except the App Center.

I have configured our proxy in the following locations:
1. In the operating systems Network -> Proxy tap. First via a WPAD file, but I also tried it with a manuel configuration.
2. For APT in : /etc/apt/apt.conf.d/ (http, ftp and https)
3. For the terminal in "etc/enviornment" (HTTP and HTTPS)

I always get the following error in the App center:
![image](https://github.com/ubuntu/app-center/assets/33636936/838e3779-4324-4815-b571-3957dcfc42dd)

In the Logs I get the following error: 

SnapdException: SnapdException(kind: null, message: cannot list updates: Post "https://api.snapcraft.io/v2/snaps/refresh": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers))

Thank you for you help!

Thomas

d-loose:
Thanks for your report! Can you successfully interact with snapd on the command line? Are there any errors if you run `snap install hello` or `snap refresh`? If that's the case I'd point you in the direction of the [snapcraft forums](https://forum.snapcraft.io/) for support, as we don't have any influence on that in the frontend.

In any case I think we need a more graceful error handling here.

SergeyNovak777:
Hello, you can trying write proxy for snap
```
sudo snap set system proxy.http="your_proxy:port"
sudo snap set system proxy.https="your_proxy:port"
```
in my work pc:
```
sudo snap set system proxy.http="http://proxy.unn.ru:8080"
sudo snap set system proxy.https="http://proxy.unn.ru:8080"
```

d-loose:
Error handling of this issue is addressed in #1723

Issue #1479, opened 2023-11-01
Snapd exception shown when connection timeout


tim-hm:
When trying to browse the app center without an internet connection this error was surfaced. We should fail more gracefully here.

@anasereijo @d-loose 

<img width="1358" alt="SCR-20231101-spgu" src="https://github.com/ubuntu/app-center/assets/1335525/f30e06f1-5109-4318-a1cd-002ff4036db2">

dishankjindal1:
Hi @BLKKKBVSIK - I would like to pick up the issue and give it a try. Can you give me hint on where should I look for this issue? And how do I emulate it?

BLKKKBVSIK:
Hi @dishankjindal1 

You simply have to run the project without being connected to internet.
One good way to fix it, is to catch this error and display a generic text like "No connection or couldn't connect"
Don't hesitate if you need more instructions.

anasereijo:
@BLKKKBVSIK After getting connection back, will users need to reload the app, or will it update automatically?

anasereijo:
For copy, I think it should read - We can't load content in the App Center without an internet connection.

One cool thing we could do is to use the animated icon as suggested [here](https://github.com/ubuntu/app-center/issues/1432#issuecomment-1752160392) :)

Thanks for your help @dishankjindal1 !

BLKKKBVSIK:
> @BLKKKBVSIK After getting connection back, will users need to reload the app, or will it update automatically? 

For now, the user will need to reload the tab at least. But in the case where the device has no internet connection, we could refresh it automatically in a second time.

tim-hm:
When offline, we could show a modal or screen to prevent users from interacting with the app. This avoids the complexities of handling numerous corner cases that arise when users navigate the app without an internet connection. The modal could include a 'retry' button and perform periodic connectivity checks, effectively freezing the app until internet is available.

anasereijo:
Could we have two solutions for two different scenarios?

### a) user is using the app and then the internet disconnects:
I think we should show an error message within the page they had opened: 

We can't load content in the App Center without an internet connection. 
Check your connection and reload.
[Retry]

this way, once the connection is back, the user will see the page they were looking at before, or the page they were trying to open, so they don't lose the 'flow'.



### b) user is offline and tries to open the app
have @tim-hm's approach of having a modal preventing users from interacting with the app. the window will be empty (dark or light depending on the user's theme)

**Unable to connect to the App Center**
Check your connection and reload.
[Retry]

[...]

PR #1723, opened 2024-06-28
feat: add custom error view

d-loose:
@anasereijo we can go over design and copy for all of those next week:

Error message when refreshing a single snap or switching the channel while it is running:
Screenshot from 2024-06-27 15-41-46

Generic error page with custom message for `SnapdException(kind: 'network-timeout')` (see #1479 and #1668)
Screenshot from 2024-06-28 17-01-26

Generic error page with custom message for `SnapdException(message: 'too many requests')` (see #1684)
Screenshot from 2024-06-28 17-00-44

Ref #1479, #1668, #1684

UDENG-3056
UDENG-3264


d-loose:
Updated error pages matching the Figma designs:

Network error:
Screenshot from 2024-07-10 16-09-26

Server error (e.g. 'too many requests')
Screenshot from 2024-07-10 16-08-01

Unknown error:
Screenshot from 2024-07-10 16-07-44

'Retry' will retry the action prior to encountering the error (loading search results, displaying a snap page, ..). 'Check status' will open https://status.snapcraft.io/ in a browser.

spydon (approved):
Lgtm functionality-wise!
I'm sure Ana will want to have a say in the design for the generic ones later. 😄

d-loose:
> I'm sure Ana will want to have a say in the design for the generic ones later. 😄

I've double-checked Figma now and realized that I forgot to actually include the action texts for each error in the view. Added them now and updated the screenshots in my previous comment!

Feichtmeier:
Imho a centered, vertical layout could be better, especially for when the window size varies

 Image
   Text 
Buttons

spydon:
> Imho a centered, vertical layout could be better, especially for when the window size varies
> 
> ```
>  Image
>   Text 
> Buttons
> ```

I agree, it feels a bit "off" now, let's check that with Ana :)

madsrh:
> Imho a centered, vertical layout could be better, especially for when the window size varies

+1 for centered layout 👍 

> I've double-checked Figma now... 

Obviously, us mortals can't see the Figma mockups, but shouldn't the "headline" have a different styling? I know that it is slightly larger, but it's barely visible and doesn't feel like a propper heading. Maybe just use the same weight as we use on the [Update page](https://private-user-images.githubusercontent.com/31812582/325912976-fcac94eb-ab52-4fc7-a8b0-0a5153e4632c.png)?

image

anasereijo:
It does indeed look better centered 👍 
The heading looks almost the same as the text, so if that can be changed, go for it. 

The spacing between elements looks odd - the first line of text looks closer to the heading than to the second line of text. I suggest adding more padding between the heading and the text, and decrease (or remove) the padding between the two lines of text.

d-loose:
Column layout with adjusted spacing and font size:

Screenshot from 2024-07-11 11-46-13
Screenshot from 2024-07-11 11-45-49
Screenshot from 2024-07-11 11-44-31

anasereijo (approved):

PR #1726, opened 2024-07-02
fix: Move back button to the title bar

spydon:
This moves up the back button to the title bar.

image

Supersedes: #1664
Fixes: #1727

Follow up: Move flag and share icons to be at the same height as the title


d-loose (approved):
Really nice, thanks!

PR #1731, opened 2024-07-04
fix: styling improvements

### LINKED ISSUE #1558: Manage page: button content aligment is not consistent

### LINKED ISSUE #1657: ManagePage: update buttons padding need to be bigger
The horizontal padding is not there and the vertical padding is not enough. They should be as big as the rest of the buttons elsewhere 

Bildschirmfoto vom 2024-04-26 19-40-41

Meta: in the long run, better use Material Buttons with yaru.dart instead of the home cooked PushButton

### LINKED ISSUE #1719: borders missing from update available
When there is only one update available, it displays that one 'row' without borders. We should keep the borders even if it's a single row.

how it looks like:
VirtualBox_ubuntu 23 10_26_06_2024_18_43_33

how it should look like (note that this is just a mock-up, not following yaru precisely!):
only one update

Feichtmeier:
|                                                   | before| after| 
| ---| ---| ---| 
| explore page category button borderradius |grafik|grafik| 
|manage page open button padding               |grafik|grafik|
|search field alignment                                    |grafik|grafik|
|overflow|grafik|grafik|
|manage: single snap missing border|grafik|grafik|

manage page update button spinner
before

https://github.com/ubuntu/app-center/assets/15329494/ab21b7f3-c785-41b9-95a9-00ec1631fc48

after

https://github.com/ubuntu/app-center/assets/15329494/a1e8bfe7-0cac-4e98-9260-a49758713d90

Closes #1657
Closes #1558
Closes #1719

### DIFF
--- a/packages/app_center/lib/constants.dart
+++ b/packages/app_center/lib/constants.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:yaru/icons.dart';
 
 const kAppName = 'App Center';
 const kGitHubRepo = 'ubuntu/app-center';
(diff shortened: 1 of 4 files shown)


spydon (approved):
Lgtm, we should add all "magic number" margins and paddings as constants at some point

PR #1733, opened 2024-07-09
Add launcher for "updates"

sergio-costas:
This patch adds an extra, invisible, .desktop file that opens the store in the "updates" page directly. This is needed because in the new Refresh App Awareness, we want to allow the user to open the snap-center in the "updates" page directly from a "pending update" notification, but since it is `snapd-desktop-integration` the application that opens it, and it is inside another snap, the only way that it has to launch the `snap-center` is using the `desktop-launch` DBus interface. This interface, unfortunately, doesn't allow to pass parameters, only allows to pass the name of a `.desktop` file to launch, so this workaround is needed.

### DIFF
--- a/packages/app_center/assets/app-center-updates.desktop
+++ b/packages/app_center/assets/app-center-updates.desktop
@@ -1,4 +1,7 @@
 [Desktop Entry]
+# This desktop entry is required to allow Snapd Desktop Integration
+# to open the Snap Center in the "updates" page when the user
+# clicks "Show updates" in a notification. DO NOT DELETE.
 Type=Application
 Version=1.0
 Exec=snap-store --updates %U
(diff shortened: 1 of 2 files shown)


seb128 (changes requested):
Could you add a comment to the .desktop explaining the context? That .desktop doesn't make much since it's not visible nor claiming at mimetype, by reading side chats it seems to be to workaround Ubuntu core desktop limitations? If so it would be nice to state so it will not confuse futur contributors who might think it's noise and has no point and decide to clean it up

seb128 (changes requested):
sorry to bother you there, but could you also drop a small note in the .desktop itself to avoid having someone wanting to remove it without thinking about going to the git log to figure out the rational...

seb128 (approved):
Thanks

PR #1734, opened 2024-07-09
feat: add --updates flag to open manage page

d-loose:
Adds an `--updates` flag that can be used to immediately show the manage page when the app is launched.

The way App Center currently manages routes and pages is a bit messy, so this would place the manage page on top of the navigator stack in the right pane, instead of making it the initial selection in the menu on the left side. See screenshots for more details.
To make this work properly, we'd need to refactor the store navigation and routes quite a bit.

|Proper implementation|Quick implementation (this PR)|
|-|-|
|Screenshot from 2024-07-09 18-01-29|Screenshot from 2024-07-09 18-01-22|

Ref #1733
UDENG-3532


d-loose:
@spydon would you be fine with merging this as-is, to unblock @sergio-costas?

spydon (approved):

Issue #1676, opened 2024-05-24
Migrate to modern style providers


spydon:
This migrates away from the custom and intricate `ChangeNotifierProvider` approach.

It also makes sure that futures aren't completed until the polling of the `changeId` has returned which makes the the structure slightly cleaner.

Example: https://riverpod.dev/docs/migration/from_change_notifier

PR #1737, opened 2024-07-11
refactor: Migrate manage snaps providers

### LINKED ISSUE #1551: Apps installed locally are showing grey screen after delete
[Screencast from 2023-12-29 12-41-52.webm](https://github.com/ubuntu/app-center/assets/72045785/466a2d43-e987-4681-ad00-73cc421b3e9f)

spydon:
This migrates the providers for the manage page to the modern style (#1676) it also fixes #1551.

If there is no internet connection, we currently show this:
Screenshot from 2024-07-16 19-43-54
And if you click check for updates and the connection is back, it will be going back to its normal state.


packages/app_center/lib/snapd/snap_page.dart
```diff
@@ -41,7 +41,8 @@ class SnapPage extends ConsumerWidget {
     if (snapDataNotFound) {
       WidgetsBinding.instance.addPostFrameCallback((_) {
         if (Navigator.canPop(context)) {
-          ref.invalidate(localSnapsProvider);
+          print('we invalidate');
```
d-loose, line 44: :smile:

d-loose:
Looks really good at first glance! I'll add some more detailed comments later. One thing I noticed so far is that the manage page won't work without a network connection (and even the tile in the sidebar breaks)

image

d-loose:
Another nitpick: changing the filters for the list of installed snaps now shows a loading spinner for a very short time, causing a bit of a flicker

spydon:
> Another nitpick: changing the filters for the list of installed snaps now shows a loading spinner for a very short time, causing a bit of a flicker

Good catch! Fixed in [5c4e9a9](https://github.com/ubuntu/app-center/pull/1737/commits/5c4e9a9d8281ec4f4ded18ef6c4dc1ef7912f16c)

packages/app_center/lib/snapd/snap_model.dart
```diff
@@ -34,6 +37,11 @@ class SnapModel extends _$SnapModel {
       unawaited(_listenUntilDone(activeChangeId, snapName, ref));
     }
 
+    if (localSnap == null && storeSnap == null) {
+      throw SnapDataNotFoundException(snapName);
```
d-loose, line 41: Just a remark: in principle this case should now be excluded by [this](https://github.com/ubuntu/app-center/pull/1723/files#diff-280431a536b0b0642fb7b42758289afb55109e60df38c183c56c5b1125ab5264L29-R28) change from the error handling PR, but I think it's good to treat this explicitly here anyways.

d-loose: Ah, I suppose this is handling the very specific case of #1551

packages/app_center/lib/snapd/snap_page.dart
```diff
@@ -34,15 +35,22 @@ class SnapPage extends ConsumerWidget {
...
+    final snapDataNotFound =
+        snap.hasError && snap.error is SnapDataNotFoundException;
+    if (snapDataNotFound) {
+      WidgetsBinding.instance.addPostFrameCallback((_) {
+        if (Navigator.canPop(context)) {
+          ref.invalidate(filteredLocalSnapsProvider);
+          Navigator.pop(context);
+        }
+      });
+    }
```
d-loose, line 48: This feels a bit hacky, but I guess going down the rabbit hole of checking for error before navigating to the page, like we do in the installer :mage:  is a bit overkill here..

packages/app_center/lib/manage/local_snap_providers.dart
```diff
@@ -20,16 +20,30 @@ class FilteredLocalSnaps extends _$FilteredLocalSnaps {
...
-        .where(
-          (snap) => snap.titleOrName
-              .toLowerCase()
-              .contains(ref.watch(localSnapFilterProvider).toLowerCase()),
-        )
+    void refreshFunction(_, __) => _refreshWithFilters(nonRefreshableSnaps);
+    ref.listen(localSnapFilterProvider, refreshFunction);
+    ref.listen(showLocalSystemAppsProvider, refreshFunction);
+    ref.listen(localSnapSortOrderProvider, refreshFunction);
+    return _refreshWithFilters(nonRefreshableSnaps, updateState: false);
```
d-loose, line 27: Haven't seen this before, seems like a nice solution! Is this a commonly used riverpod pattern?

packages/app_center/test/manage_page_test.dart
```diff
@@ -246,23 +224,27 @@ void main() {
         const SnapdTask(id: '', progress: SnapdTaskProgress(done: 1, total: 4)),
       ],
     );
+
+    await resetAllServices();
```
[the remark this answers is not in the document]
spydon: This one is used to reset the ones that are in setup, since there are custom ones added in this test

packages/app_center/lib/snapd/snap_model.dart
spydon: Yeah, exactly

packages/app_center/lib/snapd/snap_page.dart
spydon: Yeah, I thought the same

packages/app_center/lib/manage/local_snap_providers.dart
spydon: I'm not sure how common it is, I couldn't think of another way to not go into the loading state if all of these were `watch`ed

packages/app_center/lib/manage/manage_model.dart
```diff
@@ -1,63 +1,22 @@
...
-  }
-
-  Future<void> _getInstalledSnaps() async {
-    _installedSnaps = await snapd.getSnaps().then(
-          (snaps) => snaps.sortedBy((snap) => snap.titleOrName.toLowerCase()),
-        );
+  Future<ManageSnapsData> build() async {
+    final installedSnaps = await _snapd.getSnaps();
+    final refreshableSnapNames =
+        (await ref.watch(updatesModelProvider.future)).snapNames;
```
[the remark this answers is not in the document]
spydon: Yeah, I forgot about this, I had them separately first, but there was some problem with the slivers.
I'll try that again!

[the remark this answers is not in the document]
d-loose: With the latest changes the loading state behavior is still a bit different

[Screencast from 2024-07-15 17-01-24.webm](https://github.com/user-attachments/assets/fa19636a-94cb-463c-a61e-8b1cedac0f25)

I'm wondering what's the best option here - do we need to hide the list of installed snaps during the loading state?
@anasereijo do you have an opinion on that?

packages/app_center/lib/snapd/snap_model.dart
```diff
@@ -23,7 +23,7 @@ class SnapModel extends _$SnapModel {
     } on SnapdException catch (e) {
       // Since the snap is just not installed when 'snap-not-found is thrown we
       // can ignore this exception.
-      if (e.kind != 'snap-not-found') rethrow;
+      if (e.kind != 'snap-not-found' && e.kind != null) rethrow;
```
d-loose, line 26: Why is this change needed?

packages/app_center/lib/error/error_l10n.dart
```diff
@@ -10,7 +10,7 @@ sealed class ErrorMessage {
   const ErrorMessage();
 
   factory ErrorMessage.fromObject(Object? e) {
-    if (e is! SnapdException) return ErrorMessageUnkown();
+    if (e is! SnapdException) return ErrorMessageUnknown();
```
d-loose, line 13: Thanks for fixing my typos :D

packages/app_center/lib/snapd/updates_model.dart
```diff
@@ -1,76 +1,57 @@
...
-      } on SnapdException catch (e) {
-        _handleError(e);
-      }
-    });
+  @override
+  Future<Iterable<Snap>> build() {
+    try {
+      return _snapd.find(filter: SnapFindFilter.refresh);
+    } on SnapdException catch (_) {
+      // TODO: Should we ignore all SnapdExceptions here?
```
d-loose, line 32: No, with the new error handling setup, we probably shouldn't catch any errors at all here, and let them be handled by the custom error view (like [here](https://github.com/ubuntu/app-center/blob/e3dd562c29a7bc1cd6f3b2a6082d3097624f00a6/packages/app_center/lib/snapd/snap_page.dart#L48-L51)) or the top-level error dialog

spydon: Then it won't work when we don't have internet though, will it?

d-loose: Ah very good point - I'm obviously contradicting myself here :D
But we need to somehow inform the user that we couldn't fetch updates due to network issues.

packages/app_center/lib/snapd/snap_model.dart
```diff
@@ -21,9 +21,17 @@ class SnapModel extends _$SnapModel {
...
     } on SnapdException catch (e) {
-      // Since the snap is just not installed when 'snap-not-found is thrown we
-      // can ignore this exception.
-      if (e.kind != 'snap-not-found') rethrow;
+      switch (e.kind) {
+        // Since the snap is just not installed when 'snap-not-found is thrown
+        // we can ignore this exception.
+        case 'snap-not-found':
+        // When kind is null it is most likely a problem with the internet
+        // connection.
```
d-loose, line 30: That's not generally true, see https://github.com/ubuntu/app-center/issues/1684, which is a case that we'd need to handle differently I think.

spydon: Aaah, how annoying, if we expose `statusCode` and maybe `status` in the SnapdClient maybe we can solve this in a better way.

d-loose (approved):
Thanks, LGTM now :+1:
We should leave further improvements concerning error handling and offline behavior to future PRs :slightly_smiling_face:

PR #1740, opened 2024-07-12
feat: Add rated category snaps to the Games tab

ashuntu:
This PR adds an ordered list of snaps based on their rating to the Games tab. Doing this required adding support for the top charts endpoints from the ratings service to the ratings client.

I've added this as a new variation of the `CategorySnapList`, so hopefully if we intend to incorporate top charts into other areas of the app, we can easily reuse the `RatedCategorySnapList`. A spinner is displayed while the list is loading; it does take a moment since we have to fetch the snap details for each snap ID.

This is more or less what it looks like, though the actual snaps displayed will not be the same:
image

UDENG-3351

### DIFF
--- a/packages/app_center/lib/games/games_page.dart
+++ b/packages/app_center/lib/games/games_page.dart
@@ -39,8 +39,16 @@ class GamesPage extends ConsumerWidget {
             const SizedBox(height: kPagePadding),
           ],
         ),
-        const CategorySnapList(
-          category: SnapCategoryEnum.games,
+        const RatedCategorySnapList(
+          categories: {
+            SnapCategoryEnum.games,
+            SnapCategoryEnum.kdeGames,
+            SnapCategoryEnum.gnomeGames,
+            SnapCategoryEnum.gameLaunchers,
+            SnapCategoryEnum.gameEmulators,
+            SnapCategoryEnum.gameContentCreation,
+            SnapCategoryEnum.gameDev,
+          },
         ),
         SliverList.list(
           children: [
--- a/packages/app_center/lib/games/games_page.dart
+++ b/packages/app_center/lib/games/games_page.dart
@@ -40,15 +40,7 @@ class GamesPage extends ConsumerWidget {
           ],
         ),
         const RatedCategorySnapList(
-          categories: [
-            SnapCategoryEnum.games,
-            SnapCategoryEnum.kdeGames,
-            SnapCategoryEnum.gnomeGames,
-            SnapCategoryEnum.gameLaunchers,
-            SnapCategoryEnum.gameEmulators,
-            SnapCategoryEnum.gameContentCreation,
-            SnapCategoryEnum.gameDev,
-          ],
+          categories: [SnapCategoryEnum.games],
         ),
         SliverList.list(
           children: [
--- a/packages/app_center/lib/games/games_page.dart
+++ b/packages/app_center/lib/games/games_page.dart
@@ -40,7 +40,7 @@ class GamesPage extends ConsumerWidget {
           ],
         ),
         const RatedCategorySnapList(
-          categories: {
+          categories: [
             SnapCategoryEnum.games,
             SnapCategoryEnum.kdeGames,
             SnapCategoryEnum.gnomeGames,
@@ -48,7 +48,7 @@ class GamesPage extends ConsumerWidget {
             SnapCategoryEnum.gameEmulators,
             SnapCategoryEnum.gameContentCreation,
             SnapCategoryEnum.gameDev,
-          },
+          ],
         ),
         SliverList.list(
           children: [
--- a/packages/app_center/lib/ratings/ratings_service.dart
+++ b/packages/app_center/lib/ratings/ratings_service.dart
@@ -37,42 +37,10 @@ class RatingsService {
 
   Future<List<ratings.ChartData>> getChart(SnapCategoryEnum category) async {
     await _ensureValidToken();
-
-    final categoryToChart = {
-      SnapCategoryEnum.artAndDesign: ratings.Category.ART_AND_DESIGN,
-      SnapCategoryEnum.booksAndReference: ratings.Category.BOOK_AND_REFERENCE,
-      SnapCategoryEnum.development: ratings.Category.DEVELOPMENT,
-      SnapCategoryEnum.devicesAndIot: ratings.Category.DEVICES_AND_IOT,
-      SnapCategoryEnum.education: ratings.Category.EDUCATION,
-      SnapCategoryEnum.entertainment: ratings.Category.ENTERTAINMENT,
-      SnapCategoryEnum.featured: ratings.Category.FEATURED,
-      SnapCategoryEnum.finance: ratings.Category.FINANCE,
-      SnapCategoryEnum.gameContentCreation: ratings.Category.GAMES,
-      SnapCategoryEnum.gameDev: ratings.Category.GAMES,
-      SnapCategoryEnum.gameEmulators: ratings.Category.GAMES,
-      SnapCategoryEnum.gameLaunchers: ratings.Category.GAMES,
-      SnapCategoryEnum.games: ratings.Category.GAMES,
-      SnapCategoryEnum.gnomeGames: ratings.Category.GAMES,
-      SnapCategoryEnum.healthAndFitness: ratings.Category.HEALTH_AND_FITNESS,
-      SnapCategoryEnum.kdeGames: ratings.Category.GAMES,
-      SnapCategoryEnum.musicAndAudio: ratings.Category.MUSIC_AND_AUDIO,
-      SnapCategoryEnum.newsAndWeather: ratings.Category.NEWS_AND_WEATHER,
-      SnapCategoryEnum.personalisation: ratings.Category.PERSONALISATION,
-      SnapCategoryEnum.photoAndVideo: ratings.Category.PHOTO_AND_VIDEO,
-      SnapCategoryEnum.productivity: ratings.Category.PRODUCTIVITY,
-      SnapCategoryEnum.science: ratings.Category.SCIENCE,
-      SnapCategoryEnum.security: ratings.Category.SECURITY,
-      SnapCategoryEnum.serverAndCloud: ratings.Category.SERVER_AND_CLOUD,
-      SnapCategoryEnum.social: ratings.Category.SOCIAL,
-      // SnapCategoryEnum.ubuntuDesktop: 0, // no valid mapping
-      // SnapCategoryEnum.unknown: 0, // no valid mapping
-      SnapCategoryEnum.utilities: ratings.Category.UTILITIES,
-    };
-
     return client.getChart(
       ratings.Timeframe.unspecified,
       _jwt!,
-      categoryToChart[category],
+      category.toProto(),
     );
   }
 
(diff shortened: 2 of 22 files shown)


anasereijo:
This is looking nice! 

- It would be helpful if you could share a screenshot of how it looks next to other sections, such as app cards, etc. Is this something you can share here?

- The way I designed it originally was to fit 4 cols (in the wide window) to add some interesting layout to the page. So the cards are 3 columns, and this is 4. See the screenshot below (grid in pale pink)

Screenshot 2024-08-13 at 16 56 53

- I suppose that the titles (Top rated, etc) are the same size as the ones in the Explore page, right?

- The numbers seem too big. Can you decrease the font size?


Hope this makes sense :)

ashuntu:
I decreased the font size of the numbers from 24 to 16. Here are some additional screenshots (with the font size change):
image
image

And this is what it looks like fullscreen (I have a 1440p 16:10 display):
image

I had some trouble fitting in 4 apps in a row without compromising the titles/descriptions, but if you'd like, I can take another look at doing that.

The titles are the same size as the other page headings, yes.

anasereijo:
Thanks @ashuntu, the smaller numbers are looking better.

If you have the capacity, I think it is worth trying to fit the charts in 4 cols on the widest screen. Looking at the screenshot seems that you have a 'card' around the app when on hover. That isn't necessary (the whole area should be clickable though) and maybe it will save up some space to fit the 4 cols? It will also decrease the padding between the number and the icon, I think.

Screenshot 2024-08-16 at 09 44 49

ashuntu:
How does this look? (this still includes an outline on hover)

image

anasereijo:
I think this could work, thanks for trying @ashuntu!

- What is the padding between each column? It should match the padding used in the other sections.
- We don't need the outline on hover. It can be just clickable but without the outline

Thank you!

ashuntu:
I remove the border on hover now.

For the columns, they should have the same padding as the other cards (they use the same grid, just have different column count and aspect ratio). 

I think the numbers are slightly out of the grid in the same way that the other titles (like the F in "Featured Titles" in the screenshot you showed) are. This happens on all the pages it looks like, where the first letter in titles is slightly out of bounds. If I had to guess, that's just due to the font, but I'm not sure. Maybe someone else from the apps squad would know better than me on that.

anasereijo:
Thanks, @ashuntu! I can bring that up with the squad, as it's something not really related then to this PR. 

That said, LGTM! thanks for making the changes.

packages/app_center/lib/widgets/snap_grid.dart
```diff
@@ -24,6 +25,21 @@ class AppCardGrid extends StatelessWidget {
...
+  factory AppCardGrid.fromRatedSnaps({
+    required List<Snap> snaps,
+    required ValueChanged<Snap> onTap,
+  }) =>
+      AppCardGrid(
+        appCards: snaps.asMap().entries.map(
+              (entry) => AppCard.fromRatedSnap(
+                snap: entry.value,
+                onTap: () => onTap(entry.value),
+                rating: entry.key + 1,
```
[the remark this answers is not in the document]
ashuntu: The key in this is just from the `.asMap` call, so I can get the index and give each snap a number next to it based on the index (its rank). Is there a different way you would do it?

d-loose (commented):
Could you please rebase the PR? Let me know if you need any help!

packages/app_center/lib/ratings/ratings_service.dart
```diff
@@ -32,6 +34,11 @@ class RatingsService {
     return client.getRating(snapId, _jwt!);
   }
 
+  Future<List<ChartData>> getChart(SnapCategoryEnum category) async {
+    await _ensureValidToken();
+    return client.getChart(Timeframe.unspecified, _jwt!, category.index);
```
d-loose, line 39: The entries of `SnapCategoryEnum` and `Category` aren't identical, we'll need to to an explicit conversion between the two and can't simply use the index here

packages/app_center_ratings_client/lib/src/ratings_client.dart
```diff
@@ -53,8 +53,15 @@ class RatingsClient {
...
-  Future<List<ChartData>> getChart(Timeframe timeframe, String token) async {
-    final request = chart_pb.GetChartRequest(timeframe: timeframe.toDTO());
+  Future<List<ChartData>> getChart(
+    Timeframe timeframe,
+    String token, [
+    int? category,
+  ]) async {
+    final request = chart_pb.GetChartRequest(
+      timeframe: timeframe.toDTO(),
+      category: category != null ? chart_pb.Category.valueOf(category) : null,
```
d-loose, line 63: See previous comment, we need to map the categories explicitly between the internal representation and the grpc definition.

d-loose (changes requested):
Did some testing with the staging environment @matthew-hagemann managed to deploy now :)
There's a small bug due to mismatching category definitions, see comments for details

packages/app_center/lib/widgets/app_card.dart
```diff
@@ -73,30 +89,58 @@ class AppCard extends StatelessWidget {
...
+        : YaruBanner(
+            padding: const EdgeInsets.all(kCardSpacing),
+            onTap: onTap,
+            color: Theme.of(context).cardColor,
+            child: Flex(
+              direction: compact ? Axis.vertical : Axis.horizontal,
+              crossAxisAlignment: CrossAxisAlignment.start,
+              children: appContent,
+            ),
+          );
```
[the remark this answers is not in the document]
ashuntu: I was thinking of doing that originally, but the cards are used in `snap_grid` and those take types of `AppCard`. Is there a way you'd recommend splitting this up? Maybe a separate grid entirely for rated grids?

ashuntu:
@d-loose I rebased and made most of the changes you pointed out.

I added the explicit mapping for chart categories, let me know if you think there is a better way than the way I chose in https://github.com/ubuntu/app-center/pull/1740/commits/73e1b4e691606a783608b3cd2c6626c2c563049c.

packages/app_center/lib/widgets/app_card.dart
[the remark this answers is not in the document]
d-loose: Yes, I think a separate grid would be clearer, I'd prefer some mild duplication over increasingly complex layout logic in the widgets

packages/app_center/lib/ratings/ratings_service.dart
```diff
@@ -29,17 +30,53 @@ class RatingsService {
...
     await _ensureValidToken();
     return client.getRating(snapId, _jwt!);
   }
 
-  Future<List<ChartData>> getChart(SnapCategoryEnum category) async {
+  Future<List<ratings.ChartData>> getChart(SnapCategoryEnum category) async {
     await _ensureValidToken();
-    return client.getChart(Timeframe.unspecified, _jwt!, category.index);
+
+    final categoryToChart = {
```
d-loose, line 41: I like doing these conversions with explicit `switch` expressions to ensure the enum is matched exhaustively, see e.g. [here](https://github.com/canonical/prompting-client/blob/02e97f3b22c339cbadd67e9977778997aee0ae30/flutter_packages/prompting_client/lib/src/prompting_client.dart#L50-L67). We'd need two separate `toProto` and `fromProto` conversions here and make sure we handle our custom UI categories properly in each of them

packages/app_center/lib/widgets/small_banner.dart
```diff
@@ -0,0 +1,41 @@
...
+  final VoidCallback? onTap;
+  final Widget child;
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+    final borderRadius = BorderRadius.circular(kYaruBannerRadius);
+    final light = theme.brightness == Brightness.light;
+    final defaultSurfaceTintColor =
+        light ? theme.cardColor : const Color.fromARGB(255, 126, 126, 126);
```
d-loose, line 17: Could we also remove the hardcoded color here? Unless absolutely necessary, it'd be great if we could avoid maintaining customizations on top of Yaru altogether.
We can probably remove the explicit `theme.cardColors` here entirely, now that https://github.com/ubuntu/app-center/pull/1818 is merged.

For reference, this is what it currently looks like:

| light | dark |
| - | - |
| Screenshot from 2024-10-16 13-39-55 | image |

packages/app_center/lib/ratings/ratings_service.dart
ashuntu: I don't recall anywhere we need to convert from the ratings category to the enum category (would be `fromProto`), do you know if that conversion is actually needed anywhere?

packages/app_center/lib/widgets/small_banner.dart
ashuntu: Yeah, removing the `surfaceTintColor` entirely seems to retain the correct colors anyway.

packages/app_center/lib/games/games_page.dart
```diff
@@ -39,8 +39,16 @@ class GamesPage extends ConsumerWidget {
...
-          category: SnapCategoryEnum.games,
+        const RatedCategorySnapList(
+          categories: [
+            SnapCategoryEnum.games,
+            SnapCategoryEnum.kdeGames,
+            SnapCategoryEnum.gnomeGames,
+            SnapCategoryEnum.gameLaunchers,
+            SnapCategoryEnum.gameEmulators,
+            SnapCategoryEnum.gameContentCreation,
+            SnapCategoryEnum.gameDev,
```
d-loose, line 50: Since all of those map to `Category.GAMES` now, I think we only need to include `SnapCategoryEnum.games` here.

packages/app_center/lib/widgets/small_banner.dart
d-loose: A small UI nitpick I'm noticing now: the card with rank `10` is misaligned with the others, so we should make sure the rank label has a fixed width.

packages/app_center/lib/snapd/snapd_service.dart
```diff
@@ -1,8 +1,23 @@
...
+  Future<Snap?> findById(String snapId) async {
+    final queryParams = {
+      'series': '16',
+      'remote': 'true',
+      'snap-id': snapId,
+    };
+    final result =
+        await getAssertions(assertion: 'snap-declaration', params: queryParams);
+    final declaration = SnapDeclaration.fromJson(result);
+    final findResult = await find(name: declaration.snapName);
```
d-loose, line 19: Sorry, another nitpick: while testing locally I noticed this can throw an exception if the snap id can't be resolved (they're randomly generated in the tests, so almost always invalid). That usually shouldn't be an issue in production, but doesn't hurt to add a check here :)

d-loose (commented):
Thanks, those are great improvements! I have two more tiny nitpicks and would be happy to merge this soon :)

@sminez, @matthew-hagemann should we wait until the ratings service with support for charts is deployed before merging this? Otherwise we'd still need to add some UI fallback state that hides the top charts in case the service doesn't support the feature.

packages/app_center/lib/ratings/ratings_service.dart
d-loose: Looks like we don't need it

ashuntu:
> Otherwise we'd still need to add some UI fallback state that hides the top charts in case the service doesn't support the feature.

I added fallback to the old behavior (just a regular `CategorySnapList`) when the ratings service for charts isn't available https://github.com/ubuntu/app-center/pull/1740/commits/23375de704a8d3d5e434c660085b8a2c09643eb2.

d-loose (approved):
Awesome, thanks! I'd still wait until tomorrow morning (European time) to merge this, if that's alright :)

ashuntu:
I added a try/catch here instead for the assertions endpoint to handle the YAML errors we talked about. So for now the rated grids will just ignore any snap with invalid YAML assertions until a raw response is sorted in snapd.dart.

d-loose:
@sminez I've pushed a `gaming-top-charts` branch (currently the same as `main`) and changed the base of this PR to target that one, so we can land additional changes before merging to `main` without bothering @ashuntu too much :)

sminez (approved):

sminez:
@d-loose thanks for re-pointing the PR to the new branch :ok_hand: please merge at your leisure :sunglasses:

PR #1747, opened 2024-07-17
feat: Uninstall snaps from the manage page

### LINKED ISSUE #1745: Add uninstall option to all snaps directly on the manage page
Currently you have to click on a snap to install it, but it would be much more useful if you can uninstall directly from the manage page.

spydon:
[Screencast from 2024-07-17 21-45-32.webm](https://github.com/user-attachments/assets/85ab5ee3-f764-4474-b27e-63fbb1259356)

Closes #1745

### DIFF
--- a/packages/app_center/lib/manage/local_snap_providers.dart
+++ b/packages/app_center/lib/manage/local_snap_providers.dart
@@ -33,6 +33,18 @@ class FilteredLocalSnaps extends _$FilteredLocalSnaps {
     );
   }
 
+  /// Used to remove a snap from the list without reloading the whole provider.
+  /// Should be used when a snap is uninstalled directly from the manage page
+  /// list for example.
+  void remove(String snapName) {
+    if (!state.hasValue) return;
+    state = AsyncData(
+      state.value!.copyWith(
+        snaps: state.value!.snaps.where((s) => s.name != snapName),
+      ),
+    );
+  }
+
   Iterable<Snap> _refreshWithFilters(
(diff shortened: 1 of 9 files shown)


packages/app_center/lib/manage/local_snap_providers.dart
```diff
@@ -33,6 +33,18 @@ class FilteredLocalSnaps extends _$FilteredLocalSnaps {
     );
   }
 
+  /// Used to remove a snap from the list without reloading the whole provider.
+  /// Should be used when a snap is uninstalled directly from the manage page
+  /// list for example.
+  void remove(String snapName) {
```
d-loose, line 39: Nitpick: I'd make the method name more explicit in this case - `removeFromList`

packages/app_center/lib/manage/updates_model.dart
```diff
@@ -50,6 +50,18 @@ class UpdatesModel extends _$UpdatesModel {
     return result;
   }
 
+  /// Used to remove a snap from the list without reloading the whole provider.
+  /// Should be used when a snap is uninstalled directly from the manage page
+  /// list for example.
+  void remove(String snapName) {
```
d-loose, line 56: Same as above

d-loose (approved):
Nice, LGTM :+1:

packages/app_center/lib/manage/local_snap_providers.dart
spydon: True, otherwise it could probably be confused with an uninstallation

PR #1749, opened 2024-07-18
snap: update to core24

jssotomdz:
this pull request migrates the app center to core24
migration seems very straightforward only requiring the change in the base parameter and adding git as build parameter (how did it get it before by the way?)

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -2,7 +2,7 @@ name: snap-store
 summary: App Center
 description: App Center
 confinement: strict
-base: core22
+base: core24
 grade: stable
 version: git
 
@@ -18,6 +18,7 @@ parts:
       - clang
       - cmake
       - curl
+      - git
       - libgtk-3-dev
       - ninja-build
       - unzip


d-loose:
> this pull request migrates the app center to core24 migration seems very straightforward only requiring the change in the base parameter and adding git as build parameter (how did it get it before by the way?)

Hm interesting, it wasn't needed for core22, I don't know why though. Maybe @kenvandine knows

d-loose (approved):
LGTM :+1:

kenvandine:
> > this pull request migrates the app center to core24 migration seems very straightforward only requiring the change in the base parameter and adding git as build parameter (how did it get it before by the way?)
> 
> Hm interesting, it wasn't needed for core22, I don't know why though. Maybe @kenvandine knows

Sounds like a snapcraft bug.  Snapcraft included git as a build-package when sources came from git automatically.

d-loose:
> Sounds like a snapcraft bug. Snapcraft included git as a build-package when sources came from git automatically.

The source-type isn't set here, so maybe something changed in the detection mechanism?

MrSIHAB:
> > this pull request migrates the app center to core24 migration seems very straightforward only requiring the change in the base parameter and adding git as build parameter (how did it get it before by the way?)
> 
> Hm interesting, it wasn't needed for core22, I don't know why though. Maybe @kenvandine knows

actually it was built on ubuntu 22.04 and probably core 24 hadn't that much dependency or plugin then.

PR #1751, opened 2024-07-24
feat: App Center update info about itself

spydon:
image

### DIFF
--- a/packages/app_center/lib/manage/manage_page.dart
+++ b/packages/app_center/lib/manage/manage_page.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:app_center/constants.dart';
 import 'package:app_center/error/error.dart';
 import 'package:app_center/l10n.dart';
@@ -53,7 +55,7 @@ class ManagePage extends ConsumerWidget {
                 l10n.managePageDebUpdatesMessage,
                 style: textTheme.titleMedium,
               ),
-              const SizedBox(height: 48),
+              _SelfUpdateInfoBox(),
               ...updatesModel.when(
                 data: (snapListState) {
                   final refreshableSnaps = snapListState.snaps;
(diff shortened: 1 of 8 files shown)


packages/app_center/lib/manage/manage_page.dart
```diff
@@ -336,3 +338,48 @@ class _ActionButtons extends ConsumerWidget {
...
+          spacing: 16,
+          runSpacing: 16,
+          children: [
+            Text(l10n.managePageOwnUpdateDescription),
+            PushButton.outlined(
+              child: Text(
+                l10n.managePageOwnUpdateQuitButton,
+                overflow: TextOverflow.ellipsis,
+              ),
+              onPressed: () => exit(0),
```
spydon, line 376: @d-loose is this enough?

d-loose: I think it'd be cleaner to use `YaruWindow.of(context).close()` like we do in the installer. That way it's also testable

packages/app_center/lib/manage/manage_page.dart
```diff
@@ -336,3 +338,48 @@ class _ActionButtons extends ConsumerWidget {
...
+
+class _SelfUpdateInfoBox extends ConsumerWidget {
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final l10n = AppLocalizations.of(context);
+    final refreshInhibitModel = ref.watch(
+      refreshInhibitSnapsProvider.select(
+        (value) => value.whenData(
+          //(data) => data.firstWhere((s) => s.name == 'snap-store'),
+          (data) => data.firstWhere((s) => s.name == 'element-desktop'),
```
d-loose, line 350: That's probably left-over testing code :)

spydon: Haha, oooops!

Feichtmeier:
I would make the outlined button inside have the same outline as the info box

spydon:
> Small design feedback: I would make the outlined button inside have the same outline as the info box

I guess maybe it's actually the background that should change, at least according to the design:
image

spydon:
@anasereijo what do you think about this button placement instead? It's the result of a long discussion on the community discord, I think it looks pretty good!
image

d-loose (approved):
Thanks, LGTM :+1:

P.S. would be great to have a test case for that :)

spydon:
> Thanks, LGTM 👍
> 
> P.S. would be great to have a test case for that :)

True, added!

Issue #1752, opened 2024-07-24
Many loading icon when app still loading installed and to be updated apps


alculete:
![image](https://github.com/user-attachments/assets/a02e9c85-727b-43a4-901b-7769667e6f08)

PR #1754, opened 2024-07-25
fix: Avoid double loader for updates available section

spydon:
[Screencast from 2024-07-25 15-33-20.webm](https://github.com/user-attachments/assets/4407889d-1319-45b0-9206-1f363a3b1a72)

Also fixed so that the height doesn't change when the loaded switches to the text, which looked kind of janky.

Closes: #1752


d-loose (approved):
Nice!

PR #1763, opened 2024-08-09
feat: Show individual refresh statuses when refreshing all snaps

spydon:
Closes: #1767


sminez (approved):
Looks good to me :+1:

PR #1777, opened 2024-08-13
chore: Use fvm for handling the Flutter version

spydon:
This also automatically fixes the broken integration test.


spydon:
This was the issue: https://github.com/flutter/flutter/issues/153083

scripts/install-fvm.sh
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env bash
+
+# Function to log messages with date and time
+log_message() {
+    echo -e "$1"
```
sminez, line 5: The comment implies that the date and time should be getting injected here.

sminez (commented):
The fvm side of things all looks good to me. From the PR description I'm assuming the integration test issue is related to the changes of `MaterialStatePropertyAll` to `WidgetStatePropertyAll` and `ButtonBar` to `OverflowBar`, but is that also the case for removing the `filterQuality` parameter in `app_icon.dart`?

scripts/install-fvm.sh
spydon: I haven't written this script, I just copied it from here so that we don't curl things from uncontrolled sources: 
https://fvm.app/documentation/getting-started/installation

spydon:
> The fvm side of things all looks good to me. From the PR description I'm assuming the integration test issue is related to the changes of `MaterialStatePropertyAll` to `WidgetStatePropertyAll` and `ButtonBar` to `OverflowBar`, but is that also the case for removing the `filterQuality` parameter in `app_icon.dart`?

It's due to that this PR also upgrades to 3.24.0, and those things were deprecated there (or the default filter quality changed).

sminez (approved):

packages/app_center/.fvmrc
d-loose, line 1: does every package necessarily need its own `.fvmrc`?

snap/snapcraft.yaml
```diff
@@ -11,21 +11,11 @@ parts:
...
-      # necessary when reruns using the same container
-      if [ -d "flutter" ]; then
-        rm -rf "flutter"
-      fi
-
       craftctl default
-      FLUTTER_VERSION=$(sed -n "s/^flutter \([0-9.]\+\).*/\1/p" .tool-versions)
-      git clone -b $FLUTTER_VERSION --depth 1 https://github.com/flutter/flutter.git
+      bash $CRAFT_PART_SRC/scripts/install-fvm.sh
+      fvm install
```
d-loose, line 16: This shouldn't be part of the `pull` step, but rather be in `override-build`. I think we can completely omit `override-pull`.

snap/snapcraft.yaml
```diff
@@ -43,17 +33,18 @@ parts:
     plugin: nil
     override-build: |
       set -eux
-      dart pub global activate melos
+      fvm install
```
d-loose, line 36: We shouldn't have to run `fvm install` again here

d-loose (commented):
I've left some additional questions/comments since I was OOO when this got merged

packages/app_center/.fvmrc
spydon: I think that they do for like `fvm flutter run` etc to work, I have it symlinked to the repo root currently

d-loose: ah, thanks that makes sense - didn't notice that since I've setup my editor to use flutter from fvm's sdk

spydon: Yeah, we should include that in the contributors doc, if I didn't already do that...

PR #1778, opened 2024-08-14
fix: Escape markdown links and images

spydon:


packages/app_center/lib/extensions/string_extensions.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:flutter/widgets.dart';
+
+extension StringExtension on String {
+  String escapedMarkdown() {
+    final imagePattern = RegExp(r'!\[.*?\]\(.*?\)');
```
sminez, line 5: Does `flutter_markdown` allow for anchor text in links to span multiple lines (see example below)? If it does then we probably need these to be `!?\[(.|\n)*?\]\(.*?\)`.

[this is
a multi-line anchor for a link
to this pull request](https://github.com/ubuntu/app-center/pull/1778)

packages/app_center/test/string_extensions_test.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:app_center/extensions/string_extensions.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  test('escapedMarkdown', () {
+    const input =
+        'A great image ![image](https://i.imgur.com/LlOSp3I.gif) and a great [link](https://en.wikipedia.org/wiki/List_of_lists_of_lists), right?';
+    const expected =
+        'A great image \\![image](https://i.imgur.com/LlOSp3I.gif) and a great \\[link](https://en.wikipedia.org/wiki/List_of_lists_of_lists), right?';
```
sminez, line 9: The escaped markdown still renders in github. Does `flutter_markdown` handle this differently?

\\image
\\[link](https://en.wikipedia.org/wiki/List_of_lists_of_lists)


I can only get it to work here with a single slash before the first bracket (so the escaping of image links would need to be tweaked if this is the same in `flutter_markdown`)
!\[image](https://i.imgur.com/LlOSp3I.gif)
\[link](https://en.wikipedia.org/wiki/List_of_lists_of_lists)

spydon: It is only one slash, the first slash is for escaping the slash so that it doesn't become a control character (like `\n` etc) :)

packages/app_center/lib/extensions/string_extensions.dart
spydon: Good catch! I don't think it does, but I fixed it up anyways, if we should change markdown renderer in the future.

packages/app_center/test/string_extensions_test.dart
sminez: ... :facepalm: 

I'll go get more coffee :joy:

sminez (approved):

PR #1785, opened 2024-08-19
SearchBar autofocus when click on clear btn

MrSIHAB:
The search bar should automatically focused when 'clear search text' button is clicked. This is small tweak but usefull too.
I do search, do some activities. Then search bar become unfocused. Then I click on clear search button and start typing my query, And then I see that Search bar is not focused. I have to manually click on the textfeild to focus it. This pull request will automatically focus search bar when someone click on clear search button. I made this changes from my experience.

### DIFF
--- a/packages/app_center/lib/search/search_field.dart
+++ b/packages/app_center/lib/search/search_field.dart
@@ -188,7 +188,10 @@ class _SearchFieldState extends ConsumerState<SearchField> {
                       return YaruIconButton(
                         icon: const Icon(YaruIcons.edit_clear, size: 16),
                         onPressed:
-                            controller.text.isEmpty ? null : controller.clear,
+                            controller.text.isEmpty ? null : (){
+                              controller.clear();
+                              node.requestFocus();
+                            } 
                       );
                     },
                   ),


packages/app_center/lib/search/search_field.dart
```diff
@@ -188,7 +188,10 @@ class _SearchFieldState extends ConsumerState<SearchField> {
                       return YaruIconButton(
                         icon: const Icon(YaruIcons.edit_clear, size: 16),
                         onPressed:
-                            controller.text.isEmpty ? null : controller.clear,
+                            controller.text.isEmpty ? null : (){
+                              controller.clear();
+                              node.requestFocus();
+                            }
```
spydon, line 194: Should it not always request focus, even if the box is already cleared?

spydon (commented):
Thanks for your contribution, have you signed the CLA?
(If you have there might be some problems with the check, it has been a bit flaky lately)

spydon:
@MrSIHAB your commits are not connected to your e-mail, hence there is no way for the CLA check to check that you've signed it. You can update the commits to contain the e-mail, or if you want, since the PR is pretty small, I can take it over.

packages/app_center/lib/search/search_field.dart
MrSIHAB: I signed the CLA after making this pull request

MrSIHAB:
There may any technical problem, I'll try again

Issue #1772, opened 2024-08-12
Update error message when updating all snaps


spydon:
When updating all snaps it should follow the new copy for when snaps can't update and it shows an error message (different design for one and multiple).

Issue #1656, opened 2024-04-26
ManagePage: one failed update should not cancel all updates


Feichtmeier:
If one presses update all and one of the updates can not be made because the app is running, the update all should still happen with the ones that are updateable (if this is a word)

![Bildschirmfoto vom 2024-04-26 19-40-20](https://github.com/ubuntu/app-center/assets/15329494/fed5ad43-d359-43d4-8065-801731352ad5)

d-loose:
Related issues: #1391, #1643

d-loose:
Just a small update - the snapd API docs say:
> POST /v2/snaps
> ...
> Fields
> ...
> transaction: optional string field, defaulting to per-snap if not present. If all-snaps is used, the action field will be performed such that the corresponding change will have a single transaction covering all the snaps. If in this case there is a failure for any snap, the state of all affected snaps will be fully reverted, even if for some snap the action has been successful. This field is valid for install and refresh actions. See [Transactional updates](https://snapcraft.io/docs/transactional-updates) for more details

So in principle it should be possible to refresh multiple snaps via a single API call and errors that occur during the refresh of an individual snap shouldn't affect the other ones.
However, in practice that doesn't seem to work:

Example with two snaps that are up to date:
```
$ snap refresh steam signal-desktop
All snaps up to date.
```
Example with two snaps, where one is currently running:
```
$ snap refresh steam firefox
error: cannot refresh "steam", "firefox": snap "firefox" has running apps (firefox), pids:
       6167,...
```

The first command creates a snap change, as expected, but the second one doesn't:
```
$ snap changes
ID    Status  Spawn                      Ready                      Summary
2621  Done    today at 14:58 CEST        today at 14:58 CEST        Refresh snaps "steam", "signal-desktop": no updates
```

### Possible workarounds
* Passing an empty list of snaps to snapd (has the same effect as running `snap refresh`). In this case errors that occur while refreshing an individual snap will be silently ignored.
* Refreshing each snap individually (as previously done in the community store). Besides known issues with that (polkit authentication would be invoked multiple times, unless we wait until the first snap in the list is refreshed, and loop through the remaining ones afterwards), we'd need to handle each transaction separately to track their progress and watch for possible errors.

d-loose:
[...]

PR #1786, opened 2024-08-20
feat: Support individual snap updates statuses on refresh all

spydon:
image

Closes: #1772, #1656

### DIFF
--- a/packages/app_center/lib/manage/updates_model.dart
+++ b/packages/app_center/lib/manage/updates_model.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 
 import 'package:app_center/constants.dart';
+import 'package:app_center/error/error.dart';
 import 'package:app_center/providers/error_stream_provider.dart';
 import 'package:app_center/snapd/snapd.dart';
 import 'package:collection/collection.dart';
@@ -33,7 +34,7 @@ class SnapListState with _$SnapListState {
 }
 
 final currentlyRefreshAllSnapsProvider = StateProvider<List<String>>((_) => []);
-final isSilentlyiCheckingUpdatesProvider = StateProvider<bool>((_) => false);
+final isSilentlyCheckingUpdatesProvider = StateProvider<bool>((_) => false);
 
 @Riverpod(keepAlive: true)
 bool hasUpdate(HasUpdateRef ref, String snapName) {
@@ -63,7 +64,7 @@ class UpdatesModel extends _$UpdatesModel {
   }
 
   Future<void> silentUpdatesCheck() async {
-    ref.read(isSilentlyiCheckingUpdatesProvider.notifier).state = true;
+    ref.read(isSilentlyCheckingUpdatesProvider.notifier).state = true;
     try {
       final newSnapListState = await fetchRefreshableSnaps();
       final isSameList = const ListEquality().equals(
@@ -75,7 +76,7 @@ class UpdatesModel extends _$UpdatesModel {
         state = AsyncData(newSnapListState);
       }
     } finally {
-      ref.read(isSilentlyiCheckingUpdatesProvider.notifier).state = false;
+      ref.read(isSilentlyCheckingUpdatesProvider.notifier).state = false;
     }
   }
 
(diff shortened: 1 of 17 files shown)


packages/app_center/lib/manage/updates_model.dart
```diff
@@ -101,21 +102,53 @@ class UpdatesModel extends _$UpdatesModel {
...
-      ref.read(errorStreamControllerProvider).add(e);
+      if (e.kind != 'auth-cancelled') {
+        ref.read(errorStreamControllerProvider).add(e);
+      }
+    }
+    if (errors.isNotEmpty) {
+      ref.read(errorStreamControllerProvider).add(
+            errors.length == 1
+                ? errors.values.first
+                : ConsolidatedSnapdException(errors),
```
d-loose, line 150: I think we should propagate the `ConsolidatedSnapdException` in either case (and maybe even give it a more specific name, since it's treated differently in the error message logic.
With only a single failed update it looks like this:

[Screencast from 2024-08-21 11-47-24.webm](https://github.com/user-attachments/assets/c9c21a4a-3168-4f7c-9180-259623c2b46f)

but I guess we'd want to show the 'Some updates failed' dialog instead

packages/app_center/lib/widgets/dialogs.dart
```diff
@@ -28,7 +28,7 @@ Future<void> showErrorDialog({
     actions: [
       DialogAction(
         value: null,
-        label: UbuntuLocalizations.of(context).closeLabel,
+        label: AppLocalizations.of(context).okayLabel,
```
d-loose, line 31: There is `okLabel` in `ubuntu_localizations`, which defaults to "OK". Should we use that for consistency with other applications?

packages/app_center/lib/manage/updates_model.dart
spydon: Ah, that's a bug, I thought that I made sure that if it's a single it would just re-throw the existing exception.
My thought for `ConsolidatedSnapdException` was to possibly have it for more scenarios in the future where you can have multiple different errors for different snaps.

packages/app_center/lib/widgets/dialogs.dart
spydon: @anasereijo?

packages/app_center/lib/manage/updates_model.dart
d-loose: > Ah, that's a bug, I thought that I made sure that if it's a single it would just re-throw the existing exception.

It does exactly that, no? But that means the error message shown in the dialog will look like a generic snapd exception, instead of what you're showing in the screenshot above.

> My thought for `ConsolidatedSnapdException` was to possibly have it for more scenarios in the future where you can have multiple different errors for different snaps.

I see, but currently it's being treated very specifically:
```dart
...
        ErrorMessageConsolidated() => l10n.managePageUpdatesFailed(
            (this as ErrorMessageConsolidated).errors.length,
          ),
...
```

spydon: > It does exactly that, no? But that means the error message shown in the dialog will look like a generic snapd exception, instead of what you're showing in the screenshot above.

It doesn't show any exception above? Since you created `ErrorMessageRunningApps` it should be remodeled to that if it only gets the original exception right?

> I see, but currently it's being treated very specifically:

Yeah, since we don't have any messaging for other errors with multiple exceptions yet I thought we could just continue building on this once we do.

spydon: Oooh, it didn't play the whole video for me apparently, now I saw the error message!

spydon: I'll just have to update `ErrorMessageRunningApps` then to accommodate for the new title

spydon: Updated in https://github.com/ubuntu/app-center/pull/1786/commits/0f3186258a810239d9667a95da7a97dbd369115d

packages/app_center/lib/widgets/dialogs.dart
anasereijo: good spot! That makes sense @d-loose

packages/app_center/lib/src/l10n/app_en.arb
```diff
@@ -82,13 +82,29 @@
...
     "managePageUpdatesAvailable": "Updates available ({n})",
-    "@managePageUpdatesAvaialble": {
+    "@managePageUpdatesAvailable": {
         "placeholders": {
             "n": {
                 "type": "int"
             }
         }
     },
+    "managePageUpdatesFailed": "{n, plural, =1{{n} One update} other{Some updates}} failed",
```
d-loose, line 92: For one update this shows "1 One Update" and for more than one it shows "Some updates".
I'd suggest

```suggestion
    "managePageUpdatesFailed": "{n, plural, =1{One update} other{{n} updates}} failed",
```
so that it shows "One update" and e.g. "3 updates"

d-loose (approved):
Besides one final nitpick, LGTM - massive improvement!

P.S. As always, some tests would be very welcome :)

packages/app_center/lib/src/l10n/app_en.arb
anasereijo: if we can tell the user how many failed even better! thanks for the proposal @d-loose

Issue #1782, opened 2024-08-18
Manage button shows as manag… (as if part of it doesn’t fit the GUI)


antonispgs:
This happens on every 24.04 fresh install I tried on different machines and on a VM. It only happens after the first snap refresh that updates app center to the latest version and persists after switching channel to the edge version as well. See this picture for reference.
https://imgur.com/a/5LZdVp1

PR #1787, opened 2024-08-21
fix: Update manage page label

spydon:
image

Fixes: #1782

### DIFF
--- a/packages/app_center/lib/src/l10n/app_en.arb
+++ b/packages/app_center/lib/src/l10n/app_en.arb
@@ -43,7 +43,7 @@
     "managePageDescription": "Check for available updates, update your apps, and manage the status of all your apps.",
     "managePageDebUpdatesMessage": "Debian package updates are handled by the Software Updater.",
     "managePageInstalledAndUpdatedLabel": "Installed and updated",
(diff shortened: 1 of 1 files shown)


sminez (approved):
do we know if there is currently a similar issue for any of the translations?

spydon:
> do we know if there is currently a similar issue for any of the translations?

Yes, pretty much all, but the translators will be notified that this has been updated now and will have to translate it again.

packages/app_center/lib/src/l10n/app_en.arb
```diff
@@ -43,7 +43,7 @@
     "managePageDescription": "Check for available updates, update your apps, and manage the status of all your apps.",
     "managePageDebUpdatesMessage": "Debian package updates are handled by the Software Updater.",
     "managePageInstalledAndUpdatedLabel": "Installed and updated",
-    "managePageLabel": "Manage installed Snaps",
+    "managePageLabel": "Manage",
```
d-loose, line 46: That was my fault, see https://github.com/ubuntu/app-center/pull/1720/files.
However, we need to introduce a separate l10n string that defaults to 'Manage installed Snaps', which is supposed to be the headline of the page, like here:

image

Issue #1769, opened 2024-08-12
Align numbers to the right in tables


spydon:
In the tables the numbers should be aligned to the right, for example the sizes.
So that the user can more easily see which number is bigger.

PR #1789, opened 2024-08-21
fix: Set alignment to right for sizes in tables

spydon:
image

Closes: #1769


packages/app_center/lib/manage/manage_snap_tile.dart
```diff
@@ -109,9 +109,11 @@ class ManageSnapTile extends StatelessWidget {
                         ),
                         maxLines: 1,
                         overflow: TextOverflow.ellipsis,
+                        textAlign: TextAlign.right,
```
d-loose, line 112: Opening another can of worms here, but maybe this is a good moment to think about RTL text alignment again? :D

image

spydon: Looks like there is a looot to do there? 😅
Like the menu surely shouldn't be on the right, right?
Did the line I changed look better before right aligning that row? Maybe we should just lock that row to LTR?

d-loose: Flutter/Material mirrors the entire layout for RTL directionality, see https://m2.material.io/design/usability/bidirectionality.html#mirroring-layout so I think that's correct. I think the only thing that's off is the alignment of the app titles and "updated n days ago" strings in the tiles. For the sizes I think we can simply use `TextAlign.end`.

Here's what it looks like with the `fa_IR` locale
Screenshot from 2024-08-22 11-13-58

I think the overall alignment should follow the RTL directionality in this case, even if there are LTR elements like the snap names

spydon: Good idea! Fixed in https://github.com/ubuntu/app-center/pull/1789/commits/9a7dedc55effdb2b633b205b8394dc00f441dc57

d-loose: Almost there - I think the last bit that needs changing is the `Align` in the `ListTile`'s title, it should be `AlignmentDirectional.centerStart` instead of `Alignment.centerLeft`. (Is this even needed at all?)

image

spydon: Right, so then I guess only the numbers will have to keep their "end" alignments then.

spydon: Fixed in https://github.com/ubuntu/app-center/pull/1789/commits/b01ab1c13d72fd538290f8caf0b709806aba08a6

d-loose (approved):
LGTM :+1:
(Beware of mock changes due to updated snapd.dart :))

Issue #1732, opened 2024-07-05
[Touch] Scrolling does not work


cm-t:
Hi,

On a Ubuntu (gnome-shell) 24.04 LTS installation (Tablet StarLite from starlabs). App-center up to date.

Expected:
 * I open an apps
 * I need to scroll
 * I use touchscreen to scroll
 * works

Result on app-center:
 * I open app-center
 * I browser categories, or search result; I need to scroll
 * I use touch screen (_1 finger like I do in gnome apps or firefox_)
 * does not work
     * Then I try 2 finger scroll
         *  still not works
         
Current user workaround:
 * put finger on the right side of the screen
 * scroll bar appears
 * drag and drop it
     * with small touch area of it, and its infinite scroll (_search result)_ it is not a good UX

Feichtmeier:
I think we need to add this to MaterialAPp

scrollBehavior: const MaterialScrollBehavior().copyWith(
        dragDevices: {
          PointerDeviceKind.mouse,
          PointerDeviceKind.touch,
          PointerDeviceKind.stylus,
          PointerDeviceKind.unknown,
          PointerDeviceKind.trackpad,
        },
      ),

I think we already have an issue for this somewhere

spydon:
Fixed in #1791

PR #1791, opened 2024-08-21
fix: Support scrolling on touchscreens

spydon:
I have no way to verify that this works, it simply uses the code suggested in #1732, so let's keep the issue open until someone with a touch screen can verify.


jssotomdz:
works as expected, tested with a surface laptop go

d-loose (approved):

PR #1792, opened 2024-08-22
fix: restore manage page title

d-loose:
See https://github.com/ubuntu/app-center/pull/1787#discussion_r1725242693


spydon (approved):
Ooops, good catch!

PR #1794, opened 2024-08-22
fix: Use non-synthetic translation files

spydon:
The synthetic translation system is unfortunately not stable at all together with the analyzer, so this PR refactors so that we use explicitly generated l10n files (like in the other projects).


snap/snapcraft.yaml
```diff
@@ -45,8 +45,9 @@ parts:
       
       fvm dart pub global run melos exec -c 1 --fail-fast --depends-on=build_runner -- \
       fvm dart run build_runner build --delete-conflicting-outputs
-
+
```
d-loose, line 48: nitpick: some leftover whitespace

d-loose (approved):
LGTM :+1:

snap/snapcraft.yaml
spydon: Haha, found some more whitespace in that file (most likely added by me previously...).
I thought my editor removed trailing whitespace automagically, but apparently not for yaml files.

snap/snapcraft.yaml
```diff
@@ -46,7 +46,11 @@ parts:
       fvm dart pub global run melos exec -c 1 --fail-fast --depends-on=build_runner -- \
       fvm dart run build_runner build --delete-conflicting-outputs
 
+      fvm dart pub global run melos exec -c 1 --fail-fast --depends-on=build_runner -- \
+      fvm dart run build_runner build --delete-conflicting-outputs
```
d-loose, line 50: I suppose this got accidentally duplicated when rebasing/merging

PR #1795, opened 2024-08-23
chore: Remove all generated files

spydon:
This also adds a CI step that generates the files and caches them for the other steps.


d-loose (approved):
LGTM :+1:

PR #1796, opened 2024-08-24
fix: SearchBar autofocus when click on clear btn

MrSIHAB:
The search bar should automatically focused when 'clear search text' button is clicked. This is small tweak but usefull too. I do search, do some activities. Then search bar become unfocused. Then I click on clear search button and start typing my query, And then I see that Search bar is not focused. I have to manually click on the textfeild to focus it. This pull request will automatically focus search bar when someone click on clear search button. I made this changes from my experience.

### DIFF
--- a/packages/app_center/lib/search/search_field.dart
+++ b/packages/app_center/lib/search/search_field.dart
@@ -187,11 +187,12 @@ class _SearchFieldState extends ConsumerState<SearchField> {
                     builder: (context, child) {
                       return YaruIconButton(
                         icon: const Icon(YaruIcons.edit_clear, size: 16),
-                        onPressed:
-                            controller.text.isEmpty ? null : (){
-                              controller.clear();
-                              node.requestFocus();
-                        },
+                        onPressed: controller.text.isEmpty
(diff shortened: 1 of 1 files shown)


packages/app_center/lib/search/search_field.dart
```diff
@@ -188,7 +188,10 @@ class _SearchFieldState extends ConsumerState<SearchField> {
                       return YaruIconButton(
                         icon: const Icon(YaruIcons.edit_clear, size: 16),
                         onPressed:
-                            controller.text.isEmpty ? null : controller.clear,
+                            controller.text.isEmpty ? null : (){
+                              controller.clear();
+                              node.requestFocus();
+                        },
```
spydon, line 194: I think it should always focus the field, even if there isn't any text in it.

spydon (commented):
Run `dart format` to format the code properly. :)

spydon (approved):
I fixed up the formatting, so now it looks shippable!

MrSIHAB:
Thanks. Next time I'll take care of it

PR #1799, opened 2024-08-27
feat: Clickable categories

spydon:
[Screencast from 2024-08-27 13-20-23.webm](https://github.com/user-attachments/assets/da42cef6-d8b3-430a-90bb-a2035991cb97)


packages/app_center/lib/widgets/app_title.dart
```diff
@@ -116,6 +117,49 @@ class AppTitle extends StatelessWidget {
   }
 }
 
+class _CategoryText extends StatefulWidget {
```
d-loose, line 120: Maybe it'd be worth putting [this](https://github.com/canonical/desktop-security-center/blob/fa951a7d2334f9bc2ba65d097d74330526fbc033/lib/widgets/markdown_text.dart) widget somewhere into `ubuntu_flutter_plugins`? It seems like we keep re-inventing these common things in various places using many different solutions :D

spydon: Yeah, that's a good idea!
I'll wait until I have some feedback from Ana first though. :)
Probably overkill to have something like this dependent on markdown.

anasereijo:
LGTM

packages/app_center/lib/widgets/app_title.dart
spydon: https://github.com/canonical/ubuntu-flutter-plugins/pull/428

d-loose (approved):
LGTM :+1:

PR #1801, opened 2024-08-28
fix: Move featured section to top of explore page

spydon:
image

UDENG-3749

### DIFF
--- a/packages/app_center/lib/explore/explore_page.dart
+++ b/packages/app_center/lib/explore/explore_page.dart
@@ -25,17 +25,6 @@ class ExplorePage extends ConsumerWidget {
       slivers: [
         SliverList.list(
           children: const [
-            SizedBox(height: kPagePadding),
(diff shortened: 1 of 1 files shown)


d-loose (approved):
:+1:

UDENG-3749

spydon:
> 👍

I'm adding the arts category too, but I guess it would actually cleaner to do that in a separate PR, so I'll merge this.

anasereijo:
can we have the following gradient as a background? I had a look at the page and the colours of banners and changed a few. These gradients should exist somewhere, I believe. 

Screenshot 2024-08-28 at 15 50 51

spydon:
> can we have the following gradient as a background? I had a look at the page and the colours of banners and changed a few. These gradients should exist somewhere, I believe.
> Screenshot 2024-08-28 at 15 50 51

I opened a new issue for that!

Issue #1775, opened 2024-08-12
Same dropdown menu look on manage page as on the snap page


spydon:
The manage page's dropdown menues for snaps are more condensed than the one on the snap page, they should look the same (like the snap page).

PR #1804, opened 2024-08-28
fix: Remove secondary action icons

spydon:
image

image


anasereijo:
The padding seems to be quite different in both dropdowns. The one on the app page seems to be wider than the one on manage. can this be matched?

spydon:
> The padding seems to be quite different in both dropdowns. The one on the app page seems to be wider than the one on manage. can this be matched?

Yup, that is something we have discussed before, we have #1775 open for that. :)

anasereijo:
LGTM then!

spydon:
> LGTM then!

If you put it as a review I can merge it 😉

d-loose (approved):

PR #1825, opened 2024-10-01
fix: add stubs for missing l10n files

d-loose:
It seems like we're missing quite a few translation files here, as we occasionally get requests to add individual languages through weblate. This PR adds stubs for the languages which are supported but still missing here.


spydon (approved):
Ah, good catch!

Issue #1513, opened 2023-11-29
'Change channel' shown in 2 lines


fthx:
Hi,

![image](https://github.com/ubuntu/app-center/assets/23138504/2949f84d-8b84-4354-b1c5-060d17a3e1ea)

Could be better in 1 line only, FR version here.

matthew-hagemann:
Hi @fthx, thanks for reporting this! 

Feels like instead of doing a text wrap and spanning it over two lines we should swap to cutting off lines with an ellipsis if we run out of screen space. (ie: `Changer de can...`) @anasereijo?

fthx:
Ellipsis does not look great, IMO.
![image](https://github.com/ubuntu/app-center/assets/23138504/070acfc7-a4fa-4e1f-a93d-1e129e6cefbf)
Here we have some place for a longer string, because reducing the width shows another layout:
![image](https://github.com/ubuntu/app-center/assets/23138504/0da367b4-c255-456a-943a-062470e015c2)
where there is a lot of place for change-channel string.

BLKKKBVSIK:
> Can we resize the button according to the text length?

We can, but on long text, it will break the UI on the smallest breakpoint. Or we can cut the text off with an ellipsis only on the smallest breakpoint.

anasereijo:
>  Or we can cut the text off with an ellipsis only on the smallest breakpoint.

That sounds good! I suppose there will be a tooltip on hover to tell the user the full label?

fthx:
Means that the label will not be fully displayed?
Sorry, I don't understand this regression here.

BLKKKBVSIK:
> Means that the label will not be fully displayed? Sorry, I don't understand this regression here.

The label will try to be displayed fully, but depending on the window size, there might be an ellipsis with a tooltip.

fthx:
Ok.
But the point I do not understand is that when the window's width is low, the "Change channel" label comes above the main info row.
![image](https://github.com/ubuntu/app-center/assets/23138504/cb56b6e8-3d7b-4676-8397-e2ceb06cd190)
So it seems there's not really a label's width issue?

fthx:
I don' t understand the eyes.

In fact, there is a wider probleme IMO here:
![image](https://github.com/ubuntu/app-center/assets/23138504/0509c286-8793-448c-8944-7138e5476930)
There are two labels (package, channel/open) those widths are not dynamically set and there is a lot of space left to not ellipsize any of them.

[...]

PR #1828, opened 2024-10-03
fix: Accommodate long switch channel buttons texts

spydon:
Screenshot from 2024-10-03 14-02-08
Screenshot from 2024-10-03 14-02-18

Instead of putting using the overflow bar we'll make the switch button flexible so that it can shrink when it needs to (and be larger when the text is long).

Fixes: #1513


anasereijo:
great improvement, thanks!

d-loose (approved):

Issue #1880, opened 2025-01-13
Add GNOME-style "type to search" function to find apps faster, align better with DE


jpgaubier:
### Is there an existing issue for this?

- [X] I have searched the existing issues and found none matching what I'm reporting.

### Is the feature related to a problem or existing issue?

GNOME apps allow users to open the app and start typing to search. App Center currently lacks this helpful feature.

### Describe the feature, why the feature matters to you, and what benefits the feature will bring about.

The feature would allow users to simply open the App Center and type for the app they're looking for, without navigating to it with a mouse. There might be some accessibility benefits on account of this. It would certainly help bring App Center more in line with the GNOME defaults that Ubuntu otherwise incorporates. It would help App Center "fit in" better with the rest of the DE.

PR #1574, opened 2024-01-16
feat: ctrl+f shortcut to focus search field

jssotomdz:
This pull request adds a keyboard shortcut to focus the search text field by clicking ctrl+f
For some reason the formatting (indenting) destroys the diff look. And kind of makes sense because I made the `YaruTheme` a child of the `CallbackShortcuts`. Is there a formatting helper or a guide for this? I could push directly to this branch to fix these code formatting issues if needed
Also, unit tests are failing because I changed the SearchField constructor signature so I have to rewrite the unit tests for them to succeed. I can do that in a separate commit if you're ok with it.

PR #1890, opened 2025-01-27
ux:Add GNOME-style type to search function to find apps faster

IldySilva:
references the #1880 

1.  Adding the feature and the test

### DIFF
--- a/packages/app_center/lib/store/store_app.dart
+++ b/packages/app_center/lib/store/store_app.dart
@@ -38,8 +38,15 @@ class StoreApp extends ConsumerStatefulWidget {
 }
 
 class _StoreAppState extends ConsumerState<StoreApp> {
+  @override
+  void initState() {
+    super.initState();
+    _focusScope.requestFocus();
+  }
+
   final _navigatorKey = GlobalKey<NavigatorState>();
   final searchFocus = FocusNode();
+  final _focusScope = FocusNode();
 
   NavigatorState get _navigator => _navigatorKey.currentState!;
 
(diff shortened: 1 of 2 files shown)


IldySilva:
@Feichtmeier  can you take a look?

Feichtmeier:
HI @IldySilva, thanks for your work. I need to forward this to @d-loose as I am not leading this project here anymore :)

IldySilva:
@IldySilva  can you check it?

d-loose:
Hey, thanks for your contribution. In principle I'm not opposed to focus the search bar when the user starts typing instead of having to use [ctrl+f](https://github.com/ubuntu/app-center/pull/1574) (@anasereijo, @juanruitina what do you think?). However, we'd need to be about more careful of capturing keyboard events - in its current state, this PR breaks the search field on the manage page as well as the 'report' dialog.

IldySilva:
I Understand, @d-loose  Could you please tell me how to reproduce this break?
I just saw it at some point while typing in 'report' and 'manager', but I can't reproduce it on purpose.

d-loose:
[Screencast From 2025-01-29 14-19-49.webm](https://github.com/user-attachments/assets/b14fd46e-e12a-4bfe-a213-f4090c03c2c2)

IldySilva:
@d-loose  any ideia about why the cla workflow still failing even after creating the launchpad account?

IldySilva:
@d-loose  why the  PR/Cla checks fail even after being registred on launchpad?

juanruitina:
Just to give my two cents on this: I think it makes sense to use GNOME-style search, that's a nice pattern that many users will expect; I see it all over. Let's just make sure indeed that it isn't triggered when typing on any input field, as flagged by @d-loose. It shouldn't work either when there's an overlay, for instance with the report modal opened (in this case, `Ctrl`+`F` shouldn't work either, that's a bug).

PR #1896, opened 2025-02-06
chore: disabling tics CI job until config is sorted

sminez:


d-loose (approved):
Unrelated CI failure: looks like we need to bump the version of the upload-artifact action

d-loose (approved):

PR #1897, opened 2025-02-12
maint: migrate to org GitHub to Jira application

didrocks:
We have a new GitHub to Jira global organization application. The previous gh workflow action is now deprecated.

UDENG-5950


d-loose:
Does this work for the ubuntu organization as well?

didrocks:
> Does this work for the ubuntu organization as well?

Yes! This is why we are transitionning now :)

The test project and issue was: https://github.com/ubuntu/GoWSL/issues/130

d-loose (approved):

PR #1906, opened 2025-03-25
ci: reenable tics action

d-loose:
This seems to work correctly now :)

The TiCS backend searches for coverage files in the `coverage` directory in the repository root. It seems like it actually only reads the lcov files, so the conversion to cobertura XML is unnecessary (for Dart code). Since the file paths in the coverage output are relative, they need to be adjusted when moving the output files around, so I'm just turning them into absolute paths.
I've also added a "cleanup" step, since symlinks in `.fvm` and `packages/*/linux/flutter/ephemeral` that point to directories outside of the repository, seem to cause problems with TiCS.

UDENG-5819

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -99,6 +99,18 @@ jobs:
         with:
           token: ${{secrets.CODECOV_TOKEN}}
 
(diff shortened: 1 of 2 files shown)


matthew-hagemann (approved):
Awesome!

PR #1909, opened 2025-03-27
fix: Contrast theme issues

ashuntu:
This PR should encompass most of the theme-only fixes for contrast issues. This does not include fixes that are required at a widget level.

Relevant accessibility issues:
- 1861761 (search contrast)
- 1860625 (version info)
- 1860587 (version info)

---

UDENG-6005


d-loose (approved):
Thanks!

PR #1926, opened 2025-06-03
fix(a11y): Add labels to page actions and expandable sections

ashuntu:
Adds accessible labels to the share and report buttons on snap pages, as well as the labels for expand/collapse sections. These are not visual labels, just text read by the screen reader. For example, it will read "Report this snap push button" when the report button receives focus.

This also adds the missing "Link copied" snack bar that appears when copying the share link from the snap page to the deb page. I think this fits in the PR since it addresses feedback of the button.

---

Audit 1859619
UDENG-7136

### DIFF
--- a/packages/app_center/lib/deb/deb_page.dart
+++ b/packages/app_center/lib/deb/deb_page.dart
@@ -8,6 +8,7 @@ import 'package:app_center/error/error.dart';
 import 'package:app_center/l10n.dart';
 import 'package:app_center/layout.dart';
 import 'package:app_center/packagekit/packagekit.dart';
+import 'package:app_center/store/store_app.dart';
 import 'package:app_center/widgets/widgets.dart';
 import 'package:appstream/appstream.dart';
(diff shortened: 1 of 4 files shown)


juanruitina:
I think "Report this snap" is good enough, yes. As for collapsing sections, I understand labels need to be unique, so I'd suggest to use "Expand [section name]", so "Expand gallery", "Collapse description".

juanruitina:
They should be consistent, so the slightly larger headings are OK as long as it's the same for both.

Minor tweaks for the labels (sorry, still figuring labels out):

* Instead of "Share this app", use "Copy link". It's more accurate, and it's implcit that it's for this app.
* Same with "Report this app", "Report" is sufficient.

ashuntu:
Related: https://github.com/ubuntu/yaru.dart/pull/999

This change could be adjusted to utilize the above when it is included in a Yaru release.

d-loose (approved):
Thanks :+1:

PR #1941, opened 2025-06-26
feat(a11y): Keyboard focusable hyperlinks

ashuntu:
Makes all text hyperlinks able to be focused via the keyboard. Previously, hyperlinks were all styled slightly differently and were never able to be focused or activated by the keyboard alone. This PR changes that by adding a unified `HyperlinkText` widget for all instances of hyperlinks.

Design notes:
- All hyperlinks now use Vanilla's blue hyperlink color (theme dependent)
- Hyperlinks have focus rings with the same corner radius as Yaru buttons, with a width of 2

<details>
<summary>Screenshots</summary>

About page non-focused
image
About page focused
image

---

Snap page non-focused
image
Snap page focused
image

</details>

---

Audit 1860919
UDEAA-45

### DIFF
--- a/packages/app_center/lib/about/about_page.dart
+++ b/packages/app_center/lib/about/about_page.dart
@@ -2,6 +2,7 @@ import 'package:app_center/about/about_providers.dart';
 import 'package:app_center/constants.dart';
 import 'package:app_center/l10n.dart';
 import 'package:app_center/layout.dart';
+import 'package:app_center/widgets/hyperlink_text.dart';
 import 'package:app_center/widgets/widgets.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_markdown/flutter_markdown.dart';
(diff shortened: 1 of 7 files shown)


ashuntu:
This doesn't adjust padding around the text/border, so the rounded border can be quite close to text. We could define a more smaller custom border radius for hyperlink focus rings instead of using the Yaru button radius.

@juanruitina how do you feel about this look?

juanruitina:
I see the border is inset, can we make show it outside the element? Maybe using [strokeAlignOutside](https://api.flutter.dev/flutter/painting/BorderSide/strokeAlign.html)?

I would say no border radius is needed: the focus ring should have the same shape as the element, but links don't have a shape, so. A 2px offset (akin to CSS `outline-offset`) would also be desirable, as per https://github.com/ubuntu/yaru.dart/issues/979, how easy would that be to implement? With offset, we could keep the border radius; without, we shouldn't so it doesn't affect text readability.

juanruitina:
Yup, `strokeAlign` of 2 seems to work: link is perfectly legible, legibility of text above is not compromised either. It does have a slight border radius (perhaps 1px outside), but I think it looks good, please leave it.

It would be great to bring this upstream.

PR #1946, opened 2025-07-03
fix(a11y): Fix empty publishers being announced

ashuntu:
With the new labels added by https://github.com/ubuntu/app-center/pull/1936, some publishers are sometimes announced despite the publisher being empty. This is because some Debian packages have an empty string instead of a null publisher. This fixes that by removing the publisher announcement entirely when there is none.

### DIFF
--- a/packages/app_center/lib/widgets/app_card.dart
+++ b/packages/app_center/lib/widgets/app_card.dart
@@ -80,7 +80,7 @@ class AppCard extends StatelessWidget {
     final l10n = AppLocalizations.of(context);
     final cardLabel = [
       '${title.title}.',
-      title.publisher != null
+      title.publisher != null && title.publisher != ''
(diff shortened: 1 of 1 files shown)


packages/app_center/lib/widgets/app_card.dart
```diff
@@ -80,7 +80,7 @@ class AppCard extends StatelessWidget {
     final l10n = AppLocalizations.of(context);
     final cardLabel = [
       '${title.title}.',
-      title.publisher != null
+      title.publisher != null && title.publisher != ''
```
d-loose, line 83: Nitpick: `isNotEmpty` might make this slightly more readable
```suggestion
      title.publisher?.isNotEmpty ?? false
```

ashuntu: Always forget about those utility fields, thanks :)

d-loose (approved):
Thanks!

PR #1952, opened 2025-07-10
fix(a11y): Add semantics container to search groups

ashuntu:
The items in the search combobox have no selected or group semantics, this adds a selected state to the search results as you arrow key through them and containerizes the search groups.

Note: this will probably sit in draft until `SemanticsRole.comboBox` is fully implemented in Flutter, and we can support announcements for selected elements. Needs https://github.com/flutter/flutter/issues/172023

---

Audit 1861859
UDEAA-137


ogres:
Is this still valid @ashuntu ?

ashuntu:
> Is this still valid @ashuntu ?

I believe the issue is still valid, yes. This PR would need some updates before its a valid fix, though.

PR #1987, opened 2025-10-26
Improve ErrorView & Error Message

Nexushunter:
# Problem

I noticed that there is an opportunity to improve coverage across the project. I couldn't find anything in the active pr list, so here's one.

# Solution
Target the `lib/error` of `app-center` since it has a low code surface area.

- Add tests for consolidated / bulk errors for the `ErrorMessage` types.
- Add tests for the `ErrorView` widget
  - capture subcomponent visibility expectations
  - capture action interactions
  - Cleaned up the scope of ownership for the widget by isolating the rendering of the actions for actionable errors.


packages/app_center/test/error/error_l10n_test.dart
Nexushunter: (Not new, file was moved from parent directory here)

ogres:
@Nexushunter Sorry for the late review, it seems that some files were changed meanwhile. Would you be able to update your PR? Thanks

PR #2009, opened 2026-01-18
Added StartupWMClass=snap-store to the .desktop file.

Kyuyrii:
Currently, when pinning the Snap Store to the KDE panel and opening it, it is not considered the same app that is pinned and uses the yellow icon that symbolizes Wayland.
Captura de imagem_20260118_131719

By modifying the .desktop file to add StartupWMClass=snap-store, the problem was resolved:
Captura de imagem_20260118_142802

### DIFF
--- a/packages/app_center/assets/app-center.desktop
+++ b/packages/app_center/assets/app-center.desktop
@@ -7,6 +7,7 @@ Terminal=false
 Categories=System;Utility;PackageManager;SoftwareManagement;Network;Settings;
 Keywords=Ubuntu;Applications;Apps;Store;Software;Snaps;
 MimeType=x-scheme-handler/snap;application/vnd.debian.binary-package;
+StartupWMClass=snap-store
 Name=App Center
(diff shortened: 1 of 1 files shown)


aleasto (commented):
LGTM. As for why this is necessary:

The traditional logic for matching a window with an application is for the client to identify itself with an app-id that must match either the .desktop filename (here mangled by snapd into `snap-store_snap-store`) or the `StartupWMClass` entry of the desktop file. Unless instructed otherwise, a GTK3 window will identify itself using the filename of the binary executable (here `snap-store`).

On top of that some desktops like GNOME Shell can recognize that the application is sandboxed, and will ignore the application-provided identifier (could be maliciously impersonating a different app) and always match it against its sandboxed identifier (always trusted, provided by the system): this is why GNOME can show an icon even without this change.

ogres (approved):

ogres:
Thank you for the contribution @Kyuyrii

PR #2032, opened 2026-03-08
fix: align card icon

### LINKED ISSUE #1471: AppIcons could be better aligned in cards
grafik

The AppIcons would look better if their size would be aligned to the title and subtitle.

ivoheck:
This PR improves the alignment of the AppIcon in app cards. The previous icon size and layout caused the icon to appear slightly misaligned relative to the title and publisher.

The icon size has been slightly reduced and unnecessary layout constraints were removed so the icon aligns more naturally with the text content.

Fixes #1471.

image


ivoheck:
Any thoughts on this? I'm happy to adjust the code if needed.

ogres:
@juanruitina What do you think about it? Is this acceptable?

juanruitina:
I appreciate the improved alignment, but also that it's slightly more compact. I wonder if the height of the card should also be reduced given the unused space at the bottom (I infer from the code that it's 170px high), and if the icons should be a more standard 48x48px (this PR seems to bring them down from 56 to 52px). @anasereijo what do you think?

PR #2034, opened 2026-03-12
Move runner into standard format

robert-ancell:
The current layout is the older format.


ogres:
@robert-ancell Could you resolve the merge conflict in `my_application.cc`? I'd like to have this merged if it simplifies things

robert-ancell:
@ogres should be good to land now.

PR #2040, opened 2026-03-15
fix: add localisation for app center label on about page

ivoheck:
Fixed the hardcoded title in about_page.dart. It now uses l10n.appCenterLabel instead of kAppName to support translations, as requested in #2026.

Fixes: #2026


ogres (approved):
Thanks for the contribution 👍

PR #2053, opened 2026-03-23
fix: parse channel parameter from snap:// URLs (#1982)

gauravsaini:
## Summary

- Properly parses `snap://` URLs using `Uri.parse` to extract snap name and query parameters
- Adds support for `channel` parameter in snap URLs (e.g., `snap://gimp?channel=preview/stable`)
- Fixes #1982

## Problem

When opening a `snap://` URL with a channel parameter from snapcraft.io (e.g., `snap://gimp?channel=preview/stable`), the entire string after `snap://` was being used as the snap name, causing "Something went wrong" error.

## Changes

- `store_providers.dart`: Use `Uri.parse` to properly extract snap name and channel parameter
- `store_routes.dart`: Add `channelOf` method and `channel` parameter to `namedSnap`
- `store_app.dart`: Pass channel from route to `SnapPage`
- `snap_page.dart`: Accept channel parameter and select it when snap data loads
- Added tests for channel parsing in routes and initial route

## Testing

- Unit tests verify channel parsing from routes
- Unit tests verify snap URL with channel is correctly converted to named route


brunvonlope:
Seems this superseeds https://github.com/ubuntu/app-center/pull/2021

gauravsaini:
Only if I could get somebody to review 😅
@ashuntu ?

packages/app_center/lib/snapd/snap_page.dart
```diff
@@ -48,11 +49,20 @@ class SnapPage extends ConsumerWidget {
...
-          return _SnapView(snapData: snapData);
-        },
-      ),
+      data: (snapData) {
+        if (channel != null &&
+            snapData.availableChannels?.containsKey(channel) == true &&
+            snapData.selectedChannel != channel) {
+          WidgetsBinding.instance.addPostFrameCallback((_) {
+            ref.read(snapModelProvider(snapName).notifier).selectChannel(channel!);
+          });
```
ogres: Can you check other approaches to avoid side-effect in the build method?

ogres:
@gauravsaini Thanks for the contribution. Could you please update your branch?
Regarding the change itself, let's avoid side effects in the build method, and see if we can instead setup necessary data before the page is open.

gauravsaini:
@ogres Thanks for the review! Updated the branch on main and moved channel setup to `onGenerateRoute` so it is configured before the page opens, avoiding side effects in `SnapPage.build`.

packages/app_center/lib/store/store_app.dart
```diff
@@ -188,15 +188,32 @@ class _StoreAppHome extends ConsumerWidget {
...
+                  .then((snapData) {
+                    if (snapData.availableChannels?.containsKey(channel) ??
+                        false) {
+                      ref
+                          .read(snapModelProvider(snapName).notifier)
+                          .selectChannel(channel);
+                    }
+                  })
+                  .catchError((_) {});
+            }
```
ogres: Can we extract this into a separate function?
This way, we keep the build method small and readable.

Plus, it is still a side-effect, but just in a different build method.
We need to find a different way to pass this before the `build` method of snap-model, so that it can set up the necessary field from the beginning.

gauravsaini: @ogres Addressed! Extracted snap route generation into a separate helper method (`_createSnapRoute`) to keep the build method clean and concise.

To eliminate the build-time side-effect, `SnapModel.build` now watches `snapInitialChannelProvider` and passes it to `defaultSelectedChannel(..., preferredChannel: ...)`, setting up the requested channel right from the start without async mutation callbacks.

PR #2054, opened 2026-03-23
fix: case-insensitive filtering for category search results

gauravsaini:
## Summary

Fixes case-insensitive filtering for featured snaps in category search results.

## Problem

When filtering featured snaps by name within categories, the search was case-sensitive. Searching for 'py' would not match 'Python'.

## Changes

- Convert search query to lowercase once before filtering
- Compare lowercase snap name with lowercase query

### DIFF
--- a/packages/app_center/lib/snapd/snap_search.dart
+++ b/packages/app_center/lib/snapd/snap_search.dart
@@ -39,10 +39,11 @@ final snapSearchProvider =
       searchParameters,
     ) async* {
       final snapd = getService<SnapdService>();
+      final queryLower = (searchParameters.query ?? '').toLowerCase();
       if (searchParameters.category == SnapCategoryEnum.ubuntuDesktop) {
         yield* snapd.getStoreSnaps(
           searchParameters.category!.featuredSnapNames
-                  ?.where((name) => name.contains(searchParameters.query ?? ''))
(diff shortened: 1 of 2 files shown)


ogres:
@gauravsaini Thnks for the contribution. We are thinking about redesigning some of the parts of the page, and the sorting options as well. Currently, we have multiple options and it is already slightly confusing, adding more to it would be even confusing. We won't be able to accept this change at the moment.

Could you separate your changes regarding the case sensitive search? That sounds like a good addition.

gauravsaini:
Thabks for finally coming back on this.
I would surely be able to do that in a while.
Cheers

ogres:
@gauravsaini 

1. Could you [sign the commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) ?
2. Could you add relevant tests, to ensure we do not regress in future?

ogres:
@gauravsaini It seems that some of the tests are failing. Could you please update your branch and run unit tests locally?

gauravsaini:
> @gauravsaini It seems that some of the tests are failing. Could you please update your branch and run unit tests locally?

Will try to do this by tomorrow

gauravsaini:
> @gauravsaini It seems that some of the tests are failing. Could you please update your branch and run unit tests locally?

Squashed and force pushed

ogres (approved):

PR #2058, opened 2026-03-23
fix: reduce CPU usage on Manage page (#1998)

gauravsaini:
## Summary

- Fixes high CPU usage (10%+) on the Manage screen when many snaps are installed
- Fixes #1998

## Root Cause

The `ManageAppActions` widget called `ref.watch(snapModelProvider(snap.name))` for every snap tile, creating a full `SnapModel` for each installed snap. With 100+ snaps, this caused significant overhead as each `SnapModel` independently fetches data from snapd.

## Fix

Refactored `_buildSnapActions` to avoid creating per-snap `SnapModel` instances on each tile:

1. **Display decisions use `Snap` data directly** - `refreshInhibit`, `apps.isNotEmpty`, `updateVersion` are available from `ManageSnapData.snap` without needing a full `SnapModel`

2. **Active changes tracked at manage page level** - Uses `currentlyInstallingModelProvider` instead of per-snap `SnapModel.activeChangeId`

3. **Lazy SnapModel creation** - Only creates/reads `snapModelProvider` when an action button (update, open, remove) is actually pressed

4. **Simplified launchability check** - Uses `snap.apps.isNotEmpty` instead of `LaunchProvider`

## Testing

Verify that:
- Manage page loads quickly with 100+ installed snaps
- Update, Open, and Remove buttons still work correctly
- Active change status still shows when a snap operation is in progress

### DIFF
--- a/packages/app_center/lib/manage/manage_app_actions.dart
+++ b/packages/app_center/lib/manage/manage_app_actions.dart
@@ -52,10 +52,9 @@ class ManageAppActions extends ConsumerWidget {
     );
   }
 
-  /// Builds snap action buttons using the per-snap [SnapModel]. Shows a loading
-  /// indicator while the snap model loads, an active change status when a snapd
-  /// operation is in progress, or the appropriate action buttons (update, open,
-  /// remove) otherwise.
+  /// Builds snap action buttons. Uses data from [snap] directly for display
+  /// decisions to avoid creating per-snap [SnapModel] instances on each tile.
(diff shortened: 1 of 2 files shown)


wassupluke:
> when many snaps are installed

Happens even on a fresh install from my experience today. I'm just here because the stinking Manage tab does nothing but spin the circles. I'm trying to give Ubuntu a try for moving from my Debian, but doing a fresh install and then immediately not being able to update the system with the GUI...blows my mind this is a release.

PR #2061, opened 2026-03-24
fix: filter held snaps from update available list (#1983)

gauravsaini:
## Summary

- Snaps held via `snap refresh --hold` are now excluded from the updates available list
- Fixes #1983

## Problem

When a snap's updates are manually held (via `snap refresh snapName --hold`), the snap still appeared in the App Center's "Updates available" list. This is confusing because the user explicitly held the update.

## Fix

Added a filter in `fetchRefreshableSnaps()` to exclude snaps where `snap.hold` is set and in the future. The `hold` field on the `Snap` class indicates when updates are held until, so we check if `hold` is null or has passed.

## Testing

1. Run `snap refresh --hold <snap-name>`
2. Open App Center → Manage
3. The held snap should no longer appear in the updates list

### DIFF
--- a/packages/app_center/lib/manage/snap_updates_model.dart
+++ b/packages/app_center/lib/manage/snap_updates_model.dart
@@ -104,9 +104,11 @@ class SnapUpdatesModel extends _$SnapUpdatesModel {
 
   Future<SnapListState> fetchRefreshableSnaps() {
     return connectionCheck(
-      () => _snapd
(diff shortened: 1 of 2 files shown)


ogres:
We might need to rethink how we show the snaps that are on hold. Simply filtering them out might not be a good idea.
We probably need a proper way to show that they are on hold. I will need to double check this.

PR #2085, opened 2026-04-07
feat: Unify action cancel button

ashuntu:
Debs and Snaps used to use separate methods for  cancelling installing/uninstalling/etc., this unifies them to use the same widget.

---

UDENG-9616

### DIFF
--- a/packages/app_center/lib/deb/deb_page.dart
+++ b/packages/app_center/lib/deb/deb_page.dart
@@ -3,7 +3,6 @@ import 'dart:async';
 import 'package:app_center/apps/app_page.dart';
 import 'package:app_center/apps/app_title_bar.dart';
 import 'package:app_center/appstream/appstream.dart';
-import 'package:app_center/constants.dart';
 import 'package:app_center/deb/deb_model.dart';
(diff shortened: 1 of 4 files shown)


matthew-hagemann (changes requested):
Couple small function signature change requests, otherwise LGTM!

packages/app_center/lib/deb/deb_page.dart
```diff
@@ -157,37 +156,21 @@ class _DebActionButtons extends ConsumerWidget {
...
+            child: Text(primaryAction.label(l10n)),
           );
 
-    final cancelButton = OutlinedButton(
-      onPressed: DebAction.cancel.callback(ref, debModel),
-      child: Text(DebAction.cancel.label(l10n)),
-    );
+    if (debModel.activeTransactionId != null) {
+      return ActiveChangeStatus(
+        onCancelPressed: () => DebAction.cancel.callback(ref, debModel),
```
matthew-hagemann: We need to drop the whole lambda, `onCancelPressed: DebAction.cancel.callback(ref, debModel),`

matthew-hagemann (approved):
Thanks! LGTM!

PR #2089, opened 2026-04-10
l10n: wording update

matthew-hagemann:
depends on:
#2087


ogres:
@matthew-hagemann I will close this one, let's create a new key in future if we still need it. That way, other languages can also be properly updated

PR #2091, opened 2026-04-14
feat: Load local theme icon when available for debs

ashuntu:
For deb icons, search for local themed icons instead of always using the upstream icons. This is mostly relevant for "generic" apps like Nautilus and the Ubuntu terminal, so now they show with the icon the user generally expects in search results.

Also a note from working on this: maybe we should have offline fallbacks for *all* icons (where possible) in the future since the App Center is doing more and more general app management and may not always be used just to install apps from the internet. In addition, Debs in general need a caching solution similar to Snaps to speed up load times.

---

UDENG-9784


packages/app_center/lib/apps/app_title_bar.dart
```diff
@@ -38,7 +38,7 @@ class AppTitleBar extends StatelessWidget {
     Widget? banner,
   }) =>
       AppTitleBar(
-        iconUrl: debData.component.icon,
+        iconWidget: DebAppIcon(component: debData.component, size: 96),
```
matthew-hagemann: nit: its only used in two places, but maybe worth pulling 96 into a const for this file

packages/app_center/lib/apps/apps_utils.dart
```diff
@@ -70,3 +77,101 @@ extension AppConfinementL10n on AppConfinement {
...
+}) async {
+  final settings = GSettings(
+    'org.gnome.desktop.interface',
+    backend: backend,
+  );
+  try {
+    final value = await settings.get('icon-theme');
+    final themeName = value.asString().trim();
+    if (themeName.isNotEmpty) return themeName;
+  } on Exception {
```
matthew-hagemann: I'd log the error here for future debugging, just in case some change in integration or sand boxing causes an issue here in the future (I had an xdg integration issue earlier this week, took a bit to track it down 😅 )

packages/app_center/lib/apps/apps_utils.dart
```diff
@@ -70,3 +77,101 @@ extension AppConfinementL10n on AppConfinement {
...
+                p.join(dataDir, 'icons', theme, size, subdir, '$name$ext');
+            if (File(path).existsSync()) return path;
+          }
+        }
+      }
+    }
+  }
+
+  for (final ext in supportedExtensions) {
+    final path = p.join(pixmapsDir, '$name$ext');
```
matthew-hagemann: If I'm reading this correctly, we first search the theme for an icon, then fallback to `pixmaps`. My understanding from the  spec is we need to recursively look for inherited themes, going back to `hicolor` before we resort to `pixmaps`? This is just my reading though, not a part of GTK I know well.

https://specifications.freedesktop.org/icon-theme/latest/#icon_lookup

It also may be worth this living upstream in https://github.com/canonical/gtk.dart

ashuntu: Originally I was going for just looking for the active theme, and relying on upstream if the theme icon isn't found, but I think this would probably make more sense.

I'll see about an addition to gtk.dart; I think it would need to be a new class that provides an icon wrapper of some sort.

packages/app_center/lib/apps/apps_utils.dart
```diff
@@ -70,3 +77,101 @@ extension AppConfinementL10n on AppConfinement {
...
+  const sizes = ['scalable', '256x256', '128x128', '64x64', '48x48'];
+  const iconSubdirs = ['apps', 'categories'];
+
+  if (theme != null) {
+    for (final dataDir in dataDirs) {
+      for (final size in sizes) {
+        for (final subdir in iconSubdirs) {
+          for (final ext in supportedExtensions) {
+            final path =
+                p.join(dataDir, 'icons', theme, size, subdir, '$name$ext');
```
matthew-hagemann: I cleared my cache and ran this build from a snap and can't see any icons resolving:

Image

After some debugging, it doesn't look like we have access to any of them 🙁
```sh
snap run --shell snap-store -c 'ls /usr/share/pixmaps'
python3.12.xpm  python3.xpm  ubuntu-logo-text-dark.png  ubuntu-logo-text.png  ubuntu-logo.svg

snap run --shell snap-store -c 'ls $XDG_DATA_HOME/icons'
// nothing

snap run --shell snap-store -c '
 for dir in $(echo $XDG_DATA_DIRS | tr ":" " "); do
   echo "=== $dir/icons/ ==="
   ls $dir/icons/ 2>/dev/null || echo "(empty or missing)"
 done
 '
=== /home/matt/snap/snap-store/x3/.local/share/icons/ ===
=== /home/matt/snap/snap-store/x3/icons/ ===
(empty or missing)
=== /snap/snap-store/x3/data-dir/icons/ ===
Adwaita                    Radiant-MATE       Yaru-olive-dark          breeze_cursors-2
Ambiant-MATE               Yaru               Yaru-prussiangreen       default
Breeze_Light               Yaru-MATE-dark     Yaru-prussiangreen-dark  elementary
Breeze_Snow                Yaru-MATE-light    Yaru-purple              elementary-xfce
DMZ-Black                  Yaru-bark          Yaru-purple-dark         elementary-xfce-dark
DMZ-White                  Yaru-bark-dark     Yaru-red                 elementary-xfce-darker
HighContrast               Yaru-blue          Yaru-red-dark            elementary-xfce-darkest
Humanity                   Yaru-blue-dark     Yaru-sage                handhelds
Humanity-Dark              Yaru-dark          Yaru-sage-dark           hicolor
Papirus-Adapta-Maia        Yaru-magenta       Yaru-viridian            redglass
Papirus-Adapta-Nokto-Maia  Yaru-magenta-dark  Yaru-viridian-dark       ubuntu-mono-dark
Papirus-Dark-Maia          Yaru-mate          breeze                   ubuntu-mono-light
Papirus-Light-Maia         Yaru-mate-dark     breeze-dark              whiteglass
Papirus-Maia               Yaru-olive         breeze_cursors
=== /snap/snap-store/x3/gnome-platform/usr/share/icons/ ===
Adwaita   Humanity-Dark  gnome-logo-text-dark.svg  hicolor           ubuntu-mono-light
Humanity  LoginIcons     gnome-logo-text.svg       ubuntu-mono-dark
=== /home/matt/.local/share/flatpak/exports/share/icons/ ===
(empty or missing)
=== /var/lib/flatpak/exports/share/icons/ ===
(empty or missing)
=== /usr/local/share//icons/ ===
(empty or missing)
=== /usr/share//icons/ ===
hicolor  locolor
=== /var/lib/snapd/desktop/icons/ ===
(empty or missing)
=== /snap/snap-store/x3/gpu-2404/usr/share/icons/ ===
(empty or missing)
```

(themes in `/snap/snap-store/x3/data-dir/icons/` don't contain icons either):
```sh
snap run --shell snap-store -c 'ls /snap/snap-store/x3/data-dir/icons/Yaru-dark/22x22'
animations
snap run --shell snap-store -c 'ls /snap/snap-store/x3/data-dir/icons/Yaru-dark/16x16'
snap run --shell snap-store -c 'ls /snap/snap-store/x3/data-dir/icons/Yaru-dark/22x22'
animations
snap run --shell snap-store -c 'ls /snap/snap-store/x3/data-dir/icons/Yaru-dark/16x16'
snap run --shell snap-store -c 'ls /snap/snap-store/x3/data-dir/icons/Yaru-dark/22x22'
animations
snap run --shell snap-store -c 'ls /snap/snap-store/x3/data-dir/icons/Yaru-dark/24x24'
animations
```



I'm not quite sure what we should do? If we need a new plug?

ashuntu: For Snaps I think we will need a special consideration for the paths, though I'm unsure if we'll need special permissions for finding the icons through enumeration. For example, you should have read access to `/var/lib/snapd/hostfs/usr/share/icons/Yaru/index.theme` within the snap.

This could be as easy as appending the hostfs paths to XDG_DATA_DIRS, though I'll need to investigate that a bit.

matthew-hagemann: ah yea, there they are. 
❯ snap run --shell snap-store -c 'ls /var/lib/snapd/hostfs/usr/share/icons/hicolor/scalable/apps/org.gnome.Calculator.svg'
/var/lib/snapd/hostfs/usr/share/icons/hicolor/scalable/apps/org.gnome.Calculator.svg

I think then if you add the `/var/lib/snapd/hostfs/usr/share` path it should work. The enumeration shouldn't be an issue if you have the direct path, which you construct already via that for loop.

matthew-hagemann: The hardcoded path to pixmaps will need to be updated aswell then, seems to exist via `/var/lib/snapd/hostfs`
❯ snap run --shell snap-store -c 'ls /var/lib/snapd/hostfs/usr/share/pixmaps/debian-logo.png'
/var/lib/snapd/hostfs/usr/share/pixmaps/debian-logo.png

ashuntu:
Marking this as draft pending a (maybe) addition to gtk.dart.

ashuntu:
Added themed icon support to Gtk.dart, and updated this PR to utilize it: https://github.com/canonical/gtk.dart/pull/57.

The gtk.dart implementation doesn't attempt to re-implement the logic like this was before, but instead calls gtk directly, so this should work much better. From my testing this seems to work in a snap as well, but @matthew-hagemann let me know if it doesn't for you.

matthew-hagemann (approved):
Thanks! LGTM and works well in testing.

PR #2098, opened 2026-04-21
fix: document path resolution

matthew-hagemann:
Adds document path resolution for FUSE paths using the `GetHostPaths` API on the `org.freedesktop.portal.Documents` session bus endpoint. Falls back to the provided path if that call fails.

I've manually tested opening a local deb and installing it via the app center on both 24.04.4 and 26.04

### DIFF
--- a/packages/app_center/lib/packagekit/packagekit_service.dart
+++ b/packages/app_center/lib/packagekit/packagekit_service.dart
@@ -43,7 +43,7 @@ class PackageKitService {
   final XdgDocumentsPortal? _documentsPortal;
   final FileSystem _fs;
   final String? _runtimeDir;
-  XdgDesktopPortalClient? _portalClient;
+  XdgDesktopPortalClient? _desktopPortalClient;
   Future<io.Directory>? _mountPointFuture;
 
   bool get isAvailable => _isAvailable;
@@ -226,33 +226,34 @@ class PackageKitService {
       return snapArch;
     }
 
-    final result = await io.Process.run('/usr/bin/dpkg', ['--print-architecture']);
+    final result =
+        await io.Process.run('/usr/bin/dpkg', ['--print-architecture']);
     return (result.stdout as String).trim();
   }
 
   String _getAbsolutePath(String path) => _fs.file(path).absolute.path;
 
-  Future<XdgDocumentsPortal> _getPortal() async {
-    return _documentsPortal ?? (_portalClient ??= XdgDesktopPortalClient()).documents;
-  }
+  XdgDocumentsPortal get _portal =>
+      _documentsPortal ??
+      (_desktopPortalClient ??= XdgDesktopPortalClient()).documents;
 
   Future<io.Directory> _getMountPoint() {
-    return _mountPointFuture ??= _getPortal().then((p) async => p.getMountPoint());
+    return _mountPointFuture ??= _portal.getMountPoint();
   }
 
   Future<bool> _isPortalPath(String path) async {
     try {
       final mountPoint = await _getMountPoint();
       return isWithin(mountPoint.path, path);
-    } on Exception {
+    } on Exception catch (e) {
+      log.warning('Failed to check if $path is a portal path: $e');
       return false;
     }
   }
 
   Future<String?> _resolvePortalPath(String path) async {
-    final portal = await _getPortal();
+    final portal = _portal;
     final mountPoint = await _getMountPoint();
-    // Path is /<mountPoint>/<docId>/filename — extract docId as first segment after mount
     final relativePath = relative(path, from: mountPoint.path);
     final docId = split(relativePath).first;
     try {
@@ -263,13 +264,12 @@ class PackageKitService {
         'Documents portal returned no path for $docId. Falling back to original path.',
       );
       return null;
-    } on DBusUnknownMethodException {
-      rethrow;
     } on Exception catch (e) {
       log.warning(
-        'Failed to resolve portal path $path via Documents portal: $e. Falling back to original path.',
+        'Failed to resolve portal path $path via Documents portal '
+        '(${e.runtimeType}): $e — triggering copy fallback.',
       );
-      return null;
+      rethrow;
     }
   }
 
@@ -289,18 +289,14 @@ class PackageKitService {
         resolvedPath: resolved ?? _getAbsolutePath(path),
         tempCopy: null,
       );
-    } on DBusUnknownMethodException {
-      log.info(
-        'GetHostPaths not supported, copying $path to runtime directory',
-      );
+    } on Exception catch (_) {
       final copyPath = await _copyToRuntime(path);
       return (resolvedPath: copyPath, tempCopy: copyPath);
     }
   }
 
   Future<String> _copyToRuntime(String path) async {
-    final dir =
-        _runtimeDir ??
+    final dir = _runtimeDir ??
         io.Platform.environment['SNAP_USER_COMMON'] ??
         io.Directory.systemTemp.path;
     final dest = join(dir, basename(path));
@@ -445,6 +441,6 @@ class PackageKitService {
     await _dbus.close();
     await _client.close();
     await _errorStreamController.close();
-    await _portalClient?.close();
+    await _desktopPortalClient?.close();
   }
 }
--- a/packages/app_center/lib/packagekit/packagekit_service.dart
+++ b/packages/app_center/lib/packagekit/packagekit_service.dart
@@ -7,7 +7,7 @@ import 'package:file/file.dart';
 import 'package:file/local.dart';
 import 'package:flutter/material.dart';
 import 'package:packagekit/packagekit.dart';
-import 'package:path/path.dart';
+import 'package:path/path.dart' as p;
 import 'package:ubuntu_service/ubuntu_service.dart';
 import 'package:xdg_desktop_portal/xdg_desktop_portal.dart';
 
@@ -128,7 +128,7 @@ class PackageKitService {
     try {
       await action?.call(transaction);
     } on Exception {
-      subscription.cancel();
+      await subscription.cancel();
       _transactions.remove(id);
       try {
         onDone?.call();
@@ -267,7 +267,7 @@ class PackageKitService {
   Future<bool> _isPortalPath(String path) async {
     try {
       final mountPoint = await _getMountPoint();
-      return isWithin(mountPoint.path, path);
+      return p.isWithin(mountPoint.path, path);
     } on Exception catch (e) {
       log.warning('Failed to check if $path is a portal path: $e');
       return false;
@@ -277,7 +277,7 @@ class PackageKitService {
   Future<String?> _resolvePortalPath(String path) async {
     final portal = _portal;
     await _getMountPoint();
-    final docId = basename(dirname(path));
+    final docId = p.basename(p.dirname(path));
     try {
       final hostPaths = await portal.getHostPaths([docId]);
       final file = hostPaths[docId];
@@ -324,14 +324,14 @@ class PackageKitService {
           io.Directory.systemTemp.path,
     );
     final tempDir = await baseDir.createTemp('packagekit-');
-    final dest = join(tempDir.path, basename(path));
+    final dest = p.join(tempDir.path, p.basename(path));
     await _fs.file(path).copy(dest);
     return dest;
   }
 
   Future<void> _deleteTempCopy(String path) async {
     try {
-      await _fs.directory(dirname(path)).delete(recursive: true);
+      await _fs.directory(p.dirname(path)).delete(recursive: true);
     } on Exception catch (e) {
       log.warning('Failed to delete temporary file $path: $e');
     }
--- a/packages/app_center/lib/packagekit/packagekit_service.dart
+++ b/packages/app_center/lib/packagekit/packagekit_service.dart
@@ -1,5 +1,5 @@
 import 'dart:async';
-import 'dart:io';
+import 'dart:io' as io;
 
 import 'package:app_center/packagekit/logger.dart';
 import 'package:dbus/dbus.dart';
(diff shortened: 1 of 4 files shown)


packages/app_center/lib/packagekit/packagekit_service.dart
```diff
@@ -215,6 +222,46 @@ class PackageKitService {
...
   String _getAbsolutePath(String path) => _fs.file(path).absolute.path;
 
+  static bool _isPortalPath(String path) {
+    final parts = path.split('/');
+    // /run/user/<uid>/doc/<docId>/filename
+    return parts.length > 5 &&
+        parts[1] == 'run' &&
+        parts[2] == 'user' &&
+        parts[4] == 'doc';
+  }
```
d-loose: Instead of relying on string manipulations, we should use [`path`](https://pub.dev/packages/path) here

packages/app_center/lib/packagekit/packagekit_service.dart
```diff
@@ -215,6 +222,46 @@ class PackageKitService {
...
+
+  Future<String> _resolvePortalPath(String path) async {
+    final docId = path.split('/')[5];
+    final client = _sessionDbus ?? DBusClient.session();
+    try {
+      final object = DBusRemoteObject(
+        client,
+        name: 'org.freedesktop.portal.Documents',
+        path: DBusObjectPath('/org/freedesktop/portal/documents'),
+      );
```
d-loose: We have an existing dart client library for the [XDG desktop portal](https://github.com/canonical/xdg_desktop_portal.dart/), so ideally we should implement the missing method there and import the library here

packages/app_center/test/packagekit_service_test.dart
```diff
@@ -468,6 +470,96 @@ void main() {
     expect(packages.length, equals(2));
   });
 
+  group('portal path resolution', () {
+    const portalPath =
+        '/run/user/1000/doc/8cf4b075/balena-etcher_2.1.0_amd64.deb';
+    const realPath = '/home/user/Downloads/balena-etcher_2.1.0_amd64.deb';
```
d-loose: We shouldn't hardcode names of random third party packages in our tests

packages/app_center/lib/packagekit/packagekit_service.dart
```diff
@@ -215,6 +222,46 @@ class PackageKitService {
...
+      // Result is a{say}: doc_id -> null-terminated byte array path
+      final pathsDict = result.values[0] as DBusDict;
+      final pathBytes =
+          (pathsDict.children[DBusString(docId)] as DBusArray).children;
+      return String.fromCharCodes(
+        pathBytes.map((v) => (v as DBusByte).value).where((b) => b != 0),
+      );
+    } on Exception catch (e) {
+      log.warning(
+        'Failed to resolve portal path $path via Documents portal: $e. Falling back to original path.',
```
3v1n0: As mentioned in mattermost, in this case (or acutally on UnknosmMehfod error), could you please do the workaround of:
 - copy / move the file to the snap store runtime dir
 - pass that to pkgkit

So in case we backport the glib fix to old versions we don't break this

packages/app_center/lib/packagekit/packagekit_service.dart
```diff
@@ -215,6 +222,46 @@ class PackageKitService {
...
+    final parts = path.split('/');
+    // /run/user/<uid>/doc/<docId>/filename
+    return parts.length > 5 &&
+        parts[1] == 'run' &&
+        parts[2] == 'user' &&
+        parts[4] == 'doc';
+  }
+
+  Future<String> _resolvePortalPath(String path) async {
+    final docId = path.split('/')[5];
```
d-loose: This will throw an exception in case `path` doesn't have the proper format. Maybe we could integrate the validation logic here and return `null` if it doesn't parse. It could then be called as
```dart
final resolvedPath = _resolvePortalPath(path) ?? _getAbsolutePath(path);
```

Edit:
or rather
```dart
final resolvedPath = _resolvePortalPath(path) ?? p.absolute(path);
```
once we import the `path` package.

d-loose (changes requested):

packages/app_center/lib/packagekit/packagekit_service.dart
```diff
@@ -215,6 +222,46 @@ class PackageKitService {
 
   String _getAbsolutePath(String path) => _fs.file(path).absolute.path;
 
+  static bool _isPortalPath(String path) {
+    final parts = path.split('/');
+    // /run/user/<uid>/doc/<docId>/filename
```
3v1n0: This should really not be hard coded. There's an API for it:

 - https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Documents.html#org-freedesktop-portal-documents-getmountpoint

3v1n0 (changes requested):

packages/app_center/lib/packagekit/packagekit_service.dart
3v1n0: ~~Or maybe we can just use https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Documents.html#org-freedesktop-portal-documents-info that should be supported forever? ~~

It is available, but sadly not exposed to sandboxes, even though it should be safe to use for the requester, but anyways...

matthew-hagemann: Ah  I was about to say It seems blocked in sandboxes:
image

I'll try your first suggestion now :slightly_smiling_face:

packages/app_center/lib/packagekit/packagekit_service.dart
matthew-hagemann: I believe if we do the second option, we bypass the mock fs, so the first option allows us to still inject the mock for tests

packages/app_center/lib/packagekit/packagekit_service.dart
```diff
@@ -204,17 +221,96 @@ class PackageKitService {
...
+  Future<({String resolvedPath, String? tempCopy})> _resolveLocalPath(
+    String path,
+  ) async {
+    if (!await _isPortalPath(path)) {
+      return (resolvedPath: _getAbsolutePath(path), tempCopy: null);
+    }
+    try {
+      final resolved = await _resolvePortalPath(path);
+      return (
+        resolvedPath: resolved ?? _getAbsolutePath(path),
```
matthew-hagemann, line 307: This fallback is questionable, not sure if we have a better idea? But this is we found it was a portal path, but then couldn't find it when looking it up so fell back to the abs path. Not sure if this is possible so maybe overthinking it?

packages/app_center/lib/packagekit/packagekit_service.dart
```diff
@@ -265,6 +273,48 @@ class PackageKitService {
...
+      );
+      final copyPath = await _copyToRuntime(path);
+      return (resolvedPath: copyPath, tempCopy: copyPath);
+    }
+  }
+
+  Future<String> _copyToRuntime(String path) async {
+    final dir =
+        _runtimeDir ??
+        io.Platform.environment['SNAP_USER_COMMON'] ??
```
3v1n0: I would use `$XDG_RUNTIME_DIR` so it also get cleared up automatically, also ideally put this into a temporary subfolder with an unique name so that we do not end up potentially overwriting files.

packages/app_center/lib/packagekit/packagekit_service.dart
```diff
@@ -214,29 +215,40 @@ class PackageKitService {
...
   }
 
   Future<String> _resolvePortalPath(String path) async {
-    final docId = split(path)[5];
-    final portal = _documentsPortal ?? (_portalClient ??= XdgDesktopPortalClient()).documents;
+    final portal = await _getPortal();
+    final mountPoint = await _getMountPoint();
+    // Path is /<mountPoint>/<docId>/filename — extract docId as first segment after mount
+    final relativePath = relative(path, from: mountPoint.path);
+    final docId = split(relativePath).first;
```
3v1n0: Is there an equivalent of `$(basename $(dirname "$path"))` here? It should be enough rather than relying on splitting.

packages/app_center/lib/packagekit/packagekit_service.dart
```diff
@@ -204,17 +221,96 @@ class PackageKitService {
...
+    final dest = join(dir, basename(path));
+    await _fs.file(path).copy(dest);
+    return dest;
+  }
+
+  void _deleteTempCopy(String path) {
+    _fs.file(path).delete().catchError((Object e) {
+      log.warning('Failed to delete temporary file $path: $e');
+      return _fs.file(path);
+    });
```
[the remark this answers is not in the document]
matthew-hagemann: No unawaited

packages/app_center/lib/packagekit/packagekit_service.dart
```diff
@@ -204,17 +226,102 @@ class PackageKitService {
...
+
+  Future<String?> _resolvePortalPath(String path) async {
+    final portal = _portal;
+    await _getMountPoint();
+    final docId = basename(dirname(path));
+    try {
+      final hostPaths = await portal.getHostPaths([docId]);
+      final file = hostPaths[docId];
+      if (file != null) return file.path;
+      log.warning(
```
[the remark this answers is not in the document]
matthew-hagemann: You have now taken me in a circle

matthew-hagemann:
@3v1n0 sorry you got pinged there, I was iterating against the CI, but ready for another look when you can :slightly_smiling_face:

packages/app_center/lib/packagekit/packagekit_service.dart
```diff
@@ -1,13 +1,15 @@
...
-import 'dart:io';
+import 'dart:io' as io;
 
 import 'package:app_center/packagekit/logger.dart';
 import 'package:dbus/dbus.dart';
 import 'package:file/file.dart';
 import 'package:file/local.dart';
 import 'package:flutter/material.dart';
 import 'package:packagekit/packagekit.dart';
+import 'package:path/path.dart';
```
d-loose:
```suggestion
import 'package:path/path.dart as p';
```
let's follow the convention from the [docs](https://pub.dev/packages/path#using)

packages/app_center/lib/packagekit/packagekit_service.dart
```diff
@@ -101,14 +113,30 @@ class PackageKitService {
...
+    } on Exception {
+      subscription.cancel();
+      _transactions.remove(id);
+      try {
+        onDone?.call();
+      } on Exception catch (e) {
+        log.warning('onDone callback threw during transaction cleanup: $e');
+      }
+      rethrow;
+    }
```
d-loose, line 139: These changes seem unrelated to the document path resolution. Can we do such fixes in a separate PR?

matthew-hagemann: `onDone` was introduced in this PR to tidy up the temp files if we hit the cp fallback

packages/app_center/lib/packagekit/packagekit_service.dart
```diff
@@ -27,14 +29,22 @@ class PackageKitService {
...
+        _fs = fs ?? const LocalFileSystem(),
+        _runtimeDir = runtimeDir;
 
   final PackageKitClient _client;
   final DBusClient _dbus;
+  final XdgDocumentsPortal? _documentsPortal;
   final FileSystem _fs;
+  final String? _runtimeDir;
+  XdgDesktopPortalClient? _desktopPortalClient;
+  Future<io.Directory>? _mountPointFuture;
```
d-loose: Why is this a `Future`? Can't we just store the directory?

packages/app_center/lib/packagekit/packagekit_service.dart
```diff
@@ -204,17 +241,102 @@ class PackageKitService {
...
+      final mountPoint = await _getMountPoint();
+      return p.isWithin(mountPoint.path, path);
+    } on Exception catch (e) {
+      log.warning('Failed to check if $path is a portal path: $e');
+      return false;
+    }
+  }
+
+  Future<String?> _resolvePortalPath(String path) async {
+    final portal = _portal;
```
d-loose: Is a local variable necessary here?

packages/app_center/test/packagekit_service_test.dart
```diff
@@ -514,3 +646,38 @@ MockDBusClient createMockDbusClient() {
...
+}) {
+  final portal = MockXdgDocumentsPortal();
+  when(portal.getMountPoint()).thenAnswer(
+    (_) async => io.Directory(mountPoint),
+  );
+  when(portal.getHostPaths([docId])).thenThrow(
+    DBusUnknownMethodException(DBusMethodErrorResponse.unknownMethod()),
+  );
+  return portal;
+}
```
d-loose, line 676: Following the style of the existing tests, it'd be more consistent to have a singe `creatMockDocumentsPortal` method that creates a mock with a specific behavior based on its arguments, instead of having many dedicated methods

packages/app_center/lib/packagekit/packagekit_service.dart
d-loose: ah, my bad, should've read the entire diff :sweat_smile:

packages/app_center/lib/packagekit/packagekit_service.dart
matthew-hagemann: Oh sorry, no not anymore

packages/app_center/lib/packagekit/packagekit_service.dart
matthew-hagemann: Remanent of an over complication, removing!

d-loose (approved):
Thanks, LGTM now :+1:

3v1n0:
Nice, thanks

PR #2100, opened 2026-04-22
fix: drop mime from packagekit session

matthew-hagemann:


aleasto:
LGTM; this line should never have been added, sorry :grimacing:

ashuntu (approved):

PR #2107, opened 2026-04-29
Show download progress percentage in active change status

stsfaroz:
Currently the active change status shows only the action label (e.g. "Installing", "Uninstalling") with no progress percentage.

This change displays the percentage below the label so users can see how far along the operation is.

Works for all action labels since `actionLabel` is passed dynamically - nothing is hardcoded.

Percentage is shown below the label using a `Column` rather than inline, because the parent Row has a constrained width. Appending the percentage inline (e.g. "Uninstalling 45%") causes a `RenderFlex` right overflow. Stacking them vertically with a `Column` fixes the overflow while keeping both pieces of information visible.

## Screenshots

**Before:**

old

old_uninstall

**After:**

new

new_uninstall

### DIFF
--- a/packages/app_center/lib/widgets/active_change_content.dart
+++ b/packages/app_center/lib/widgets/active_change_content.dart
@@ -43,6 +43,7 @@ class _ActiveChangeText extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final percentage = (progress * 100).toStringAsFixed(0);
     return Row(
       children: [
         SizedBox.square(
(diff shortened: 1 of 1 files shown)


juanruitina:
I tested this in a few different screen sizes and languages and it works well. I think the extra label will also be good for screen readers. UX+1!

ashuntu:
@stsfaroz You'll need to complete the CLA and sign your commits for this to move forward: https://github.com/ubuntu/app-center/blob/main/CONTRIBUTING.md#contributor-license-agreement

stsfaroz:
@ashuntu I've already signed the CLA on April 29. Could you re-run the CLA check?

ashuntu (approved):

PR #2108, opened 2026-05-04
feat: Move switch channel to a dialog

ashuntu:
Moves the switch channel drop-down and button to a separate dialog found under the more actions button.

| Screenshots |
|-------------|
| image |
| image |
| image |

---

UDENG-9910

### DIFF
--- a/packages/app_center/lib/snapd/snap_page.dart
+++ b/packages/app_center/lib/snapd/snap_page.dart
@@ -292,12 +292,15 @@ class _SwitchChannelButton extends ConsumerWidget {
       children: [
         YaruSplitButton.outlined(
           onPressed: hasChangedChannel && snapData.activeChangeId == null
-              ? SnapAction.switchChannel.callback(
-                  snapData,
-                  snapViewModel,
-                  snapLauncher,
-                  context,
-                )
+              ? () {
+                  SnapAction.switchChannel.callback(
+                    snapData,
+                    snapViewModel,
+                    snapLauncher,
+                    context,
+                  )?.call();
+                  Navigator.of(context).pop();
+                }
               : null,
           child: Stack(
             alignment: AlignmentGeometry.center,
--- a/packages/app_center/lib/snapd/snap_page.dart
+++ b/packages/app_center/lib/snapd/snap_page.dart
@@ -124,6 +124,7 @@ class _ActionBar extends ConsumerWidget {
           ),
         if (snapData.isInstalled)
           ...[
+            _UninstallButton(snapData: snapData),
             ratingsModel.whenOrNull(
               data: (ratingsData) => _RatingsActionButtons(
                 ratingsData: ratingsData,
@@ -205,6 +206,25 @@ class _PrimaryActionButton extends ConsumerWidget {
   }
 }
 
+class _UninstallButton extends ConsumerWidget {
+  const _UninstallButton({required this.snapData});
+
+  final SnapData snapData;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final l10n = AppLocalizations.of(context);
+    final snapViewModel = ref.watch(snapModelProvider(snapData.name).notifier);
+
+    return OutlinedButton(
+      onPressed: snapData.activeChangeId == null
+          ? SnapAction.remove.callback(snapData, snapViewModel, null, context)
+          : null,
+      child: Text(SnapAction.remove.label(l10n)),
+    );
+  }
+}
+
 class _MoreActionsButton extends ConsumerWidget {
   const _MoreActionsButton({required this.snapData});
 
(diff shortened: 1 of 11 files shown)


juanruitina:
So glad to see this happening. Some missing stuff:

- Can we signal in the dropdown what is the default channel? So the user can revert to it if needed. The default channel should always be the first in the dropdown.
- The current channel is not visible in the app page. I suggest to prepend the channel name to the version under the version field _only_ when not the default. I suggest that we take this same approach for the list tiles in the manage page: we only show the channel if not the default. @anasereijo what do you think?
- I see that the version field changes even before clicking on "Switch channels". It shouldn't until user confirms.
- Given we are getting back some space, can we add the uninstall button next to the open one? It should always be outlined (not green). This is particularly relevant visually for apps that can't be opened so it doesn't look broken.

packages/app_center/lib/snapd/snap_page.dart
matthew-hagemann: I'm wondering if when the user hits the switch channel button if we should not now also close the dialog box for them?

packages/app_center/lib/snapd/snap_page.dart
```diff
@@ -124,14 +126,15 @@ class _ActionBar extends ConsumerWidget {
             snapName: snapData.name,
             isPrimary: true,
           ),
-        if (snapData.availableChannels != null &&
```
matthew-hagemann, line 127: I think we want to keep this check? Not 100% what the UI should look like, but there are a number of snaps where they only have 1 channel and this becomes an odd flow?

Image

packages/app_center/lib/snapd/snap_page.dart
```diff
@@ -235,12 +238,19 @@ class _MoreActionsButton extends ConsumerWidget {
...
                     : null;
                 return PopupMenuItem(
-                  onTap: action.callback(
-                    snapData,
-                    snapViewModel,
-                    snapLauncher,
-                    context,
-                  ),
+                  onTap: action == SnapAction.switchChannel
+                      ? () => showDialog(
```
matthew-hagemann: I don't think we should special case the show  dialog action. You can move the logic for showing the dialog to `snap_action.dart`, which already has similar handing for the UI for `SnapAction.revert`

packages/app_center/lib/snapd/snap_page.dart
juanruitina: We should!

packages/app_center/lib/snapd/snap_page.dart
juanruitina: I'd think there should be no "Switch channel" functionality whatsoever if there's a single channel.

ashuntu: This check was essentially moved to `List<SnapAction> secondaryActions([SnapLauncher? snapLauncher])`. It still behaves like stable App Center at the moment.

~~@juanruitina if we hide the switch channel from the more actions button when only 1 channel is available, where will the user see which channel the app is? I suppose in the case where there is only 1 then they probably don't care.~~ This is covered by adding the channel to the version as previously suggested :)

ashuntu:
>     * Can we signal in the dropdown what is the default channel? So the user can revert to it if needed. The default channel should always be the first in the dropdown.

@juanruitina I would think either adding "(default)" to the end of the channel name in the selector or adding a row under the "last updated" row with "Default". Which of these is preferred, if any?

@matthew-hagemann do you know if there is an easy way to get the default channel of a snap like this? I suppose the store snap holds a reference to the "default" version of the snap with `channel` and `revision`, though `channel` just outputs "stable" and not the full channel name. Further, it would be nice to sort the channel list in the same way its sorted in `snap info <snap>`, but I don't think we can reliably do this unless I'm mistaken.

matthew-hagemann:
> @matthew-hagemann do you know if there is an easy way to get the default channel of a snap like this? I suppose the store snap holds a reference to the "default" version of the snap with `channel` and `revision`, though `channel` just outputs "stable" and not the full channel name. Further, it would be nice to sort the channel list in the same way its sorted in `snap info <snap>`, but I don't think we can reliably do this unless I'm mistaken.

@ashuntu yea, I've had a look at the API's I can think of, both the api spec and the snapd impl, and it doesn't look like we have what we need to identify the default if its been updated by the developer away from `latest/stable`, so I agree we don't have a way to do this reliably.

ashuntu:
> @ashuntu yea, I've had a look at the API's I can think of, both the api spec and the snapd impl, and it doesn't look like we have what we need to identify the default if its been updated by the developer away from `latest/stable`, so I agree we don't have a way to do this reliably.

I think we forgo the default channel sorting for the moment then.

> Given we are getting back some space, can we add the uninstall button next to the open one? It should always be outlined (not green). This is particularly relevant visually for apps that can't be opened so it doesn't look broken.

This will be done in another PR.

matthew-hagemann (approved):

PR #2115, opened 2026-05-15
a11y: Featured snaps icons should have identifying labels

CarlosNihelton:
Screen readers only say "Image" when we tab over the icons in the featured snaps banner. Users that depend on screen readers cannot really know which apps are those.

With the snap name or title as labels those images become immediately more understandable.

### DIFF
--- a/packages/app_center/lib/widgets/banner.dart
+++ b/packages/app_center/lib/widgets/banner.dart
@@ -220,9 +220,13 @@ class _BannerIconState extends State<_BannerIcon> {
                     ),
                   ],
                 ),
-                child: AppIcon(
-                  iconUrl: widget.snap.iconUrl,
-                  size: widget.iconSize.height * scale,
+                child: Semantics(
+                  label: widget.snap.titleOrName,
+                  button: true,
+                  child: AppIcon(
+                    iconUrl: widget.snap.iconUrl,
+                    size: widget.iconSize.height * scale,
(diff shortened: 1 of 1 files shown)


packages/app_center/lib/widgets/banner.dart
```diff
@@ -220,9 +220,12 @@ class _BannerIconState extends State<_BannerIcon> {
...
                 ),
-                child: AppIcon(
-                  iconUrl: widget.snap.iconUrl,
-                  size: widget.iconSize.height * scale,
+                child: Semantics(
+                  label: widget.snap.titleOrName,
+                  child: AppIcon(
+                    iconUrl: widget.snap.iconUrl,
+                    size: widget.iconSize.height * scale,
+                  ),
```
ashuntu, line 229:
```suggestion
                child: Semantics(
                  label: widget.snap.titleOrName,
                  button: true,
                  child: AppIcon(
                    iconUrl: widget.snap.iconUrl,
                    size: widget.iconSize.height * scale,
                  ),
```

ashuntu (changes requested):

packages/app_center/lib/widgets/banner.dart
CarlosNihelton: Good point!

ashuntu (approved):

PR #2117, opened 2026-05-21
docs: AGENTS.md bootstrap

matthew-hagemann:
CI failures seem to be upstream issues with downloading snaps at time of opening this PR, I'll re-run them a bit later. 

Bootstraps the initial `AGENTS.md` layers, along with symlinks for Claude code to use. They'll get iterated on as I solve the next few tickets to refine them. 

I'm broadly trying to follow the recommendations for writing these from Anthropics best practices here:
https://code.claude.com/docs/en/best-practices#write-an-effective-claude-md

And the sections on `CLAUDE.md`'s from this guide on larger codebases:
https://claude.com/blog/how-claude-code-works-in-large-codebases-best-practices-and-where-to-start


matthew-hagemann:
I can't add you as reviewers? But cc @gabelluardo, @juanruitina and @SoumyaRanjanPatnaik for feedback aswell :slightly_smiling_face:

gabelluardo (commented):
LGTM 👍

packages/app_center/test/AGENTS.md
```diff
@@ -0,0 +1,46 @@
...
+- `pumpApp(builder)` — pumps a `MaterialApp` with l10n, theming, and Ubuntu font
+- `createContainer(overrides)` — creates a `ProviderContainer` with auto-dispose and `MemoryFileSystem`
+
+## Data Builders
+
+Use factory functions from @test_utils.dart to create test data with sensible defaults:
+- `createSnap(...)` — builds a `Snap` with optional overrides
+- `createAppstreamComponent(...)` — builds an `AppstreamComponent`
+- `createLocalDebInfo(...)` — builds a `LocalDebInfo`
+- Pre-built constants: `defaultInstalledDeb`, `defaultDebWithUpdate`
```
d-loose: Some of these are quite specific, do we have a strategy to keep those up to date? Could we add an instruction here to update this file whenever changes to those things are made?

matthew-hagemann: Its an interesting one, becasue while the *SOMETHING*.md files are all injected in almost the same way by all the harnesses, the way to keep them inline different per harness. I have only really got a good answer for Claude Code with [this](https://github.com/anthropics/claude-plugins-official/blob/main/plugins/claude-md-management/skills/claude-md-improver/SKILL.md) plugin, which you can set up to be invoked via a [hook](https://code.claude.com/docs/en/hooks), but that's all harness specific. I want to do a bit more research on how to share skills across harnesses, which is all to say, good question, I don't know :joy: 

100% thought this is way to specific, especially if we don't have a reliable way to keep it sync'd

d-loose (approved):

Issue #1679, opened 2024-05-28
Local deb installs: design needs


d-loose:
In order to support installing third party deb packages, we'll need to address a few UI/UX issues:

* We'll want to launch the app-center with a path to a local deb package. The user should then land on an app page (similar to the existing deb page) that shows the package metadata. Generally, there'll be less data available since we don't have any external metadata from appstream. Packagekit provides the following fields: `name/id`, `group` (similar to snap categories, but we could choose to ignore that to avoid confusion), `summary` (one-line), `description` (multiple lines), `URL` (e.g. upstream homepage), `license`, `package size`.
* We should be very clear about the security concerns that come with installing third party packages (see e.g. [here](https://github.com/ubuntu/app-center/issues/1407#issuecomment-2079635110)). It should be obvious at first glance that the package isn't part of any of the archives enabled in "Software and Updates".
* As a final step, we should display a confirmation dialog that is shown when the user presses the "Install" button.

@anasereijo we can have a chat about this early next week.

PR #2118, opened 2026-05-22
fix(deb): surface local .deb install errors instead of an endless spinner

SzymonPaczos:
Installing a local `.deb` that fails, for example one built for a different CPU architecture, left the page spinning forever (earlier reports in this area: #1485, #1679).

This PR does two things:

1. Checks the package architecture against the system's native one (via PackageKit). On a certain mismatch, like an arm64-only package on amd64, the package page still loads with all details, but the install button is disabled and an "Incompatible package" info box explains why. Unclear cases (multiarch, `all`, unknown architectures) are left to PackageKit.
2. Subscribes `LocalDebModel` to `PackageKitService.errorStream`, like `DebModel` already does, so any failed install shows an error dialog and clears the spinner.

Includes unit tests for the architecture check, the model, and the page states.


ogres:
@SzymonPaczos Could you attach the screenshot please?
In general, blocking the whole page/model might not be the best approach.
Instead, we could disable the installation with relevant info. This way, users will be able to see full details of the package.

SzymonPaczos:
> @SzymonPaczos Could you attach the screenshot please?
> 
> In general, blocking the whole page/model might not be the best approach.
> 
> Instead, we could disable the installation with relevant info. This way, users will be able to see full details of the package.

Yes, a please give me a day or two.

ogres:
> > @SzymonPaczos Could you attach the screenshot please?
> > In general, blocking the whole page/model might not be the best approach.
> > Instead, we could disable the installation with relevant info. This way, users will be able to see full details of the package.
> 
> Yes, a please give me a day or two.

No rush, feel free to reach out if you need an assistance, thanks again for the contribution!

SzymonPaczos:
Sorry it took a while. Screenshots attached.

I agree that blocking the whole page wasn't the best approach, so I reworked the branch (and rebased it on current main). The page now always loads with full package details. When the architecture doesn't match, the install button is disabled and an info box explains why.

Before (old version of this PR):

before

After:

after

The same package with a matching architecture, for comparison:

compatible

Changes:
- `LocalDebModel` no longer throws on a mismatch. `LocalDebData` now has a `systemArch` field and an `isArchitectureCompatible` getter, and the button and banner react to it.
- The check only blocks installs that are certain to fail (like arm64 on amd64). Unclear cases are left to PackageKit.
- The model also listens to `PackageKitService.errorStream`, like `DebModel` already does, so a failed install shows an error instead of spinning forever. That was the actual cause of the original reports.

PR #2127, opened 2026-06-17
chore: update flutter and dependencies

d-loose:
Updates to flutter 3.44.2


d-loose:
Hm wait, I think we finally need to add `trailing-commas: preserve` to ubuntu_lints

ashuntu (dismissed):
Yaru 10.2.0 was released last week, not sure if you want that to be part of this (explicitly specified in the pubspec, I mean).

App Center tested locally, seems to run fine, focus borders work, screen reader works.

ashuntu (approved):

PR #2129, opened 2026-06-23
fix(a11y): Announce snackbar contents

ashuntu:
Adds a `sendAnnouncement` on showing `SnackBar` widgets for screen reader users.

This might be something that Flutter should just do by default :thinking:.

---

UDENG-10299


matthew-hagemann (approved):

matthew-hagemann:
Does seem like the sort of thing that should be default :thinking: could be well worth raising upstream

PR #2135, opened 2026-07-10
feat: Move About page as a nested page under new Settings page

ashuntu:
This is to make way for future functionality that may live in the new Settings page. The About page isn't particularly important, so we move it to a subpage of the Settings page.

---

UDENG-10906

### DIFF
--- a/packages/app_center/lib/settings/settings.dart
+++ b/packages/app_center/lib/settings/settings.dart
@@ -1,0 +1,1 @@
+export 'settings_page.dart';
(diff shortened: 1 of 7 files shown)


ashuntu:
This could wait until we have more content on this page

matthew-hagemann:
> This could wait until we have more content on this page

@ashuntu Yea, feels a bit empty on its own. What else are we planning to land in Settings?

ashuntu:
Eventually, things like update cadence and things like that.

PR #2141, opened 2026-07-20
chore: Add melos start command

ashuntu:
Adds a `melos start` command to more easily bring up the dev UI without needing to `cd` around or know what commands you need to run.

Would probably be nice to add this in more places, if approved.


d-loose (approved):
Very useful, thanks

PR #2143, opened 2026-07-22
chore: send snap id instead of title

M7mdisk:
Send the snap id instead of title, this will allow us to find the reported snap more easily in the reports sheet.


d-loose (approved):
Thanks, LGTM

PR #2145, opened 2026-07-23
fix(manage): update all snaps and debs with a single auth prompt

### LINKED ISSUE #2144: Repeated Polkit password prompts for each package when clicking "Update All" in Manage page
### Is there an existing issue for this?

- [x] I have searched the existing issues and found none matching what I'm reporting.

### Bug Description

When clicking the **"Update All"** button on the Manage page, Polkit repeatedly prompts the user to enter their administrator password for every single snap and deb package scheduled for update. 

On systems with multiple updates available, this results in a barrage of sequential/parallel Polkit authentication popups.

**Root Cause:**
- `SnapUpdatesModel.refreshAll()` iterates over all refreshable snaps and invokes `refreshSnap()` per snap sequentially/in parallel, triggering separate HTTP POST requests to `snapd`. Each request causes an individual Polkit authentication challenge (`org.freedesktop.snapd.manage`).
- `LocalDebUpdatesModel.updateAll()` iterates over pending deb updates in a `for` loop calling `updateDeb()` per package, initiating separate D-Bus transactions to PackageKit (`org.freedesktop.packagekit.package-update`).

Both client libraries (`package:snapd` and `PackageKitService`) already support native batch operations (`refreshMany` and `updateAll`).

### Steps to reproduce the behavior

1. Open App Center.
2. Go to the **Manage** tab (Manage / Updates).
3. Ensure there are multiple pending updates (snaps and/or deb packages).
4. Click on **"Update All"**.
5. Observe that a Polkit password authentication dialog pops up repeatedly for every individual package being updated.

### Expected behavior

Polkit should prompt for authentication at most **once for snaps** and **once for deb packages** when clicking "Update All", rather than prompting once for every single package.

### Ubuntu release

26.04 LTS (Resolute Raccoon)

### System info

_No response_

### Additional context

_No response_

vicajilau:
Fixes #2144.

Updating everything asks for the polkit password once per package, because each snap and each deb is refreshed in its own operation. This uses snapd's `refreshMany` and a single PackageKit transaction instead, so the whole batch is authenticated once.

### Keeping cancellation working

Going from N operations to one loses the per-item handle that cancelling relied on:

- `cancelRefreshAll` aborted the change owned by each `SnapModel`, and with `refreshMany` no `SnapModel` owns one, so `cancel()` returned at its `activeChangeId == null` guard.
- `cancelAll` called `cancelTransaction(debId)` per deb, which returns at its `activeTransactionId == null` guard once `updateAll` no longer records one.

Both would have become no-ops, and both then clear the "currently updating" provider the UI watches, so pressing Cancel would have stopped showing the update while snapd and PackageKit carried on in the background.

So this tracks the id of the single operation instead, in `refreshAllSnapsChangeIdProvider` and `updateAllDebsTransactionProvider`, and cancels that. `PackageKitService.updateAll` now returns its transaction id, the way `update` already did, so there is something to cancel. Both paths have a regression test that fails without the fix.

### Reconciling with #2179

#2179 landed while this was open and hardened the sequential deb loop: a failing deb no longer stops the ones after it, and a user cancellation does, so that the next deb does not prompt for auth again. It also introduced `PackageKitTransactionCancelled`, which this PR now uses to keep a cancellation from being reported as an error.

The two designs cannot both hold. With one transaction, apt either applies the whole set or none of it, so "the others carry on after one fails" has no equivalent and its test is removed here. The cancellation test keeps its intent, that cancelling is quiet and leaves the list as it was, and now asserts it against the single transaction.

That is the trade this PR makes: one password prompt in exchange for all-or-nothing on the deb side. It is how `apt upgrade` behaves from a terminal, and the prompts are what #2144 reports. But it does undo a property #2179 deliberately added, so it seemed worth stating rather than leaving for a reviewer to discover.

### Snaps: `refreshMany([])`, as #1663 did

An earlier version of this PR passed the snap names filtered by `refreshInhibit == null`, and I argued that was enough to keep running snaps out of the batch. It was not. The list in state comes from `find(filter: refresh)`, which is store data, and snapd does not put `refresh-inhibit` in store results at all, so the filter passed everything through. That is the situation #1663 was written to avoid.

The app cannot see which snaps are inhibited or held from that list, and snapd can, so this passes an empty list and lets snapd decide, which restores what #1663 landed. The change id is still tracked so that Cancel aborts it.

Errors are no longer collected per item: a batch succeeds or fails as a whole, and snapd and PackageKit report which package broke. That leaves `ConsolidatedSnapdException` and `ErrorMessage.fromMap` unreachable. Left in place as part of the error-message API rather than of this change.


d-loose:
Just for context - we had a similar [discussion](https://github.com/ubuntu/app-center/issues/1656) quite a while ago (only focusing on snap updates) and landed a fix [here](https://github.com/ubuntu/app-center/pull/1663), but I suppose this got lost when deb package support was added.
@matthew-hagemann @ashuntu ^

vicajilau:
Thanks for the pointers @d-loose, that explains me a lot! #1663 shows the code did use `refreshMany`, so this is closer to restoring behaviour that regressed when debs arrived than to proposing something new.

One difference worth deciding on. #1663 passed an empty list, letting snapd refresh whatever it can so that a running snap doesn't block the rest. This PR passes explicit names filtered by `refreshInhibit == null`.

I didn't restore the empty list because it no longer maps cleanly onto today's model. `fetchRefreshableSnaps` excludes `kSnapName`, so the store would end up refreshing itself outside the "Quit to update" flow. And `refreshInhibit` didn't exist in 2024, it already identifies the snaps that can't refresh, so filtering on it addresses #1656 without giving up control over what gets touched.

Happy to switch to `refreshMany([])` if you'd rather keep the earlier semantics. Just wanted to flag the self-update interaction first.

vicajilau:
Rebased on current main and green locally (test, analyze --fatal-infos, format:exclude).

No rush on this, but the question above is still open: whether to keep the explicit name list filtered by `refreshInhibit == null`, or restore the `refreshMany([])` semantics from #1663. Happy to switch, I just did not want to change it without knowing how you want the self-update interaction handled.

vicajilau:
Closing the open question here so that this stops reading as blocked.

I have kept the explicit name list filtered by `refreshInhibit == null` rather than restoring `refreshMany([])` from #1663. The reasons are in my earlier comment, and #2061 points the same way: the project is moving toward controlling which snaps enter the list, and an empty list would refresh snaps that filter had just excluded. If you would rather have the #1663 semantics back, it is a one-line change and I will make it in review.

For anyone picking this up cold, the PR does two things:

1. Updates all snaps in one `refreshMany` change and all debs in one PackageKit transaction, so the polkit prompt appears once instead of once per package.
2. Tracks the id of that single operation so that Cancel still cancels. The naive version of this change silently broke cancellation on both sides, and there is a regression test for each.

Green on current main, no conflicts with anything open. @matthew-hagemann @ashuntu, d-loose handed this over on 11 August. Ready when you are.

vicajilau:
Rebased on current main and closing the open question here, so that this stops reading as blocked.

Two things changed since the last round.

**#2179 landed and touched the same function.** It hardened the sequential deb loop so that one failing deb does not stop the rest. With all debs in a single transaction that property has no equivalent, apt applies the whole set or none of it, so its test is removed and the reasoning is in the description under "Reconciling with #2179". The cancellation test keeps its intent, and this now uses `PackageKitTransactionCancelled` from #2179 so that cancelling stays quiet. That is a real trade, one prompt for all-or-nothing, and I would rather you see it stated than find it in the diff.

**On the explicit name list versus `refreshMany([])`:** I have kept the explicit list. The reasons are in my earlier comment, and #2061 points the same way. If you would rather have the #1663 semantics back, it is a one-line change and I will make it in review.

For anyone picking this up cold, the PR does two things: one authentication for the whole batch, and tracking the id of that single operation so that Cancel still cancels. The naive version silently broke cancellation on both sides, and there is a regression test for each.

CI is green on current main. The one red mark is the `osv-scan` job, where the scan itself passed and only the upload to code-scanning failed on a GitHub API error, so it needs a rerun from someone with access rather than a change here.

@matthew-hagemann @ashuntu, d-loose handed this over on 11 August. Ready when you are.

vicajilau:
A correction to something I said above, and a fix for it.

I argued that filtering on `refreshInhibit == null` was enough to keep running snaps out of `refreshMany`, so the explicit list was as safe as `refreshMany([])`. That was wrong. The snaps in state come from `find(filter: refresh)`, which is store data, and snapd does not put `refresh-inhibit` in store results at all. This is the full key set of a `find?select=refresh` result on a 26.04 machine with three updates pending:

```
apps, base, channel, confinement, contact, description, developer, devmode,
download-size, grade, icon, id, ignore-validation, jailmode, license, links,
media, name, private, publisher, revision, status, summary, title, type,
version, website
```

So the filter passed everything, and a snap with its app running went into the batch anyway. That is the case #1663 was written to avoid, and this PR would have reintroduced it.

The last commit fixes it by asking snapd which snaps are inhibited, through `refreshInhibitSnapsProvider`, the same call the manage page already uses for the "Quit to update" notice, and excluding those by name. The new test puts one inhibited and one idle snap in the list and checks that only the idle one is refreshed. Against the old filter it fails with `refreshMany([running, idle])`.

With that, the explicit list does what I claimed it did. But the finding also strengthens the case for `refreshMany([])`, since snapd already knows all of this and the app would not need to. If you would rather go that way, I will switch.

vicajilau:
Switched to `refreshMany([])`. @d-loose, you were right, and here is what settled it.

An earlier version of this PR filtered the candidates with `refreshInhibit == null` and I argued that was enough to keep running snaps out of the batch. It was not. The list in state comes from `find(filter: refresh)`, which is store data. This is the complete key set of a `find?select=refresh` result on snapd 2.76.3 with three updates pending:

```
apps, base, channel, confinement, contact, description, developer, devmode,
download-size, grade, icon, id, ignore-validation, jailmode, license, links,
media, name, private, publisher, revision, status, summary, title, type,
version, website
```

`refresh-inhibit` is not there. It is local state, so the filter passed every snap through, and a test with one inhibited and one idle snap confirmed it: `refreshMany([running, idle])`. That is the situation #1663 was written to avoid.

The app cannot see from that list which snaps are inhibited or held, and snapd can, so this passes an empty list and lets snapd decide, which is what #1663 landed and what its description relies on: running snaps do not block the rest and stay in the list. What stays here is tracking the change id so that Cancel aborts it, and the single PackageKit transaction on the deb side.

Description updated. Green on current main.

PR #2149, opened 2026-08-05
fix(manage): stop rebuilding the installed apps list twice per keystroke

### LINKED ISSUE #2069: Bug: Manage page filter can loop infinitely
The installed apps filter on the manage page has a self-triggering loop.

**File:** `manage/app_providers.dart` (InstalledApps class)

**Problem:** When a filter changes, it re-applies filters, which triggers the filter listener again. With no guard condition, this can loop forever causing UI freezes or flickering.

**Fix:** Add a guard that skips re-filtering if the result hasn't changed, or debounce the filter application.

vicajilau:
Closes #2069, though not as that issue describes it, see below.
`_onSearchChanged` writes the search text to both `localSnapFilterProvider` and `localDebFilterProvider`, and `InstalledApps.build` listens to both. But `_applyFilters` only ever reads the snap one:

```dart
final filter = ref.read(localSnapFilterProvider).toLowerCase();
final showSystemApps = ref.read(showLocalSystemAppsProvider);
final sortOrder = ref.read(appSortOrderProvider);
final packageType = ref.read(packageTypeFilterProvider);
```

So the second write re-runs the filter and pushes a fresh `AsyncData` holding a list identical to the one already in state. A full rebuild of the installed apps list, once per keystroke, for nothing.

`localDebFilterProvider` turns out to be written and listened to but never read anywhere in the tree, so removing it and its listener drops the redundant rebuild at the source rather than filtering it out afterwards. `showLocalSystemDebsProvider`, declared on the next line, is never referenced at all. I removed it in the same pass since it is the same leftover. Both look like remnants from when snaps and debs had separate lists.

No test accompanies this: what it removes is unreachable, and the analyzer passing with the declarations gone is the check that nothing used them.

### On the issue

#2069 describes this as a listener loop that can spin forever. That loop doesn't exist, `_applyFilters` only writes `InstalledApps`' own state, nothing writes back into the filter providers, and the search field already debounces at 200 ms. The real cost was the duplicate rebuild above. Happy to close #2069 as invalid instead if you would rather keep the two separate.

### One question

Was the deb filter meant to be read at some point, a separate search box for debs, say? Nothing in the tree suggests it, and today a single field writes the same value to both, so it is redundant either way. But you would know the intent better than I do; if it is a placeholder for planned work, say so and I will drop that part and only remove the listener.

### DIFF
--- a/packages/app_center/lib/manage/app_providers.dart
+++ b/packages/app_center/lib/manage/app_providers.dart
@@ -95,7 +95,6 @@ class InstalledApps extends _$InstalledApps {
 
     void refreshFunction(_, __) => _applyFilters();
     ref.listen(localSnapFilterProvider, refreshFunction);
-    ref.listen(localDebFilterProvider, refreshFunction);
     ref.listen(showLocalSystemAppsProvider, refreshFunction);
     ref.listen(appSortOrderProvider, refreshFunction);
     ref.listen(packageTypeFilterProvider, refreshFunction);
--- a/packages/app_center/lib/manage/local_deb_providers.dart
+++ b/packages/app_center/lib/manage/local_deb_providers.dart
@@ -64,9 +64,6 @@ class LocalDebInfo with _$LocalDebInfo {
   }
 }
 
-final localDebFilterProvider = StateProvider.autoDispose<String>((_) => '');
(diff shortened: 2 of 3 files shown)


ogres (approved):
Looks good, thank you for the contribution @vicajilau

Issue #2139, opened 2026-07-19
Crash on startup: Unhandled exception FormatException: Unknown release type 'snapshot'


pawseytj:
### Is there an existing issue for this?

- [x] I have searched the existing issues and found none matching what I'm reporting.

### Bug Description

When launching the App Center, the application fails to load the Manage tab (spinning loading icon) or crashes via terminal with a Dart isolate exception. The crash is caused by the AppStream parser hitting an unknown release type 'snapshot' in the local metadata cache.

### Steps to reproduce the behavior

1. Run `snap-store` in the terminal.
2. Observe the crash output.
(Note: Clearing the `/var/cache/swcatalog/` and `app-info` directories allows the store to open until `apt update` pulls the metadata again).

### Expected behavior

The App Center should gracefully ignore unrecognized release types in the AppStream metadata rather than crashing the isolate.

### Ubuntu release

Ubuntu 26.04 LTS

### System info

7.0.0-28-generic #28-Ubuntu SMP PREEMPT_DYNAMIC Sun Jun 21 01:01:36 UTC 2026 x86_64
name:      snap-store
summary:   Discover, install and manage Linux apps
publisher: Canonical✓
store-url: https://snapcraft.io/snap-store
license:   unset
description: |
  The App Center showcases featured and popular applications with useful
  descriptions, ratings, and screenshots.
  
  Apps can be found either through browsing categories or by searching.
  
  The App Center can also be used to update apps, switch channels, and view
  and submit ratings.
  
  In older Ubuntu releases, the App Center is called Snap Store and is based
  on GNOME Software.
commands:
  - snap-store.show-updates
  - snap-store
services:
  snap-store.packagekit-session-installer: dbus, enabled, inactive
snap-id:      gjf3IPXoRiipCu9K0kVu52f0H56fIksg
tracking:     2/candidate
refresh-date: 2026-04-30
channels:
  2/stable:          0+git.10310e85   2026-04-22 (1367) 16.4MB -
  2/candidate:       ↑                                         
  2/beta:            ↑                                         
  2/edge:            0+git.c1a8a57b   2026-06-25 (1382) 12.4MB -
  latest/stable:     41.3-72-g80e7130 2024-09-22 (1216) 12.8MB -
  latest/candidate:  ↑                                         
  latest/beta:       ↑                                         
  latest/edge:       0+git.c1a8a57b   2026-06-25 (1382) 12.4MB -
  preview/stable:    –                                         
[...]

PR #2151, opened 2026-08-06
fix(appstream): bound the metadata load with a timeout

vicajilau:
Related to #2139 and #2142, but deliberately not "Fixes" either of them.

Metadata parsing happens in isolates the pool spawns, so a parse that neither completes nor throws leaves `load()` pending forever. Everything waiting on the metadata waits with it: `componentsByPackage` → `localDebs` → `installedApps`/`appUpdates` all stay in loading, and the manage page sits on a spinner with nothing in the log to explain it. Traced on 26.04: PackageKit logs `get-packages` succeeding, and then no further call is ever made, because the chain stalls at the AppStream step.

This caps the load at 60 s and logs the failure, so pages that need the metadata surface an error with a retry instead of hanging. 60 s is deliberately generous — a real catalog here loads in 3.4 s.

### Scope: what this is and isn't

The catalog entry causing this in practice is a `snapshot` release type the `appstream` package didn't know about. That is fixed at the root in canonical/appstream.dart#44, and #2156 already ships it by pinning the dependency to that commit — released 7 August as revision 1390 on `2/stable`. So #2139 and #2142 are fixed in practice today; what remains there is swapping the git ref for a normal constraint once 0.2.11 reaches pub.dev.

#2153 covers the other axis: per-file resilience, so a single unparseable catalog costs one component instead of all the metadata. That is a better outcome than this PR gives on its own, and the two don't conflict.

What this adds that neither of those does is a bound on the failure mode itself. Both of them work by turning a bad parse into an error — which only helps when the parse actually fails. A parse that hangs without throwing, or an isolate that dies outside the handler, still leaves `load()` pending forever. The cap is the only thing that makes that visible.

Happy to close this if you would rather carry the other two alone.

### DIFF
--- a/packages/app_center/lib/appstream/appstream_service.dart
+++ b/packages/app_center/lib/appstream/appstream_service.dart
@@ -137,19 +137,42 @@ class _ScoredComponent {
 
 class AppstreamService {
   // TODO: cache AppstreamPool
-  AppstreamService({@visibleForTesting AppstreamPool? pool})
-    : _pool = pool ?? AppstreamPool(),
-      _l10n = _loadL10n() {
+  AppstreamService({
+    @visibleForTesting AppstreamPool? pool,
+    @visibleForTesting this._loadTimeout = _defaultLoadTimeout,
+  }) : _pool = pool ?? AppstreamPool(),
+       _l10n = _loadL10n() {
     PlatformDispatcher.instance.onLocaleChanged = () async {
       await _loader;
       _populateCache();
     };
   }
+
+  /// A catalog load that never returns must not be able to wedge the app.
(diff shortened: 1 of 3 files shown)


vicajilau:
Closing this. #2156 shipped the root fix by pinning `appstream` to the commit that adds the missing release type and stops the isolate hanging, so the failure this guarded against no longer happens in practice. #2153 covers the remaining axis — per-file resilience — and does it better than a timeout would.

What is left here is a backstop for a parse that hangs without throwing, which is real but hypothetical. Not worth reviewer time on its own; happy to bring it back if a hang ever shows up that the two above don't catch.

PR #2156, opened 2026-08-07
fix: bump appstream.dart

matthew-hagemann:
## Summary

Fixes a parsing error that was occurring due to the inclusion of the `snapshot` field, which appstream.dart did not know how to parse.

Points the `appstream` dependency at the canonical/appstream.dart git ref `12c359e5ca0a40f954025f9a42f59799b292fcee` until a new version is published to pub.dev.

## Before

```sh
...
FormatException: Unknown release type 'snapshot'
#0      _parseReleaseType (package:appstream/src/collection.dart:944)
#1      new AppstreamCollection.fromYaml (package:appstream/src/collection.dart:641)
#2      AppstreamPool._loadYamlCollectionInIsolate (package:appstream/src/pool.dart:94)
```

[Screencast From 2026-08-07 17-12-53.webm](https://github.com/user-attachments/assets/9e703ee8-f76e-4658-9052-7e155dc88297)

## After
[Screencast From 2026-08-07 17-23-11.webm](https://github.com/user-attachments/assets/2cc5ae6f-10c0-42d2-a5e9-4c70d60dbc00)

### DIFF
--- a/.github/actions/setup-fvm/action.yaml
+++ b/.github/actions/setup-fvm/action.yaml
@@ -1,0 +1,39 @@
+name: Setup FVM
+description: >
+  Installs FVM and the Flutter SDK version pinned in .fvmrc.
+
+  Equivalent to Atsumi3/actions-setup-fvm, except that FVM is installed from its
+  GitHub release using the repository's own scripts/install-fvm.sh, the same way
+  the snap build does it. That action installs FVM from the leoafarias/fvm
+  Homebrew tap, which Homebrew now ignores because third-party taps require
+  explicit trust, so every job using it fails at setup.
+
+inputs:
+  fvm-version:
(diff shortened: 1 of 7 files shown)


didrocks (approved):
I would have liked that the fvm modification from the 3rd party action to the local one be decoupled from bumping appstream.dart explicitely, but let’s make this smooth and ack for this time ;)

I have checked the apptstream unreleased ref and the change makes sense with the tests. However, we need to see in the future how we can prevent this by changing it to be more forward-compatible. I think your commit there is managing that with "Race the result against the error port so a failure to parse surfaces as an exception instead." but that can be a  discussion and confirmation next week.

Also, we will need a Yarf test for those as this was not spotten, can you create a ticket for this?

Thanks a lot for the changes and the video of the result!

PR #2167, opened 2026-08-18
application: Use a transparent background as view bg color

3v1n0:
Flutter apps will draw by default using a black background, this is because it cannot be known what will be the application main color before than drawing it.

For flutter apps that are using the yaru toolkit we can be confident that the bg color will match the gtk one so we can just show a transparent background rather than painting a black one while the app is loading


ogres (approved):

ogres:
@3v1n0 Thanks for adding this, could you please sign your commit?

PR #2170, opened 2026-08-21
feat: Add driver management page

ashuntu:
Adds the driver management page.

This allows users to install drivers for devices that require them, and switch NVIDIA device branches.

<details>
<summary>Screenshots</summary>

image4

image3

image2

image1

Snapd doesn't support driver interface:
image

System with no devices that need drivers:
image

</details>

**Reviewers**:

You need to run `ubuntu-drivers-common` D-Bus service from source (on Resolute or Stonking) since it is not all in Stonking yet:

1. `git clone https://github.com/canonical/ubuntu-drivers-common`
2. `sudo cp systemd/com.ubuntu.Drivers.conf /usr/share/dbus-1/system.d/`
3. `sudo systemctl reload dbus`
4. `sudo PYTHONPATH=$(pwd) python3 ./ubuntu-drivers-dbus-service` (you may need some python deps installed on your system, depending on your setup)

To test the snap build, you also need snapd from edge with `ubuntu-drivers-observe` connected:

1. `sudo snap refresh snapd --edge`
2. Build and install the snap from this branch, e.g. `snapcraft pack`, `snapcraft install *.snap --dangerous`
3. `sudo snap connect snap-store:ubuntu-drivers-observe`

Then you should be able to test this PR. Note that the D-Bus service will stop after 5 minutes of inactivity, so you will need to rerun it if it stops.

You will not see anything very useful on the page if your device doesn't need drivers. Good candidates to test are devices with NVIDIA graphics cards. Note that the service also filters DKMS drivers, so you won't see any drivers available on stonking yet.

---

UDENG-11228


juanruitina:
Some early feedback:

- [ ] Can we make the header match the design? I would like for these pages something closer to the Security Center (and hence GNOME), and the app search doesn't make much sense in this section, I think.
- [ ] Similarly, please make the subheadings "Installed" similar to those in the Security Center (titleSmall)
- [ ] It may just be the example, but am I right that we are not showing the vendor name before the device name?
- [ ] In the branch dialog, the subtitles seem to be too large compared to the design, can you check?

ashuntu:
@juanruitina 

> Can we make the header match the design? I would like for these pages something closer to the Security Center (and hence GNOME), and the app search doesn't make much sense in this section, I think.

https://github.com/ubuntu/app-center/pull/2170/commits/0ee765954354df3193723c251a339de143d27aaf

> Similarly, please make the subheadings "Installed" similar to those in the Security Center (titleSmall)

The designs use `body-medium` I believe for those elements (still incorrect here so I'll adjust regardless), should I use that or `title-small`?

juanruitina:
@ashuntu `titleSmall` please, it was an issue with the Figma component. It should look the same regardless.

matthew-hagemann (commented):
Looking good! I'll go through the code more thoroughly, but so far so good. Here is the flow on my pc, that had a pending nvidia driver update:

[Screencast From 2026-09-18 17-35-15.webm](https://github.com/user-attachments/assets/786195b1-bfc1-4fb2-8c9b-d3a5c5761212)

ashuntu:
> Looking good! I'll go through the code more thoroughly, but so far so good. Here is the flow on my pc, that had a pending nvidia driver update:

If you had the driver installed already and just needed and update, it should appear under an Update available section and prompt to update and not install. I think I encountered this as well, maybe the check for already being installed is too strict :thinking:

juanruitina:
Two minor comments:

- [ ] "Switch channel" should always be under the "More options" menu (under the ellipsis button). If the user clicks on "Install" straight away, the recommended driver (production) should install automatically without further confirmation.
- [ ] In the description in the driver row, let's show the branch name _only_ if it's not the recommended one.

But this is looking tight already :)

ashuntu:
>  "Switch channel" should always be under the "More options" menu (under the ellipsis button). If the user clicks on "Install" straight away, the recommended driver (production) should install automatically without further confirmation.
In the description in the driver row, let's show the branch name only if it's not the recommended one.

Both should be resolved.

matthew-hagemann (dismissed):
Current state LGTM from a code perspective. I wasn't sure if you wanted to make any behavior changes based on your last comment, or if we want to call for more testers. I know Sergo was happy to give it a run aswell if we want more hardware variety :slightly_smiling_face: 

Will approve at this checkpoint so if anything is added we can review from here onward.

ashuntu:
> I wasn't sure if you wanted to make any behavior changes

The only thing would be to disable the page on old installs without `ubuntu-drivers-common` D-Bus service, working on that now

PR #2150, opened 2026-08-05
fix(deb): stop re-showing the last PackageKit error

vicajilau:
Fixes #2113.

`DebModel._onError` stores the error in the model state, and `DebPage` surfaces it from a post-frame callback registered on every build. Nothing ever cleared it, and the notifier is `keepAlive`, so once an action failed the error stayed there: the state change from the next action rebuilt the page and popped the same dialog again before that action had a chance to complete.

Errors are one-shot events, so the page now consumes them once surfaced.

Tests cover `clearError` in the model and assert on the page that the error doesn't survive being shown.

PR #2171, opened 2026-08-22
fix: Don't let a failed catalog load hide snap updates

metehankaygsz:
## Problem

`AppstreamPool.load()` runs every catalog file through a single
`Future.wait`, so one file it cannot parse fails the entire load. Since
0.2.11 the package surfaces that as `AppstreamCollectionLoadException`
rather than hanging, but the whole-load failure remains.

App Center then amplifies it. The failure propagates
`componentsByPackage` -> `localDebs` -> `localDebUpdatesModel`, and
`appUpdates` awaits the snap and deb sources in the same function body:

```dart
final snapUpdates = await ref.watch(snapUpdatesModelProvider.future);
final debUpdates = await ref.watch(localDebUpdatesModelProvider.future);
```

So an unreadable deb catalog takes the *snap* updates down with it, even
though snapd has updates to offer and was never involved in the failure.
An unavailable PackageKit does the same thing.

This is close to what #2156 fixed: a `snapshot` release type that the
pinned `appstream` did not recognise made the catalog unparseable, and
users saw an App Center that listed no updates at all. That specific
trigger is gone, but the amplification that turned a metadata problem
into "no updates" is still here, waiting for the next catalog file the
parser cannot handle — a future release type, or metadata from a
third-party repo.

This PR does not change any parsing. It limits the blast radius.

## Changes

- `AppstreamService` logs a pool load failure and continues with an empty
  catalog instead of propagating it to every consumer.
- `appUpdates` and `InstalledApps` treat snaps and debs as independent
  sources, so a failure in one no longer hides the other.
- A warning info box on the manage page tells the user the lists only
  cover snaps, so the degraded state is visible rather than silent.

The third point is deliberate: a store that quietly lists fewer updates
than exist is its own hazard. The existing error stream was not usable
for it — the listener in `store_app.dart` only forwards `SnapdException`,
and widening that would route every uncaught zone error into a modal. An
inline box on the affected page says it where it matters. It is a
separate commit if you would rather not take it.

## Testing

New tests, each of which fails without the corresponding change:

- `appstream_service_test`: the service still initialises when the pool
  throws.
- `manage_models_test`: snap updates and installed snaps still resolve
  when the deb lookup throws; `debSourcesAvailable` reflects the failure.
- `manage_page_test`: the warning renders when the deb sources fail, and
  the snap update tile is still listed alongside it.

`melos test`, `melos analyze --fatal-infos` and `melos format:exclude`
all pass.

### DIFF
--- a/packages/app_center/lib/appstream/appstream_service.dart
+++ b/packages/app_center/lib/appstream/appstream_service.dart
@@ -159,6 +159,7 @@ class AppstreamService {
       .load()
       .onError<Exception>((error, _) {
         log.error('Failed to load the Appstream pool', error);
+        _catalogLoadFailed = true;
       })
       .then((_) {
         _populateCache();
@@ -167,6 +168,14 @@ class AppstreamService {
 
   bool get initialized => _initialized;
   bool _initialized = false;
+
+  /// Whether the catalog failed to load.
+  ///
+  /// A failed load leaves the pool empty, which is indistinguishable from a
+  /// system with no Appstream metadata. Callers that need to tell the user
+  /// their view is incomplete have to ask.
+  bool get catalogLoadFailed => _catalogLoadFailed;
+  bool _catalogLoadFailed = false;
 
   final HashSet<_CachedComponent> _cache = HashSet<_CachedComponent>();
 
--- a/packages/app_center/lib/appstream/appstream_service.dart
+++ b/packages/app_center/lib/appstream/appstream_service.dart
@@ -146,10 +146,24 @@ class AppstreamService {
     };
   }
   final AppstreamPool _pool;
-  late final Future<void> _loader = _pool.load().then((_) {
-    _populateCache();
-    _initialized = true;
-  });
+
+  /// Loads the pool, tolerating a failure to read the system catalogs.
+  ///
+  /// [AppstreamPool.load] aborts as soon as a single catalog file fails to
+  /// parse, e.g. because it contains metadata newer than the `appstream`
+  /// package understands. Letting that propagate would take down every
+  /// consumer of this service - including the deb providers that the manage
+  /// page's update list waits on - so the failure is logged and the service
+  /// carries on with whatever the pool managed to collect.
+  late final Future<void> _loader = _pool
+      .load()
+      .onError<Exception>((error, _) {
+        log.error('Failed to load the Appstream pool', error);
+      })
+      .then((_) {
+        _populateCache();
+        _initialized = true;
+      });
 
   bool get initialized => _initialized;
   bool _initialized = false;
--- a/packages/app_center/lib/manage/app_providers.dart
+++ b/packages/app_center/lib/manage/app_providers.dart
@@ -1,3 +1,4 @@
+import 'package:app_center/appstream/appstream.dart';
 import 'package:app_center/l10n.dart';
 import 'package:app_center/manage/local_deb_providers.dart';
 import 'package:app_center/manage/local_deb_updates_model.dart';
@@ -73,10 +74,14 @@ Future<List<ManageAppData>> appUpdates(Ref ref) async {
 Future<bool> debSourcesAvailable(Ref ref) async {
   try {
     await ref.watch(localDebsProvider.future);
-    return true;
   } on Exception {
     return false;
   }
+  // A catalog that failed to load leaves the service with no components, so
+  // the deb list comes back empty rather than throwing. Ask the service
+  // directly, otherwise the very failure this warning exists for is the one
+  // case it would miss.
+  return !getService<AppstreamService>().catalogLoadFailed;
 }
 
 /// Returns all installed debs, or an empty list if they could not be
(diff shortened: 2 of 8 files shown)


ogres:
@metehankaygsz Could you please add screenshot how it looks like?
We need to think about it thoroughly - it might not be the best user experience to just tell them "Information was unavailable".

metehankaygsz:
> @metehankaygsz Could you please add screenshot how it looks like? We need to think about it thoroughly - it might not be the best user experience to just tell them "Information was unavailable".

test_error


Screenshot attached, this is a local build with a catalog file that fails to parse, so it's the real path and not a mock-up.

And fair point on the wording, it tells you something broke but not what to do about it. One idea: there's already `managePageDebUpdatesMessage` ("Debian package updates are handled by the Software Updater."), so maybe something like "Debian package updates can't be shown right now — you can check them in Software Updater." Happy to change it to whatever you think fits best, or add a retry button instead.

ogres:
@juanruitina What do you think? How should we handle case if one of the sources fail for some reason?

juanruitina:
Can you give me some context about why could sources fail, real-life scenarios? Could this happen with snaps too? That will help me assess.

metehankaygsz:
> Can you give me some context about why could sources fail, real-life scenarios? Could this happen with snaps too? That will help me assess.


Sure! The one that led to this is on 26.04, the AppStream data in `resolute-updates/main` has a Nautilus entry with release `type: snapshot`, and the appstream version we had pinned didn't know that type. The pool loads every catalog file in one `Future.wait`, so that one entry failed the whole load and App Center showed no updates at all (#2139, #2142). #2156 fixed it by bumping the lib, but it could happen again whenever AppStream adds something new, and PPAs ship their own metadata too. Other case is PackageKit not being available over D-Bus.

For snaps it's a different path, they come from snapd locally and don't touch AppStream, so a catalog problem can't affect them. No network is already handled separately. Other snapd errors do still blank the whole page though, and I only handled the deb direction here.

vicajilau (commented):
I looked at this because I worked on the upstream side of the parsing bug it references (canonical/appstream.dart#44). The diagnosis is right and the layer you picked is the one that matters. The parsing fix stopped the hang, but the amplification you describe was always the more general problem and it outlives that specific trigger.

The `debSourcesAvailable` detail is the part I would not have got right first time:

> A catalog that failed to load leaves the service with no components, so the deb list comes back empty rather than throwing.

A try/catch around `localDebs` misses precisely the failure the warning exists for, because an empty pool comes back as success. Asking the service directly is the right call.

Two things worth deciding before this lands.

**The warning is on Manage, but the tolerance is global.** `AppstreamService` now swallows the load failure for every consumer, not just the manage page, and the other consumers do not all degrade gracefully. `getFromId` ends in:

```dart
assert(component != null, 'Component not found');
return component!;
```

With an empty catalog that is a null check failure rather than a degraded page, and `DebModel.build` calls it on every deb page. Reachability is limited once the catalog is gone, since search and the manage list come back empty, but an `appstream://` or `apt://` handler or a route restored from history still gets there. Before this PR the failure surfaced earlier as an error view. After it, it surfaces deeper and as a crash.

**`onError<Exception>` does not catch `Error`.** `AppstreamCollectionLoadException` implements `Exception`, so the case you are targeting is covered. The null check above is an `Error`, and so is anything else the pool throws that is not an exception, and those still take the page down. If the intent is "the service survives a bad catalog", `onError<Object>` matches it more closely.

Neither has to be solved here. But if the answer is "later", it is worth a line in the code so the next person knows the gap is known rather than missed.

For what it is worth, I have two open PRs touching the same files (#2145, #2150) and this merges cleanly against both.

juanruitina:
Thank you both. Adding the UX angle here. Debs failing to load should definitely not block snap updates. But it would be great if we could find a more resilient handling of these changes from AppStream, instead of causing (and displaying) an error. If there are problems because of the labelling for one single app, I'd consider it equally unacceptable that it's blocking the _other_ deb packages, that are already installed anyway.

Having the error could still useful. I'd go for something like this, I wouldn't mention _the other_ package type:

> Some apps may be missing
> Debian package data is not currently available.

Not actionable, but there's little the user can do anyway; the system should just behave better and I wouldn't have them do workarounds. The App Center should handle this regardless of the Software Updater being available in the system; the goal is to have all updates by perfectly manageable from here.

I imagine an alternative could be to list all deb packages with updates even without AppStream data being available, grouped somehow, but that will require further work.

vicajilau:
@juanruitina the direction you describe may be closer than it looks. `LocalDebInfo.component` is already nullable and there is a `hasAppstreamEntry` getter for it, so the model already anticipates debs without AppStream data. What excludes them is one filter in `localDebs`:

```dart
// Filter to packages with Appstream entries
final guiPackages = installed
    .where((pkg) => componentsByPkg.containsKey(pkg.packageId.name))
    .toList();
```

The catch is what happens when one of those is opened. `DebModel.build` resolves its component through `AppstreamService.getFromId`, which ends in:

```dart
assert(component != null, 'Component not found');
return component!;
```

So listing them without also making that path tolerate a missing component turns a missing label into a crash on the detail page.

That is the same line I flagged above for the empty-catalog case, which is why I think it is worth fixing first whichever route this takes. Not asking for any of it in this PR.

PR #2176, opened 2026-09-02
ci: stop tracking the one generated mocks file

vicajilau:
The l10n job currently fails on every pull request, whatever it changes, at the `Check for outdated l10n` step:

```
diff --git a/packages/app_center_ratings_client/test/ratings_client_test.mocks.dart ...
 // ignore_for_file: no_leading_underscores_for_library_prefixes
+
 import 'dart:async' as _i6;

Warning: packages/app_center_ratings_client/test/ratings_client_test.mocks.dart may be outdated
Error: Process completed with exit code 1.
```

`.gitignore` has ignored `**/*.mocks.dart` since bc85f713 ("ci: Generate files and cache them"), but this file predates the rule and stayed in the index, so git kept following it. It is the only generated file the repository still tracks:

```
$ git ls-files | grep -E '\.mocks\.dart$|\.g\.dart$|\.freezed\.dart$'
packages/app_center_ratings_client/test/ratings_client_test.mocks.dart
```

That is what makes it drift. This package does not track `pubspec.lock`, which is the usual call for a publishable package, so CI resolves dependencies fresh and a mockito release that changes the generated output leaves the committed copy stale. Today the difference is one blank line after the `ignore_for_file` comment.

Either choice is reasonable on its own. Letting dependencies float is what a package is supposed to do, and committing generated output is a normal thing to want. Together they cannot hold: the output is regenerated against versions that move, so it goes stale on its own.

Regenerating and committing it would clear the current failure and set up the next one. Untracking it puts the file on the same footing as every other generated file here: the setup job runs `melos generate` and the later jobs pick it up from the repository cache, which is already how `app_center`'s mocks work.

Verified locally that `scripts/check-outdated-files.sh` passes after a full `melos generate` with the file untracked. It is matched by `.gitignore`, so `git ls-files --others --exclude-standard` does not report it as untracked either.

---

Not proposing to change it here, but the same pattern shows up one line earlier in `.gitignore`. Line 33 ignores `pubspec.lock`, while `packages/app_center/pubspec.lock` is tracked because it predates the rule:

```
$ git ls-files | grep pubspec.lock
packages/app_center/pubspec.lock
```

The behaviour that produces is right either way. `app_center` is an application, so committing its lock is the usual call, and the ratings client is publishable, so leaving its lock out is the usual call too. It is only the rule that does not describe what the repository does, which is probably why `check-outdated-files.sh` excludes `':!**/pubspec.lock'` from both of its checks.

Worth knowing about when reading that file, and yours to decide whether it is worth touching.


ogres:
@vicajilau Thanks for finding it!

ogres (approved):

Issue #2169, opened 2026-08-21
Cannot install new updates. LibreOffice and PackageKit.


toni1elias:
### Is there an existing issue for this?

- [x] I have searched the existing issues and found none matching what I'm reporting.

### Bug Description

Today an new update came in the app center for LibreOffice and PackageKit.
When I click update all, the store asks for credentials multiple times and fails to update.
4:26.2.4.2-0ubuntu0.26.04.2 → 4:26.2.5.2-0ubuntu0.26.04.1

Trying to update apps individually returns the following error:

<img width="1600" height="900" alt="Image" src="https://github.com/user-attachments/assets/e682639e-07ae-4b68-af01-92411187c0c0" />

### Steps to reproduce the behavior

1. Open App Center.
2. Manage.
3. Update All.

### Ubuntu release

Ubuntu 26.04 LTS

### System info

7.0.0-30-generic #30-Ubuntu SMP PREEMPT_DYNAMIC Fri Jul 31 18:22:54 UTC 2026 x86_64
name:      snap-store
summary:   Discover, install and manage Linux apps
publisher: Canonical✓
store-url: https://snapcraft.io/snap-store
license:   unset
description: |
  The App Center showcases featured and popular applications with useful descriptions, ratings, and
  screenshots.
  
  Apps can be found either through browsing categories or by searching.
  
  The App Center can also be used to update apps, switch channels, and view and submit ratings.
  
  In older Ubuntu releases, the App Center is called Snap Store and is based on GNOME Software.
commands:
  - snap-store.show-updates
  - snap-store
services:
  snap-store.packagekit-session-installer: dbus, enabled, inactive
snap-id:      gjf3IPXoRiipCu9K0kVu52f0H56fIksg
tracking:     2/stable/ubuntu-26.04
refresh-date: 7 days ago, at 12:41 CEST
channels:
  2/stable:          0+git.8334e2e4   2026-08-07 (1390) 12.4MB -
  2/candidate:       ↑                                         
  2/beta:            ↑                                         
  2/edge:            0+git.ecac45b3   2026-08-18 (1397) 12.4MB -
  latest/stable:     41.3-72-g80e7130 2024-09-22 (1216) 12.8MB -
  latest/candidate:  ↑                                         
  latest/beta:       ↑                                         
  latest/edge:       0+git.ecac45b3   2026-08-18 (1397) 12.4MB -
  preview/stable:    –                                         
[...]

PR #2179, opened 2026-09-03
fix(manage): exclude blocked (phased) deb updates from update lists

Yashpal-0:
## Summary

PackageKit reports phased deb updates with `info=blocked` via `GetUpdates`. These are not installable candidates, so `app-center` listing them in the Manage page leads to update attempts that fail with `PackageKitError.packageNotFound`, repeated polkit auth prompts, and a spinner that never resolves.

```
$ ERROR packagekit: Received PackageKitErrorCodeEvent (PackageKitError.packageNotFound):
  Package id 'nautilus;1:50.2.2-0ubuntu0.1;amd64;ubuntu-resolute-updates-main' is not available as an update candidate

$ apt policy nautilus
  Candidate: 1:50.2.2-0ubuntu0.1
     1:50.2.2-0ubuntu0.1 500 (phased 0%)

$ apt-get -s upgrade
The following upgrades have been deferred due to phasing:
  libnautilus-extension4 nautilus nautilus-data
```

## Changes

**Filter blocked updates at the source** — `PackageKitService.getUpdates()` now skips `PackageKitInfo.blocked` events. Both consumers (`local_deb_providers.dart:88` and `local_deb_updates_model.dart:78`) go through this single call, so filtering once here covers the whole Manage page rather than repeating a guard in each caller.

**Cross-check the deb page against installable updates** — `DebModel._hasUpdate` previously trusted `getUpdateDetails`, whose `UpdateDetail` events carry no blocked flag, so a phased update would still light up the Update button on an individual deb page. It now intersects those package names with the installable set from `getUpdates()`.

This also fixes a pre-existing bug in the same loop: it iterated `updates` but unconditionally `break`-ed after the first entry, so only the first candidate was ever considered and the loop was misleading. The rewrite returns on the first installable match and returns `false` when none is found.

**Handle the update failures that do get through** — a failing `waitTransaction` previously escaped uncaught, leaving `activeTransactionId` set and the UI spinning indefinitely:
- `DebModel._packageKitAction` routes the exception through the existing `_onError` path, which surfaces the error and clears the transaction state.
- `LocalDebUpdatesModel.updateDeb` logs and pushes the exception to `errorStreamControllerProvider`.

Both feed the error plumbing that `deb_page.dart` and the Manage page already consume, so no changes were needed in the widget layer itself.

## Testing

Three new tests:
- `packagekit_service_test.dart` — `getUpdates` excludes blocked updates.
- `deb_model_test.dart` — `hasUpdate` is true when an installable update exists.
- `deb_model_test.dart` — `hasUpdate` is false when the only update is blocked (phased).

`melos test`, `melos analyze --fatal-infos` and `melos format:exclude` all pass.

Related to #2169 and #2142.

---

UDENG-11211

### DIFF
--- a/packages/app_center/lib/deb/deb_model.dart
+++ b/packages/app_center/lib/deb/deb_model.dart
@@ -150,18 +150,24 @@ class DebModel extends _$DebModel {
     final updates = detailsEvent?.updates.where(
       (pid) => pid != packageInfo.packageId,
     );
-    var hasUpdate = false;
+    if (updates == null || updates.isEmpty) return false;
 
-    for (final packageUpdate in updates ?? <PackageKitPackageId>[]) {
+    // getUpdateDetails doesn't flag blocked (e.g. phased) updates, so
+    // cross-check against the installable updates from GetUpdates.
+    final installableNames = (await packageKit.getUpdates())
+        .map((u) => u.packageId.name)
+        .toSet();
+
+    for (final packageUpdate in updates) {
       final packageName = packageUpdate.name;
+      if (!installableNames.contains(packageName)) continue;
       final results = await packageKit.resolve([
         packageName,
       ], installedOnly: true);
-      hasUpdate = results[packageName]?.info == PackageKitInfo.installed;
-      break;
+      return results[packageName]?.info == PackageKitInfo.installed;
     }
 
-    return hasUpdate;
+    return false;
   }
 
   Future<void> _packageKitAction(Future<int> Function() action) async {
@@ -169,7 +175,18 @@ class DebModel extends _$DebModel {
     state = AsyncValue.data(
       state.value!.copyWith(activeTransactionId: transactionId),
     );
-    await packageKit.waitTransaction(transactionId);
+    try {
+      await packageKit.waitTransaction(transactionId);
+    } on Exception catch (e) {
+      // Report via the same path as PackageKitServiceError events so the
+      // page shows the error and clears the stuck transaction state.
+      await _onError(
+        PackageKitServiceError(
+          code: PackageKitError.internalError,
+          details: e.toString(),
+        ),
+      );
+    }
     ref.invalidateSelf();
   }
 }
--- a/packages/app_center/lib/deb/deb_model.dart
+++ b/packages/app_center/lib/deb/deb_model.dart
@@ -152,8 +152,8 @@ class DebModel extends _$DebModel {
     );
     if (updates == null || updates.isEmpty) return false;
 
-    // getUpdateDetails doesn't flag blocked (e.g. phased) updates, so
-    // cross-check against the installable updates from GetUpdates.
+    /* getUpdateDetails doesn't flag blocked (e.g. phased) updates, so
+       cross-check against the installable updates from GetUpdates. */
     final installableNames = (await packageKit.getUpdates())
         .map((u) => u.packageId.name)
         .toSet();
@@ -180,8 +180,8 @@ class DebModel extends _$DebModel {
     } on PackageKitTransactionCancelled {
       // User cancelled (e.g. dismissed the polkit dialog) — not an error.
     } on Exception catch (e) {
-      // Report via the same path as PackageKitServiceError events so the
-      // page shows the error and clears the stuck transaction state.
+      /* Report via the same path as PackageKitServiceError events so the
+         page shows the error and clears the stuck transaction state. */
       await _onError(
         PackageKitServiceError(
           code: PackageKitError.internalError,
--- a/packages/app_center/lib/deb/deb_model.dart
+++ b/packages/app_center/lib/deb/deb_model.dart
@@ -177,6 +177,8 @@ class DebModel extends _$DebModel {
     );
     try {
       await packageKit.waitTransaction(transactionId);
+    } on PackageKitTransactionCancelled {
+      // User cancelled (e.g. dismissed the polkit dialog) — not an error.
     } on Exception catch (e) {
       // Report via the same path as PackageKitServiceError events so the
       // page shows the error and clears the stuck transaction state.
@@ -185,6 +187,10 @@ class DebModel extends _$DebModel {
           code: PackageKitError.internalError,
           details: e.toString(),
         ),
+      );
+    } finally {
+      state = AsyncValue.data(
+        state.value!.copyWith(activeTransactionId: null),
       );
     }
     ref.invalidateSelf();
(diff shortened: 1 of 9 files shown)


packages/app_center/lib/deb/deb_model.dart
```diff
@@ -150,26 +150,43 @@ class DebModel extends _$DebModel {
...
     );
-    await packageKit.waitTransaction(transactionId);
+    try {
+      await packageKit.waitTransaction(transactionId);
+    } on Exception catch (e) {
+      // Report via the same path as PackageKitServiceError events so the
+      // page shows the error and clears the stuck transaction state.
+      await _onError(
+        PackageKitServiceError(
+          code: PackageKitError.internalError,
```
matthew-hagemann, line 187: This is on the right track, but there are several different kinds of errors that get caught here and are now all turned into internal errors. 

The main issue is `PackageKitTransactionCancelled`, which is when the user presses cancel, is treated as an internal error, where it should be explicitly ignored. (This is existing behavior, but we might as well fix it while we are handling this while we are fixing the error handling for updates :slightly_smiling_face:)

You'll also want to clear the active transaction, see packages/app_center/lib/manage/local_deb_updates_model.dart:154

https://github.com/user-attachments/assets/12868dc5-cc49-4be4-92c7-c126581127bc

packages/app_center/lib/manage/local_deb_updates_model.dart
```diff
@@ -145,6 +146,9 @@ class LocalDebUpdatesModel extends _$LocalDebUpdatesModel {
         activeTransactionId: null,
       );
       ref.read(installedAppsProvider.notifier).addDebToList(updatedDeb);
+    } on Exception catch (e) {
+      log.warning('Update transaction failed: $transactionId for $debId: $e');
+      ref.read(errorStreamControllerProvider).add(e);
```
matthew-hagemann, line 157: Thanks! This is halfway to solved. The current behavior in the `store_app.dart` (around line 127) listens to this stream, but only currently handles snap errors. Can you update it to handle `PackageKitTransactionError`  &  `PackageKitServiceError` so this handling here is not dropped.

matthew-hagemann (changes requested):
Thanks! This solves the blocked packages issue on the managed page from my testing 🙂 

Before:
Image

After:
Image

Couple of comments on error handling:

packages/app_center/lib/deb/deb_model.dart
Yashpal-0: Thanks for the feedback! All three points are fixed in commits 6c2cf1a0 and 88e28529.

`PackageKitTransactionCancelled` is now caught explicitly and only clears the transaction state. I normalized this at the source in `packagekit_service.dart` so `waitTransaction` treats cancellations reported via error code the same as exit codes. This also fixed an issue in `local_deb_updates_model` where canceled bulk updates were mistakenly flagged as user-facing failures.

Other failures now pass through the specific error code reported by PackageKit, falling back to `internalError` only if none is provided. To make this reliable, I fixed a stream timing issue where `waitTransaction` attached its listener after `transaction.events` had already delivered the `ErrorCode` event. Event recording is now handled directly in the `_createTransaction` subscription, which attaches before actions are dispatched.

`activeTransactionId` is now cleared on both success and error paths.

Finally, I removed the page's subscription to the service-wide `errorStream`. It was picking up background transaction errors and showing them on unrelated deb pages. Since the page already tracks its own failures through `waitTransaction`, the global stream was unnecessary.

Yashpal-0: Fixed in 4c61a63c. `waitTransaction` now throws a dedicated `PackageKitTransactionCancelled` subtype when a transaction finishes with `PackageKitExit.cancelled`, so callers can tell cancellation apart from failure at a single point. `_packageKitAction` ignores cancels (no more internal-error dialog) and clears the active transaction in `finally`. `LocalDebUpdatesModel.updateDeb` does the same.

Yashpal-0: Fixed in commit 4c61a63c.

waitTransaction now throws a dedicated PackageKitTransactionCancelled exception whenever a transaction ends in PackageKitExit.cancelled (like when a polkit dialog is dismissed). This gives callers a single place to handle cancellations separately from actual errors. DebModel._packageKitAction now ignores cancels instead of treating them as internalError, and it clears activeTransactionId inside a finally block so the page cannot get stuck spinning. LocalDebUpdatesModel.updateDeb was updated to follow the exact same pattern.

packages/app_center/lib/manage/local_deb_updates_model.dart
Yashpal-0: Done in commit 4c61a63c.

I widened the global error stream type to `Object` since `PackageKitServiceError` doesn't implement `Exception`. Over in `store_app.dart`, the listener now displays dialogs for both `PackageKitTransactionError` and `PackageKitServiceError`, while explicitly skipping `PackageKitTransactionCancelled`. I've also added widget tests covering all three cases.

Yashpal-0 (commented):
Both review threads addressed in commit 4c61a63c. Resolving outdated threads.

- `PackageKitTransactionCancelled` now ignored separately; `activeTransactionId` cleared in `finally` (deb_model.dart:180-195).
- Error stream widened to `Object`; `store_app.dart` handles `PackageKitTransactionError`/`PackageKitServiceError`, explicitly skips `PackageKitTransactionCancelled` (store_app.dart:127-143, error_stream_provider.dart:10).

Ready for re-review.

matthew-hagemann (changes requested):
https://github.com/user-attachments/assets/53c9017b-32c2-4f4f-92ef-a21f3d2ba538

Video demo from our 1:1 showing the flow for triggering the new error pop up on a polkit cancel on the manage page.

https://github.com/user-attachments/assets/1f360c2d-f96c-4d0b-8666-abad993978a5

Same issue happens in the page itself, but this is not new behavior, but something we did want to fix with this pr. Its probably the same error code both are catching both can now ignore.

Doesn't happen with snap, which is correct behavior:

[Screencast From 2026-09-08 14-48-47.webm](https://github.com/user-attachments/assets/e9e8c9ce-d658-4491-81f2-c886bec7e038)

Yashpal-0:
Fixed in 3c43cb3: both the Manage page and individual deb pages now treat a
dismissed polkit dialog as a normal cancellation and stay silent:

[Screencast From 2026-09-11 05-59-29-trimmed.webm](https://github.com/user-attachments/assets/0280358a-699f-432d-9f57-2866100ed893)

While testing, I also noticed that cancelling "Update All" still prompted
for the password once per remaining update — the batch just moved on to the
next deb. Fixed in 7f89f0a: cancelling now stops the whole batch, while a
genuine package failure still lets the remaining updates continue.

matthew-hagemann (dismissed):
LGTM! Thank you for the changes! I see you have a merge conflict with a mock to fix and we can merge :slightly_smiling_face:

matthew-hagemann (approved):

PR #2180, opened 2026-09-05
fix(ratings): rate the installed revision, not the one in the store

### LINKED ISSUE #1717: Ratings: Votes are being cast against tracking revision
The ratings model uses the latest revision from the tracking channel when casting a vote. This means if there's a pending update for a given snap that hasn't been installed yet, the data sent to the ratings backend will contain the revision of the pending update rather than the one locally installed.

Discovered this while adding some debugging information:
```
flutter: INFO ratings_model: Casting vote: VoteStatus.up for snap: 3wdHCAVyZEmYsCMFDE9qt92UV8rC8Wdk, revision: 4483
```

```
$ snap info firefox
name:      firefox
summary:   Mozilla Firefox web browser
publisher: Mozilla✓
store-url: https://snapcraft.io/firefox
contact:   https://support.mozilla.org/kb/file-bug-report-or-feature-request-mozilla
license:   unset
description: |
  Firefox is a powerful, extensible web browser with support for modern web application
  technologies.
commands:
  - firefox
  - firefox.geckodriver
snap-id:      3wdHCAVyZEmYsCMFDE9qt92UV8rC8Wdk
tracking:     latest/stable
refresh-date: 4 days ago, at 18:24 CEST
channels:
  latest/stable:    127.0.2-1     2024-06-26 (4483) 281MB -
  latest/candidate: 127.0.2-1     2024-06-26 (4483) 281MB -
  latest/beta:      128.0b6-1     2024-06-21 (4466) 281MB -
  latest/edge:      129.0a1       2024-06-26 (4485) 300MB -
  esr/stable:       115.12.0esr-1 2024-06-11 (4373) 256MB -
  esr/candidate:    115.13.0esr-1 2024-06-21 (4461) 256MB -
  esr/beta:         ↑                                     
  esr/edge:         ↑                                     
installed:          127.0.1-1                (4451) 281MB -
```

vicajilau:
Fixes #1717.

Ratings are recorded per revision, but `RatingsModel.build` reads the revision from `SnapData.snap`:

```dart
Snap get snap => storeSnap ?? localSnap!;
```

That prefers the store snap, which carries the newest revision in the channel. Whenever an update is pending the two differ, so the vote is attributed to a revision the user has not run.

It cuts both ways. With firefox installed at 4400 and 4483 available:

- casting a vote records it against 4483
- an existing vote on 4400 is not found by `_getUserVote`, so the buttons come up unset and the same user can vote again

Neither is visible in the UI, and it only happens while an update is pending, which is why it goes unnoticed. The cost is on the backend, where per-revision aggregation is the point of sending `snapRevision` at all.

The rating buttons are only rendered for an installed snap:

```dart
if (snapData.isInstalled)
  ... _RatingsActionButtons(
```

so `localSnap` is always there to read when a vote can be cast, and there is no case where the store revision is the only thing available to a voter.

Both tests fail against the current code with `Expected: <42> Actual: <99>`.

One thing left alone: `RatingsData` is cached for a day, so an entry written before this change still carries the store revision until it expires or a vote deletes it. Self-correcting, and clearing the cache on upgrade seemed like more than this fix warrants, but say the word if you would rather handle it.

### DIFF
--- a/packages/app_center/lib/ratings/ratings_model.dart
+++ b/packages/app_center/lib/ratings/ratings_model.dart
@@ -16,8 +16,16 @@ class RatingsModel extends _$RatingsModel {
 
   @override
   Future<RatingsData> build(String snapName) async {
-    final snap = (await ref.watch(snapModelProvider(snapName).future)).snap;
-    final snapId = snap.id;
+    final snapData = await ref.watch(snapModelProvider(snapName).future);
+    final snapId = snapData.snap.id;
+
+    // Ratings are recorded per revision, and the rating buttons are only shown
+    // for an installed snap, so the revision being rated is the installed one.
+    // `SnapData.snap` prefers the store snap, which carries the newest revision
+    // in the channel and differs from the installed one whenever an update is
+    // pending, so reading the revision from it attributes the vote to a
+    // revision the user has not run.
+    final snapRevision = snapData.localSnap?.revision ?? snapData.snap.revision;
 
     final cacheFile = _getCacheFile(snapId);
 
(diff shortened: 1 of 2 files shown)


ogres (approved):
Thanks for fixing this 👍

PR #2182, opened 2026-09-07
fix: bundle and fallback to Noto Sans CJK for missing Chinese glyphs

wthIC:
## Summary

Flutter has a known issue on ARM64 Linux platforms (flutter/flutter#90951) where Chinese glyphs fail to render and display as tofu boxes (`□□□`). Within snap confinement, the application lacks access to necessary CJK glyph tables, causing unrendered characters across the interface.

Ahead of an upstream release for the Lenovo partnership, this provides a workaround in App Center by bundling the fonts and declaring explicit fallbacks.

## Changes

**Add Noto Sans CJK to font fallbacks** — In `packages/app_center/lib/store/store_app.dart` (inside `StoreAppThemeX`), added `Noto Sans CJK (SC, TC, HK)` to `fontFamilyFallback`. 

**Stage CJK fonts in the snap** — Added `fonts-noto-cjk` to `stage-packages` in `snapcraft.yaml` so the required font files are bundled and accessible inside snap confinement at runtime.

## Testing

Manual verification on ARM64 hardware:
1. Build and install the snap on an ARM64 test machine.
2. Launch App Center with a Chinese locale (`LANG=zh_CN.UTF-8`).
3. Verify that Chinese glyphs render properly across all views without tofu boxes (□□□).

---

UDENG-11262

### DIFF
--- a/packages/app_center/lib/store/store_app.dart
+++ b/packages/app_center/lib/store/store_app.dart
@@ -93,10 +93,7 @@ class _StoreAppState extends ConsumerState<StoreApp> {
 }
 
 class _StoreAppHome extends ConsumerWidget {
-  const _StoreAppHome({
-    required this.navigatorKey,
-    required this.searchFocus,
-  });
+  const _StoreAppHome({required this.navigatorKey, required this.searchFocus});
 
   final GlobalKey<NavigatorState> navigatorKey;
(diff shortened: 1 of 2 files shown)


ogres (approved):
👍

PR #2187, opened 2026-09-14
fix: add fallback fonts to missing places

ogres:
We missed some places where fonts need to be set, causing snackbars and popups to not be visible with proper fonts. Also the font was chosen by order list, now we get it based on locale, as they differ.

## BEFORE
before

## AFTER
after

UDENG-11262


ashuntu (approved):
LGTM, though some of this should probably be resolved in https://github.com/ubuntu/yaru.dart so all projects inherit the benefits

Yaru.dart does do some font overrides that would probably be relatively easy to resolve. E.g. https://github.com/ubuntu/yaru.dart/blob/main/lib/src/themes/common_themes.dart#L636

wthIC (approved):

ogres:
> LGTM, though some of this should probably be resolved in https://github.com/ubuntu/yaru.dart so all projects inherit the benefits
> 
> Yaru.dart does do some font overrides that would probably be relatively easy to resolve. E.g. https://github.com/ubuntu/yaru.dart/blob/main/lib/src/themes/common_themes.dart#L636

Once new version of Flutter is released in November, it will have the fix available. We can update to that, and remove our workaround fully. This is a temporary workaround until then.

Issue #2110, opened 2026-05-08
Add `apt://` URI handler


mhalano:
### Is there an existing issue for this?

- [x] I have searched the existing issues and found none matching what I'm reporting.

### Describe the feature, why the feature matters to you, and what benefits the feature will bring about.

With this, we could click in links of the type `apt://` to make installing a package easier.

### Describe any alternatives you've considered.

Years ago, the `apt://` URI handler was managed by apturl. Since it's not installed anymore, that would be nice that App Center could handle this URI in a similar way how it already does to `snap://`.

It's quite similar to this other issue, https://github.com/ubuntu/app-center/issues/2106.

I already opened a bug report for `gnome-session` to deal with the MIME type, but I found out later that the `snap-store` couldn't deal with the URI: https://bugs.launchpad.net/ubuntu/+source/gnome-session/+bug/2151743

wathika-eng:
Is it related to this? When I tried to install a .deb file.

<img width="1403" height="854" alt="Image" src="https://github.com/user-attachments/assets/9b82d66a-8ddd-4cc8-8bfd-869283b89e9e" />

mhalano:
@wathika-eng Your problem should be fixed now. I can open Nautilus and click on a .deb file and opens the App Center correctly. My issue is more related to websites that use `apt://<package name>`. This URI used to be supported by apturl in the past.

wathika-eng:
> [@wathika-eng](https://github.com/wathika-eng) Your problem should be fixed now. I can open Nautilus and click on a .deb file and opens the App Center correctly. My issue is more related to websites that use `apt://<package name>`. This URI used to be supported by apturl in the past.

Okay, let me just open a new issue, because I can still reproduce the above bug when trying to install a .deb file

Issue #2106, opened 2026-04-29
Add `appstream://` URI handler


aleasto:
### Is there an existing issue for this?

- [x] I have searched the existing issues and found none matching what I'm reporting.

### Describe the feature, why the feature matters to you, and what benefits the feature will bring about.

Since we do support package discovery via appstream from within the app, it would be nice to also support opening `appstream://` URIs, which are usually provided by online stores like https://apps.kde.org.
This would be equivalent to opening a `snap://` URI.

wathika-eng:
Seems like the same underlying gap: App Center (or the system) isn't picking up URI scheme handlers registered by installed snaps.

<img width="476" height="455" alt="Image" src="https://github.com/user-attachments/assets/2bab5dc2-7b6b-4ee3-944a-2568ea65bd83" />

mhalano:
@wathika-eng I think it's a different problem. The goal here is that Snap can recognizes the `appstream://` URI. Why App Center is being used to open the `discord://` URI, it's a different problem and probably needs its own issue.

PR #2189, opened 2026-09-15
fix: drop the share button on deb pages, show it on every snap page

### LINKED ISSUE #1929: Share button copies link to the app developer page
### Is there an existing issue for this?

- [x] I have searched the existing issues and found none matching what I'm reporting.

### Bug Description

The share button in the app page links to the app developer page. It should link to the snap store page: there's a different location for developer links further down.

For instance, the button on the top right here:

Image

Copies the link to https://brave.com instead of to https://snapcraft.io/brave

### Steps to reproduce the behavior

Any app page in the App Center.

### Expected behavior

_No response_

### Ubuntu release

None

### What architecture are you using?

None

### System info

_No response_

### Additional context

_No response_

vicajilau:
Fixes #1929.

The share button on a deb page copies `component.website`, the project's own site, which reads as sharing the app when it is not. #1951 fixed that for snaps by copying the snapcraft.io page. There is no equivalent for a deb: nothing opens one in App Center yet, and the archive page is a version and dependency table.

Following the discussion below, this removes the button from deb pages rather than pointing it somewhere else. The website is still in the metadata under a label that says what it is, and a share button that visibly behaves the same for snaps, debs and other stores is worth more than one that sometimes copies something else. If a web link for App Center's deb data ever exists, the button can come back pointing there.

The snap page only showed the button when the snap declared a website, a condition left over from when the website was what got copied. The link no longer depends on it, so the button now shows on every snap page. That is a small visible change: it appears on more snaps than before.

Two tests: a deb page offers no share button, and a snap page offers one without a website.


ogres:
@vicajilau Thanks for the contribution. I am not sure whether this is the right approach and whether this will confuse users more than it will help.
For snaps, the story is different - because each snap has a proper page, but with deb packages, the information available on packages.ubuntu.com might be too technical.
@juanruitina what do you think about it?

vicajilau:
Fair concern, and it is a UX call more than a code one, so good to have @juanruitina on it. To make the choice quick, the options as I see them from the recipient's side:

1. **`packages.ubuntu.com/<package>`**, what this PR does. Official and always exists, but it is a version and dependency table. Someone sent it to "check out this app" lands on something that looks like an admin page.
2. **The project's website**, the current behaviour. Human-facing and tells the recipient what the app is. But it is neither App Center nor Ubuntu, and not every deb has one, so the button disappears for some apps.
3. **No share button on deb pages** until there is a link that opens the app here, which #2106 and #2110 would provide. Honest, but a visible regression from today.

I would lean to 2 as the least surprising, keeping the button hidden when there is no website, which is what the code does today. If that is the call, this PR shrinks to the snap side: the button on a snap page was also gated on the website even though it copies the snapcraft.io page, so snaps without a website lost it for no reason. Happy to cut it down to that.

juanruitina:
Hey @vicajilau, thank you for looking into this. I'd say option 3 makes the most sense.

I think avoiding confusion takes precedence: users will expect consistent behaviour from this button if visible, both internal (snaps vs debs) and external (what such a button would do in other platforms, app stores). You can get to the App Center (and even get instructions on how to install snaps!) from the snap link; you cannot from the deb link to the archive or the developer's website; indeed, packages.ubuntu.com can be confusing. So I think hiding it is the best option.

I don't think it's a regression: the link to the developer's website is still available in the metadata and with a clearer label; I don't think it is conveyed at all that the button copies the link to the developer's website, which is in itself confusing. And, with this solution, we also avoid having only _some_ debs showing the button.

Of course, if we ever have a web link to show the App Center data for debs as we do for snaps, we could point there.

vicajilau:
Done, went with option 3. The deb share button is removed, and the snap page now shows its button regardless of whether the snap declares a website, since the snapcraft.io link never depended on it. Title and description updated, tests cover both. Thanks @juanruitina, the consistency argument was the one I had not weighed.

ogres (approved):

PR #2194, opened 2026-09-19
fix(l10n): use Weblate Polish app title in packagekit desktop file

### LINKED ISSUE #1842: Small translation mismatch in the app-center.desktop file
`Name[pl]=Centrum programów` should be changed to `Name[pl]=Centrum oprogramowania`, i.e. the translated name from Weblate: https://hosted.weblate.org/translate/ubuntu-desktop-translations/app-center/pl/?checksum=553580f044432930

DevShoaib78:
Fixes #1842.

`app-center.desktop` is regenerated from the ARB files by `melos sync-desktop-titles`, so on `main` it already reads `Name[pl]=Centrum oprogramowania`. The hand-maintained `app-center-packagekit.desktop` (the PackageKit session-installer entry referenced from `snap/snapcraft.yaml`) was the one file still carrying the old `Centrum programów`. This changes it to the same title that Weblate, `app_pl.arb` and the generated desktop file use.

No code change, so no tests to add. `flutter analyze` in `packages/app_center` is clean, and re-running `scripts/import_app_title_to_desktop_file.py` produces no diff, so the generated-files freshness check is unaffected.


ogres:
Thanks for the contribution @DevShoaib78 
We have a script, that keeps desktop file translations in sync with Weblate, and we use it for other `.desktop` files - https://github.com/ubuntu/app-center/blob/35949c19e7ad239b19acf1f5bb5e7008639dec53/melos.yaml#L64

It would be better if we add similar behaviour for `app-center-packagekit.desktop` as well.
This way, the file will be kept always up to date with translations, and it will be covering all languages.

Would you be able to adjust it?

DevShoaib78:
Thanks @ogres, good call, that's the better fix. Done:

- `melos sync-desktop-titles` now runs `import_app_title_to_desktop_file.py` over `app-center-packagekit.desktop` as well, with the same ARB directory and `appCenterLabel` key.
- The packagekit file is regenerated. Its 15 hand-maintained `Name[...]` entries become the same 51 that `app-center.desktop` has, and nothing else in the file changes. Since the `l10n` job already runs the sync and then `check-outdated-files.sh`, it's now covered on every PR too.

While I was in there I noticed the `l10n` job's create-pull-request step lists `packaegs/app_center/assets/app-center.desktop` under `add-paths`, so the automatic "chore: regenerate l10n" PR has never picked up the desktop file. I fixed that in a separate commit and added the packagekit file next to it. Happy to drop that commit if you'd rather handle CI separately.

Locally, the script still reproduces `app-center.desktop` byte-for-byte, and re-running the sync over both files followed by `check-outdated-files.sh` leaves no diff. The workflows here are waiting on maintainer approval to run.

ogres (dismissed):

ogres (approved):

